PR #19, opened 2026-06-10
feat: configurable truncated_reward (zero|keep|soft) to align AR length handling with verl

leviking98z-rgb:
## Summary
`RewardService` hard-zeros the reward of any AR generation that hits `max_new_tokens` (sglang `finish=="length"`), as an anti-ramble guard. This diverges from verl: its `dapo` reward manager with `overlong_buffer.enable=False` keeps the math_verify score on the partial text (no zeroing), and with `enable=True` applies a *graded* overlong penalty — never a hard zero. As responses lengthen during training and truncation approaches ~100%, the hard-zero collapses `reward_mean`, while verl stays flat. Observed on Qwen3-4B DAPO-math DRPO: unirl `reward_mean` 0.53→0.36 (as truncation rises) vs verl flat ~0.5 even at 100% truncation.

This PR adds a `truncated_reward` mode to `RewardService`:
- `zero` (default, unchanged) — force reward 0 on truncated traces (anti-ramble).
- `keep` — keep the raw score on the partial text (= verl `dapo`, overlong disabled).
- `soft` — verl DAPO graded overlong penalty over the last `overlong_buffer_len` tokens before `max_new_tokens` (mirrors `verl.workers.reward_manager.dapo`).

It opts the qwen3 GRPO/DRPO DAPO-math recipes into `truncated_reward: keep` for verl parity. Default behavior is unchanged for every other recipe.

## Test Plan
- `ruff check` and `ruff format --check` pass on `unirl/reward/service.py`.
- `scripts/check_recipe_targets.py` passes (967 _target_ paths resolve).
- Both edited recipes parse as YAML and compose; `truncated_reward` is a valid `RewardService.__init__` kwarg.
- Behavioral validation in progress: a Qwen3-4B DAPO-math DRPO run with `truncated_reward: keep` to confirm the late-stage `reward_mean` no longer declines (tracks verl).

## EXP

BEFORE.
截屏2026-06-10 12 27 47

AFTER.
截屏2026-06-10 12 25 30


haonan3 (approved):
LGTM

PR #21, opened 2026-06-10
feat: roll out DiffusionNFT under the EMA shadow adapter

Jayce-Ping:
## Summary

Roll out DiffusionNFT under its EMA-smoothed `old` shadow adapter on a SEPARATE rollout engine (SGLang / vLLM-omni), the way `requires_ema_rollout` intends. The in-process `apply_eval_ema` swap only touches the train process, so a separate engine never saw the EMA weights.

Reworked (rebased fresh on latest `main`) into a single, transport-agnostic mechanism owned by the backend:

- **Backend single source of truth** (`unirl/train/backend/fsdp.py`): `FSDPBackend.rollout_adapter_name` returns the EMA shadow (`"old"`) for adapter-EMA, else `"default"`. The in-process eval-EMA swap and the cross-process weight sync both derive from it, so they cannot disagree.
- **One unified `adapter_name` knob** across both weight-sync families (`weight_sync/full/*`, `weight_sync/lora/*`): defaults to `None` -> asks the backend; an explicit value overrides. Full-sync folds the shadow into the merged base (SGLang); LoRA-sync ships the shadow adapter directly (vLLM-omni). Also fixes a latent bug where `_extract_canonical_lora` never forwarded `adapter_name`, so the LoRA path always shipped `default`.
- **Trainer untouched**: no per-recipe config-reaching; the sync derives the adapter from the backend.
- **FSDP grad fix** (`unirl/train/inject.py`): `_activate_keep_grad` re-asserts the trainable/frozen `requires_grad` split across rollout/restore swaps, so the trainable `default` is never all-gathered grad-less under `reshard_after_forward=False`.
- **SGLang ODE path** (`unirl/rollout/engine/sglang/*`): DiffusionNFT's `num_sde_steps=0` resolves `sde_indices` to `[]`; request the deterministic ODE rollout, always collect the dit-trajectory, and keep the full trajectory.

## Related Issue

N/A

## Test Plan

- `python -m py_compile` on all touched files: passes.
- Linter on touched files: clean.
- Standalone CPU logic checks (no torch in env): backend `rollout_adapter_name` (`old` with `ema_lora_cfg`, else `default`); both sync families resolve `None -> backend`; full-sync fail-closed guard fires for non-default + `lora_merged=False`; `_extract_canonical_lora` forwards `adapter_name`; SGLang empty-`sde_indices` -> ODE branch + full trajectory; EMA update math.
- Not run; reason: no local GPU. Recommended e2e: `examples/diffusion/sd3_nft.yaml` (trainside trains past the first backward) and `examples/diffusion/sd3_nft_sglang.yaml` (separate, rolls out under EMA `old`).

## Compatibility / Risk

- Backward compatible: GRPO and plain-LoRA recipes resolve `adapter_name` to `"default"` (unchanged behavior). `FullWeightSync` gains an `adapter_name` ctor arg (defaults to `None`); the LoRA-sync `adapter_name` default changes from `"default"` to `None` (resolves to `"default"` for non-EMA backends), so existing recipes are unaffected.
- Requires the sync handler's `backend` to expose `rollout_adapter_name`; `FSDPBackend` is the only training backend, so all paths are covered.

## Reviewer Notes

- AI-assisted (Cursor). Rebased fresh on latest `main`; this force-push supersedes the earlier full-sync-only `lora_adapter` + trainer `_resolve_sync_lora_adapter` approach with the backend-sourced unified mechanism.
- Out of scope: no new vLLM-omni NFT recipe is shipped, but the unified design makes it work via either `RemoteLoraWeightSync` (ships `old`, no merge) or `IPC/NCCLWeightSync + lora_merged: true` (merged `old` base).

## Checklist


celve:
Do we have the e2e test result?

haonan3 (approved):
**This PR lands the DiffusionNFT EMA-shadow rollout.** For DiffusionNFT, trainside is the primary path: being off-policy, it runs fully and correctly in-process, and
  that's what most runs will use. The separate-mode sglang+NFT path is a secondary speed/scale option, not the mainstream one — so its remaining cleanups are relatively
  minor, and we merge now and address them in a small follow-up PR (kept separate so this one stays reviewable):

  1. Unify forward-process detection. The sglang request/engine path currently detects the forward-process (empty sde_indices) case with ad-hoc []-vs-None / truthy-vs-is
  not None checks. The follow-up will route it through a single explicit predicate (an is_forward_process flag / rollout mode). This also overlaps a pre-existing,
  repo-wide sde_indices pattern shared with GRPO, so it's cleaner to do on its own.
  2. Make the NFT rollout read as "sample → final latent" instead of implicitly reusing the GRPO rollout-trajectory API. Structure/readability only — sglang's t2i path
  exposes the final latent only through the trajectory return, so behavior is unchanged; the follow-up just makes the intent explicit.

PR #22, opened 2026-06-10
feat: log generated rollout media to wandb in DiffusionTrainer

Jayce-Ping:
## Summary

Wires up wandb logging of generated rollout media (images/videos) in `DiffusionTrainer`. The `logging.log_media` / `media_max_items` / `media_log_interval` recipe keys were previously inert — nothing read them, and `build_media_preview_for_track` / `log_generated_media` were never called, so no media reached wandb.

New `_maybe_log_media` builds a `MediaPreview` from each track's `decoded` images/videos and logs it right before `_drop_decoded`. Because `Worker.call` packs every returned tensor into a `TensorMeta` proxy, `track.decoded` is hydrated via `map_tree` + `_hydrate_tensor_meta` before building the preview. Gated by `log_media` and throttled by `media_log_interval` (hydration pulls the full decoded batch to the driver).

## Related Issue

N/A

## Test Plan

- `python -m py_compile unirl/trainer/diffusion.py` (passes).
- Static checks: `Images`/`Videos` are `Batch` subclasses, so `map_tree` recurses and hydrates `.pixels`; `_hydrate_tensor_meta` is the same driver-side path already used for `track.rewards`.
- Not run: end-to-end GPU run. Recommended: `examples/diffusion/sd3_nft.yaml` (trainside) with `REPORT_TO_WANDB=true` and a modest `media_log_interval`, then confirm panels under `rollout/generated_media`.

## Compatibility / Risk

- No-op unless `logging.log_media` is set; off by default, so existing runs are unchanged.
- Covers the trainside/colocate path; separate-engine recipes (sglang/vllm) return decoded differently and may need follow-up.
- Hydration is per-logged-rollout and pulls the full decoded batch — keep `media_log_interval` above 1 on large batches.

## Reviewer Notes

AI-assisted; checked upstream open PRs for overlap (none). Builds on existing helpers (`build_media_preview_for_track`, `UniRLWandBLogger.log_generated_media`) that were unused before this change.

## Checklist

Made with [Cursor](https://cursor.com)

### DIFF
--- a/unirl/trainer/diffusion.py
+++ b/unirl/trainer/diffusion.py
@@ -3,7 +3,7 @@ import inspect
 import logging
 import os
 import time
-from typing import Optional, Tuple
+from typing import Any, Optional, Tuple
 
 import torch
 from hydra.utils import get_class, instantiate
@@ -64,6 +64,15 @@ class DiffusionTrainer(BaseTrainer):
         # ``use_global_std=True`` scale) instead of each prompt's own std. Off by
         # default → unchanged per-group GRPO normalization for every other recipe.
         self._adv_use_global_std = bool(adv_use_global_std)
+        # WandB rollout media (generated images/videos): read from the optional
+        # ``logging`` block. ``log_media`` gates it; ``media_log_interval`` throttles
+        # it (every N rollouts); ``media_max_items`` caps samples per panel. Logged
+        # from the rollout's ``decoded`` payload before it is dropped (see train_step).
+        self._log_media = bool(logging_cfg.get("log_media", False)) if logging_cfg is not None else False
+        self._media_max_items = int(logging_cfg.get("media_max_items", 8)) if logging_cfg is not None else 8
+        self._media_log_interval = (
+            max(1, int(logging_cfg.get("media_log_interval", 1))) if logging_cfg is not None else 1
+        )
         # Set in _build_rollout: True when the rollout is the trainside
         # direct-sampling engine (it reuses the train model → must NOT offload).
         self._rollout_is_trainside = False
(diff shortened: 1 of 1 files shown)


celve:
I think we can merge this first, and later I try to absorb all logging things into the wandb logger @haonan3

haonan3:
Thanks for wiring this up — the hydration path and call-site placement (after reward hydration, before `_drop_decoded`) all check out. I pushed f4774be to this branch with two changes on top:

**1. Skip non-media tracks before any fetch.** `_maybe_log_media` previously hydrated `track.decoded` first and let `build_media_preview_for_track` reject non-media payloads afterwards, so a tensor-bearing non-media track (e.g. `Audios`) would pull its full batch to the driver for nothing. Now an `isinstance(track.decoded, (Images, Videos))` gate runs before hydration.

**2. Hydrate only a shard-aligned prefix instead of the full decoded batch.** Only `media_max_items` samples reach the preview, but hydration pulled the whole batch — for video runs that's GBs per log event, and several recipes ship `log_media: true` + `media_log_interval: 1`. Since `TensorMeta` refs partition the batch by DP shard and support ref-boundary slicing (`_slice_by_refs`), the new `_ref_aligned_prefix_len` picks the smallest shard boundary covering `media_max_items`, slices the dehydrated batch there (pure metadata op), and hydrates just that prefix — one shard pulled instead of N. For `Videos` the frame-unit ref boundaries are mapped back to sample indices via the driver-side `cu_seqlens` (which stays a real tensor through `Worker.call` packing); any unmappable boundary falls back to full hydration, i.e. the previous behavior.

Also hoisted the function-local imports to module scope — the local `_hydrate_tensor_meta` import shadowed the existing top-level one.

Verified with a driver-side harness that mirrors `Worker.call` packing (real `map_tree` / `Batch.slice` / `build_media_preview_for_track`, faked store handles with fetch counting): an 8-sample / 4-shard Images batch pulls exactly one shard and the logged preview is pixel-identical to the source samples; ragged-T Videos slice consistently through `cu_seqlens`; non-media tracks trigger zero fetches; already-hydrated (real-tensor) decoded is unchanged. pre-commit passes.

haonan3 (approved):

PR #24, opened 2026-06-10
refactor(logging): absorb all wandb/media/console logging into one logger

celve:
## Summary

Centralizes all training-time logging in `UniRLWandBLogger` so the trainers
(`diffusion` / `ar` / `pe` / `unified_model`) no longer carry wandb, metric,
media, console, or step-counter logic — they just hand the logger the rollout
objects.

- `UniRLWandBLogger` gains `log_rollout_step` (rollout/train/perf metrics + media,
  `per_update`-aware, single- and multi-track), `log_progress` (the console
  summary, prints even when wandb is off), and `should_log_media`; it now owns the
  optimizer-step counter. Dead `get_logger`/`set_logger`/`_global_logger` removed;
  `image_log_interval` → `media_log_interval`, joined by `media_max_items`/`log_media`.
- Media is wired generally at the `_drop_decoded` chokepoint via the shared
  `build_media_preview_for_track`, capped + logged per track, with a `prompts`
  override so PE/unified captions align with the expanded image track. This
  **generalizes #22** (which logged media only in `DiffusionTrainer`) to every
  trainer and folds its `TensorMeta` ref-aligned slicing + hydration
  (`_ref_aligned_prefix_len`) into the shared builder, so the preview hydrates one
  DP shard instead of the full dehydrated `decoded` batch.
- `BaseTrainer._init_wandb` is the single always-construct factory (a disabled
  null-object when reporting is off, so trainers call it without guards);
  `unified_model` is folded onto the shared path, dropping its duplicate
  `_init_wandb` and inline logging block — and gaining the reward/advantage
  distribution stats and per-step timing it previously lacked.
- Enabling wandb now inherently requires a successful init (it raises on failure);
  the separate `require_success` flag is removed.

## Related Issue

N/A. Builds on (and generalizes) the merged #22.

## Test Plan

Local, on a CPU box (no GPU):

- `python3 -m py_compile` on all changed files — passes.
- Import-cycle guard: `import unirl.utils.wandb_logger` does **not** pull in
  `unirl.train.stack`; all four trainers + `unirl.types.media_preview` import clean.
- Logic harness (monkeypatched `wandb.log`) covering:
  - disabled null-object: every wandb method no-ops, `log_progress` still prints;
  - metric routing/axes: `rollout/*` (reward/advantage/zero-std), `train/*`,
    `perf/rollout_time_s`, `train/step` = optimizer step, `rollout/step` = `rollout_id+1`,
    single-track unprefixed vs multi-track `<track>/` namespacing, `extra_metrics` merge;
  - `per_update` matrix: single → N points; multi-track equal-length → merged points;
    multi-track unequal-length → single aggregate fallback (never interleaves PE's two optimizers);
  - media: builds, caps to `media_max_items`, captions via the `prompts` override,
    `_ref_aligned_prefix_len` is a no-op on real tensors, `_drop_decoded` builds-then-nulls
    on cadence and skips off-cadence;
  - enabling wandb with init unavailable raises; disabled never raises.

Not run; reason: GPU e2e (1-GPU `unirl.train_diffusion` smoke + multi-GPU
`unified_model`), including the real dehydrated-`decoded` → hydrate media path,
needs a GPU pod. Suggested smoke once on hardware:
`report_to_wandb=false python -m unirl.train_diffusion --config-name diffusion/sd3_trainside num_rollouts=1`.

## Compatibility / Risk

- **`unified_model` wandb panels change**: `rollout/step` shifts from `rollout_id`
  to `rollout_id+1`; its previously duplicated `rollout/<track>_loss` etc. are
  dropped (training scalars now live in `train/` only); it gains
  `rollout/reward_*`/`advantage_*`, zero-std stats, and `perf/rollout_time_s`.
- **`diffusion`/`ar`/`pe` metric keys/axes are unchanged** (same helpers).
- **Init strictness**: enabling wandb now raises on init failure for every trainer
  (diffusion/ar/pe previously warned-and-continued). `require_success` config key
  removed.
- No checkpoint, data-format, or public training-API changes. New `logging.*` media
  keys (`log_media`/`media_max_items`/`media_log_interval`) are optional and default off.

## Reviewer Notes

- AI-assisted (Claude Code); the diff and local checks were reviewed by the submitter.
- Duplicate-work check: no overlapping open PR; #22 (merged) added diffusion-only
  media logging — this PR keeps that behavior, generalizes it to all trainers, and
  moves its `TensorMeta` hydration into the shared builder. Start review at
  `unirl/utils/wandb_logger.py` (the absorbed surface) and `unirl/trainer/base.py`
  (`_init_wandb` factory + `_drop_decoded` media chokepoint).
- The dehydrated-`decoded` → hydrate media path is exercised only with real tensors
  locally; please confirm it on GPU before relying on media panels.

## Checklist

- [x] I updated tests, docs, and configs where needed, or explained why not (no test suite covers this area; `examples/unified_model/hi3_vllmomni.yaml` updated with media keys).


celve:
Verified with SD3, PE (Qwen + SD3), and Hunyuan Image 3.

haonan3:
Pushed 871a784 + 5118e71: `_drop_decoded` now uploads each track's media preview on the spot (same `rollout/step` value and panel keys, single `should_log_media` gate) and frees it together with `decoded` so no preview payload rides the DP_SCATTER train dispatch, and the never-read `RolloutReq.collect_media_preview` / `media_max_items` flags plus the now caller-less `cap_media_preview` are removed

haonan3 (approved):

PR #38, opened 2026-06-12
fix(sde,flowdppo): derive KL sigma_t from each SDE strategy (flow/dance/cps)

Jayce-Ping:
## Summary

`FlowDPPO._compute_sigma_t` hardcoded the Flow-SDE std, so the KL normalizer was wrong for Dance and CPS (and, at the first `sigma == 1` step, for Flow too). This adds a single source of truth on the SDE strategies -- `_std_dev_t` + `transition_std` -- and routes each strategy's `step()` through them (numerically identical). `FlowDPPO._compute_sigma_t` now delegates to `strategy.transition_std`, so the KL `(d_mean)^2 / (2 * std^2)` uses each strategy's actual transition std: Flow/Dance use `std_dev_t * sqrt(-dt)`, CPS uses `std_dev_t` (no `sqrt(-dt)`). This also folds in the earlier fix to Flow's `sigma == 1` denominator (`sigma_max = sigmas[1]` instead of a `0.99` clamp).

## Related Issue

N/A

## Test Plan

- Pure-Python numeric check (no local torch env): the new `transition_std` equals each strategy's original `step()` std at representative sigmas including `sigma == 1` for Flow; CPS omits `sqrt(-dt)` (e.g. 0.859 vs 0.162 at step 0); `sigma < 1` Flow/Dance steps are unchanged.
- Not run; reason: no local Python/torch environment (`pre-commit`, `pytest`, Hydra config validation, and training/rollout smoke tests were not run). The `step()` refactor is numerically identical (same formula relocated into `_std_dev_t`); a rollout smoke on GPU is recommended.

## Compatibility / Risk

Low-to-moderate. Sampling/log_prob math is unchanged (each `step()`'s `std_dev_t` is the same formula moved into `_std_dev_t`). The behavior change is limited to the FlowDPPO KL normalizer, which is now correct per strategy (previously only Flow was handled, and Flow was wrong at `sigma == 1`). No config, checkpoint, data-format, or API changes. The new abstract `SDEStrategy._std_dev_t` is implemented by all SDE strategies (Flow/Dance/CPS); ODE/DPM2 is not an `SDEStrategy` and is unaffected.

## Reviewer Notes

AI-assisted. Single-source-of-truth refactor so `step()` and the KL share one std definition. Scope is limited to sigma_t consistency; other FlowDPPO gaps (reference-KL, advantage clipping, EMA) remain out of scope. Supersedes the earlier `sigma == 1`-only commit on this branch (now folded in).

## Checklist

- [x] I updated tests, docs, and configs where needed, or explained why not (no test infra exists for this path; verified numerically).

### DIFF
--- a/unirl/algorithms/flowdppo.py
+++ b/unirl/algorithms/flowdppo.py
@@ -363,7 +363,10 @@ class FlowDPPO(StageAlgorithm):
         s = sigmas[idx]
         s_next = sigmas[idx + 1]
         dt = s_next - s  # negative for denoising
-        std_dev_t = torch.sqrt(s / (1.0 - torch.clamp(s, max=0.99))) * eta
+        # Match FlowSDEStrategy.step at sigma==1: denominator uses sigma_max=sigmas[1],
+        # not a 0.99 clamp, so sigma_t equals the transition's std_var.
+        sigma_max = sigmas[1] if int(sigmas.shape[0]) > 1 else torch.tensor(0.99, device=device, dtype=sigmas.dtype)
+        std_dev_t = torch.sqrt(s / (1.0 - torch.where(s == 1.0, sigma_max, s))) * eta
         sigma_t = std_dev_t * torch.sqrt(-dt)
         return sigma_t.reshape(1, -1, 1, 1, 1)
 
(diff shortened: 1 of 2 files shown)


haonan3 (approved):
LGTM

Issue #40, opened 2026-06-12
VLM/LLM Training Efficiency


CjhHa1:
## Background
UniRL today mainly supports diffusion models, but AR and unified models are in high demand. This issue tracks a focused effort to bring UniRL's AR/VLM training and rollout infrastructure up to par, porting proven open-source (largely verl) performance techniques.

## Legend
- **Confidence** — how sure we are the technique helps *here*: ✅ validated / clear win · ⚠️ plausible, needs benchmarking on our stack.
- **Status** — ✅ merged · 🔄 in review (PR open) · 🧩 config available, benchmark pending · ⬜ not started.

## Implementation table

| Priority | # | Tech | Conf. | Modality | Benefit | Effort | Status |
|---|---:|---|:--:|---|---|---|---|
| **P0** | 1 | AR rollout prefix-cache hits (RadixAttention / prefix cache) | ⚠️ | AR/VLM | High (long prompts + multi-sampling) | S–M (expose config; real hits need fanout/routing) | ⬜ |
|  | 2 | FSDP2 `forward_prefetch` comm/compute overlap | ✅ | All | Medium (needs benchmark) | S (config + adjacent-block prefetch) | 🧩 (recipe flag) |
|  | 3 | Enable + benchmark `torch.compile` | ⚠️ | All | Medium (supported, needs measurement) | XS (config) | 🧩 |
|  | 4 | Rollout engine tuning cleanup (SGLang arg correction + defaults) | ⚠️ | AR | Low–medium | S (allowlist/recipes/docs) | ⬜ |
| **P1** | 5 | Sequence-length DP load balancing (verl `balance_batch`) | ✅ | AR/VLM | High (cuts DP stragglers on variable-length GRPO) | M (dispatch layer) | 🔄 #45 |
|  | 6 | Dynamic batching by token budget (`max_token_len_per_gpu`) | ✅ | AR/VLM | High (memory + throughput) | M (train stack) | 🔄 #42 |
| **P2** | 7 | Sequence packing / remove-padding (FlashAttention varlen) | ✅ | AR/VLM | High (removes padding waste) | L (per-model replay changes) | 🔄 #44 (Qwen3), #52 (Qwen-VL), #54 (train-side flash_attn) |
|  | 8 | Activation offload (FSDP) | ⚠️ | All | Medium (marginal after checkpointing) | M | ⬜ |
[...]

PR #6, opened 2026-06-08
refactor(algorithms): unify RL algorithm naming system

CjhHa1:
## Summary

This PR gives every RL algorithm in the repo **one consistent naming system** and applies it end-to-end across code, recipes, and docs.

### The system

Three casing worlds, kept strictly separate:

| World | Rule | Example |
| --- | --- | --- |
| Module / file names, recipe tags, config keys (engineering) | lowercase | `grpo.py`, `flowgrpo.py`, `diffusionnft.py` |
| Class names (engineering) | CapWords / uppercase acronym | `GRPO`, `DRPO`, `FlowGRPO`, `DiffusionNFT` |
| README / comments (prose) | the brand — never space- or hyphen-separated | `GRPO`, `FlowGRPO`, `DiffusionNFT` |

Canonical names:

| Algorithm | Module | Class | Prose | Recipe tag |
| --- | --- | --- | --- | --- |
| GRPO — AR (LLM/VLM) default | `grpo.py` | `GRPO` | GRPO | `grpo` |
| DRPO — AR (LLM) default | `drpo.py` | `DRPO` | DRPO | `drpo` |
| FlowGRPO — diffusion | `flowgrpo.py` | `FlowGRPO` | FlowGRPO | base recipe (`sd3_trainside`, …) |
| FlowDPPO — diffusion | `flowdppo.py` | `FlowDPPO` | FlowDPPO | `flowdppo` |
| DiffusionNFT — diffusion | `diffusionnft.py` | `DiffusionNFT` | DiffusionNFT | `nft` |
| DanceGRPO — FlowGRPO + Dance SDE | *(FlowGRPO)* | *(FlowGRPO)* | DanceGRPO | `dancegrpo` |
| MixGRPO — FlowGRPO + window sched | *(FlowGRPO)* | *(FlowGRPO)* | MixGRPO | `mixgrpo` |

**Rationale:** unprefixed `GRPO`/`DRPO` *are* the autoregressive (LLM/VLM) algorithms by default (GRPO is the LLM RL method in the literature; DRPO is our LLM RL method), so the old `AR` prefix was redundant and is dropped. Diffusion adaptations keep an explicit prefix (`Flow…` / `Diffusion…`). DanceGRPO and MixGRPO are FlowGRPO *recipe variants* (SDE-kernel / scheduler swap), not separate classes — they exist only as recipe tags + prose brands.

### The rename, in 5 reviewable layers

[...]

PR #42, opened 2026-06-12
perf(train): token-budget packed micro-batches via injected planner (verl dynamic-bsz parity) (issue #40)

leviking98z-rgb:
Part of the verl performance-parity series tracked in #40 (item #6: dynamic batching by token budget).

## Summary
`micro_batch_size=1` ran one sequence per forward/backward (32 isolated micro-steps per worker per rollout at 16 GPU) — the dominant cause of a measured 4.5x train-phase gap vs verl. This adds verl `ppo_max_token_len_per_gpu` parity: each update is packed into micro-batches under a token budget (length-sorted), exposed as a **pluggable planner** composed into `TrainStack` rather than a ctor knob or subclass.

**Composition.** `TrainStack` takes an injected `micro_planner` (default `CountPlanner` = the historical fixed-count behaviour, so the 60+ count-based configs are unchanged). `TokenBudgetPlanner` carries `token_budget` / `cost_model` and owns the seq-mean loss precondition (`validate`, checked once at construction). Recipes opt in with a `micro_planner` block — only `qwen3_drpo` does (`token_budget=10240`); the others stay count-based.

**Sort-then-slice.** The planner reorders the track **once** up front (length-sort within each update, one `track.select`) so every micro is again a **contiguous `(start, end)` range** — exactly the count-based geometry. The stack therefore only ever slices, and the π_old anchor reassembly is a plain ordered `cat`. No index-list micro plans, no slice/select branch, no original-index anchor bookkeeping threaded through `prepare_segment` — all packing-specific logic is isolated to `MicroPlanner.arrange` (a no-op for `CountPlanner`). Reorder cost is one gather per rollout (vs one per packed micro before).

Correctness:
- Update **membership is unchanged** — each update's permuted block holds its original samples, so the per-update gradient is the same set-sum.
- Micro losses are weighted by sample share (step gradient mathematically unchanged; unit-tested exact to 1e-12).
- **NCCL micro-count parity**: per optimizer step, all-reduce(MAX) the bin count and re-partition into exactly K bins — FSDP runs collectives per micro, unequal counts deadlock the process group.

**Package layout.** `unirl/train/stack` is split along its one real seam — pure planning (`planning.py`: `Plan` types, bin-packing math, the planner classes; CPU-unit-tested, no FSDP/Ray) vs the distributed driver (`base.py`: `TrainStack`). `__init__` re-exports the public surface, so recipes and `from unirl.train.stack import ...` resolve unchanged.

## Benchmark
16-GPU DRPO (Qwen3-4B-Base + DAPO, light regime, phase-timed): train phase **75.2s -> 54.3s**; ratio_mean stays 1.000.

## Test Plan
- unit tests (`tests/train/test_packing.py`, CPU-only): bin coverage/budget invariants, exact-K partition, sort-then-slice membership preservation, sample-share weighting (1e-12), seq-mean guard
- 16-GPU runs: `pack16_unirl_drpo_gz6` (wandb, project unirl-grpo) vs baseline `perfphase16_unirl_base`; ratio_mean≈1.0000 across all steps
- legacy path regression: omitting `micro_planner` (→ `CountPlanner`) preserves count-based behaviour


leviking98z-rgb:
**Review — request changes.** Great direction (packing + NCCL parity design is sound), but two blockers:

1. **`train()` grad_norm flow broken** (stack.py ~662): the CUDA-watermark block was inserted between the original `if has_backward:` / `else:`, so the `else` now binds to `torch.cuda.is_available()`. GPU + no-backward -> `UnboundLocalError`; CPU -> grad_norm silently forced to 0.0 + a spurious warning every step.
2. **Prompt lengths are dead code** (stack.py ~437): `resp_track.conditions` is a `Dict[str, Condition]`, so `getattr(conditions, "prompt", None)` always returns None -> the 2D packer is unreachable and the budget counts response tokens only, while the qwen3 packed replay forwards prompt+response — real micro size exceeds the budget by `count * prompt_len`. Use `conditions.get("prompt")`.

Also (non-blocking): example YAML should set `micro_cost_model: sum`; the "step gradient is unchanged" claim only holds for seq-mean agg modes — for `token-mean` (DRPO/GRPO default) sample-share weighting changes per-sequence weights vs both the mbs=1 baseline and verl token-count weighting; `replay_time_s`/`backward_time_s` are mean-per-micro, not phase totals; `_sync_micro_count` on the default PG is OK today but deadlocks if the lengths-missing fallback fires asymmetrically — assert symmetry.

Verified good: anchor scatter-back, multi-update plan sharing, bin non-emptiness, `Batch.select` packed-field gathering.

leviking98z-rgb:
Review findings addressed in 9dd5cbe: B1 (grad_norm else-branch restored) and B2 (dict accessor — the 2D/sum cost model now actually receives prompt lengths). Note: the published benchmark numbers were measured with the B2 bug live, i.e. budgets effectively counted response tokens only; prompt-aware accounting may slightly change bin shapes. Non-blocking items (token-mean weighting caveat, timing-metric semantics) acknowledged — will document rather than change behavior in this PR.

celve (approved):

PR #44, opened 2026-06-12
perf(qwen3): packed-varlen replay + flex_attention (replay split + gate) (issue #40)

leviking98z-rgb:
## Summary
Consolidates the **Qwen3 packed-varlen replay** story — packed varlen replay (was #43) + flex_attention/bucketing (was #44) — into one PR, based on `main`, with the `replay` refactor and a backend gate. Part of the verl-parity series (issue #40).

`replay()` is a thin branch:
```
replay():  out = packed_replay(...);  return out if out is not None else padding_replay(...)
```
- **`padding_replay`** = the dense `[B, P_max + T_max]` padded path (the original `replay` body, renamed — it pads).
- **`packed_replay`** = concat each sample's prompt+response into one zero-padded stream; restarting position_ids let transformers build the packed block-causal mask (verl `remove_padding` parity). `_replay_aware_forward` gathers per-token logp at the packed predict positions via a chunked fp32 lm_head (never materializes `[B, L, vocab]`).
- Under `flex_attention` the BlockMask skips fully-masked cross-sequence blocks; 1024-length bucketing amortizes flex's per-shape compile. Wiring: `config.attn_implementation` + `bundle` + recipe set `flex_attention`.

### Backend gate
Packed only pays with a sparse-block kernel. Pack only when `attn_implementation ∈ {flex_attention, flash_attention_2}` (skip cross-seq blocks); on plain sdpa packed attention is the full `O((ΣL)²)` and can regress (~1.36× equal-length), so it degrades to `padding_replay`. The `flash_attention_2` path activates once flash_attn is installed train-side (#54).

## Test Plan
- 16-GPU AR DRPO (Qwen3-4B, flex): the packed path was validated end-to-end in the perf-parity runs (ratio 1.0, curve matches baseline); the replay split is a behaviour-preserving refactor of that path.
- (Recommended before undraft) re-run the 5-step packed-vs-dense logp parity on the consolidated branch.

Supersedes #43 (folded in). Refs: #40, the Qwen-VL packed PR (#52), train-side flash_attn (#54).


leviking98z-rgb:
**Review — approve (no blocking issues).** Verified against transformers 5.6.0 source. The bucketing is correct by construction: filler is appended at the end, so causality alone shields every real position, and the restarting position_ids additionally give the filler its own document id via `find_packed_sequence_indices`, ANDed into the mask for flex and sdpa alike. `predict_index` is built before the append and the chunked lm_head only runs at gathered positions, so filler never touches logits or loss. Preconditions for HF packed detection (attention_mask=None, no KV cache) hold on the replay path.

Non-blocking: (1) bucketing runs even for sdpa configs — up to 1023 filler tokens with no compile benefit; consider gating on flex; (2) the knob is model-global and also governs the trainside per-token decode loop — fine for this SGLang recipe, but the config docstring should warn trainside-rollout recipes off it; (3) no graceful fallback when torch/transformers lack flex (load-time hard error; repo pins do not guarantee support); the B==1 dense path is un-bucketed so flex will shape-churn there; (4) "sdpa falls back to the math kernel" is imprecise (mem-efficient kernel; the real cost is the dense O(L^2) mask + no block skipping). Recommend a 1-step grad-norm smoke vs sdpa before merge given the flex + grad-ckpt + FSDP2 triple.

leviking98z-rgb:
Non-blocking review items addressed in 88b71a9: bucketing is now gated on _attn_implementation == "flex_attention" (sdpa configs no longer pay filler tokens). The decode-path doc warning and the B==1 bucketing nit are noted for a follow-up.

leviking98z-rgb:
Folded in the qwen-vl packed-varlen replay from #52 and addressed the review:

- **Dropped the `UNIRL_DISABLE_PACKED_REPLAY` env kill switch** (no env-var toggles).
- **Consolidated the packed prerequisites** into one predicate `_packed_replay_supported(attn_impl)` (sparse-block attn backend + transformers>=4.53 packed-mask) and **hoisted the gate up to the `replay()` dispatcher** — unsupported backends (sdpa / old transformers) warn once and fall straight to `padding_replay` without entering `packed_replay` or doing any setup.
- **Extracted the gate to a shared module** `unirl/models/_packed_replay.py` so qwen3 and qwen-vl reuse one copy; qwen-vl additionally gains the transformers feature-detect it was missing.

Verified: AST clean on all three files; `_packed_replay_supported('sdpa')=False (warns)`, `'flex_attention'/'flash_attention_2'=True`; no dangling refs to the removed names.

PR #45, opened 2026-06-12
perf(trainer): driver-side DP seqlen balancing (verl balance_batch parity) (issue #40)

leviking98z-rgb:

## Summary
FSDP collectives synchronize all ranks every micro-batch, so a step runs at the SLOWEST rank pace; the rank that drew the longest sequences straggles. verl ships this as `trainer.balance_batch` (default True). This adds the equivalent for the AR trainer: after advantages are attached (sample order is then free — the same invariant verl relies on), the driver reorders the rollout batch with greedy LPT into dp_size equal-count buckets by response token sum, then DP_SCATTER chunks contiguously. The reorder is a native `track.select(perm)` — no hydration: building on #59's TensorRef segment-views, `Batch.select` reindexes the remote packed data as lazy range views (`select_ranges`), so the per-rollout tensors stay on the workers and only the permutation crosses to the driver.

Knob: `balance_dp_batch` (default false).

## Benchmark
Ablation (16-GPU, matched window steps 6-15, run `abl_v11a_balance` vs `perfphase16_unirl`): train phase **28.4s -> 15.3s (-46%)**, whole step **78.1s -> 64.0s**; rank token spread drops to 0.6-0.9% (log line per step). This reaches train-phase parity with the verl reference (16.1s). The effect is large because GRPO groups dispatch as consecutive siblings — one rank draws all 8 long answers of a hard prompt and straggles every FSDP micro. Recipe default flipped to `balance_dp_batch: true`.

## Test Plan
- per-step log line reports rank token sums min/max/spread
- ratio_mean / loss parity vs unbalanced run
- 16-GPU ablation run `abl_v11a_balance` (wandb)

### DIFF
--- a/examples/ar/qwen3_drpo_4b_base_dapo_sglang.yaml
+++ b/examples/ar/qwen3_drpo_4b_base_dapo_sglang.yaml
@@ -37,6 +37,11 @@ adv_normalization_scope: group
 # false = mean-center-only advantage (reward - group_mean), no std division —
 # removes the GRPO difficulty bias (reference norm_adv_by_std_in_grpo=False).
 normalize_adv_by_std: false
+# Reorder the rollout batch driver-side so each DP rank gets a similar token
+# total (verl trainer.balance_batch parity). GRPO groups are dispatched as
+# consecutive siblings, so without this one rank draws a hard prompt's 8 long
+# answers and straggles the whole FSDP step. Measured: train 28.4s -> 15.3s.
+balance_dp_batch: true
 
 logging:
   report_to_wandb: true
--- a/examples/ar/qwen3_drpo_4b_base_dapo_sglang.yaml
+++ b/examples/ar/qwen3_drpo_4b_base_dapo_sglang.yaml
@@ -37,11 +37,11 @@ adv_normalization_scope: group
 # false = mean-center-only advantage (reward - group_mean), no std division —
 # removes the GRPO difficulty bias (reference norm_adv_by_std_in_grpo=False).
 normalize_adv_by_std: false
-# Reorder the rollout batch driver-side so each DP rank gets a similar token
+# Reorder the rollout batch driver-side so each DP shard gets a similar token
 # total (verl trainer.balance_batch parity). GRPO groups are dispatched as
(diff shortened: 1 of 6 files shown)


leviking98z-rgb:
**Review — approve with non-blocking changes.** The hydrate walk is provably complete (a superset of `_collect_leaves`, which creates every TensorMeta and never recurses tuples); in-place field mutation preserves the segment `_packed_cu_seqlens`, and `select` rebuilds cu correctly. The ragged right-pad matches `TextTokenCondition.concat` semantics, and the only mask consumer (`Qwen3ARStage.replay`) is mask-driven and re-pads right->left itself, so pad value/width are immaterial. Order-freedom holds (advantages pre-attached; no group-shaped worker ops). LPT equal-count bucketing is correct and deterministic.

Requested (non-blocking):
1. Cheap skip guard before hydrating: compute current-order chunk sums from `lens` (pure Python) and return unhydrated when spread < ~5%. On uniform-length workloads (VL geo3k: 28.2 -> 37.8s/step) hydrate+reorder is pure cost; this makes default-on safe to generalize.
2. Use the stack handle dp_size instead of `num_devices` — TP/SP or fraction<1 placements silently mis-balance.
3. Hardening: assert no TensorMeta survives the walk; check `len(lens) == batch_size`; log the `total % dp` skip.

Nits: within-rank descending-length order biases update 0 toward the longest samples when num_updates_per_batch>1; per-ref `.local()` could batch; `%%` typo; no unit test for the bucketing.

leviking98z-rgb:
Review suggestions landed in 5d8b973: pure-Python current-spread skip guard (<5% returns the track unhydrated — resolves the uniform-length overhead measured on VL geo3k) plus lengths/batch-size sanity. dp_size source and post-hydrate assertion noted for follow-up.

leviking98z-rgb:
Root-cause follow-up: the hydration workaround flagged in review ("hydrate_track is a bridge") is now removed by TensorMeta segment views — see the stacked PR above this branch. balance_track_for_dp permutes via native track.select(perm) with zero driver-side data motion.

celve (approved):

PR #48, opened 2026-06-12
perf(recipe): enable token-budget packing + DP balancing for Qwen-VL recipes

leviking98z-rgb:

## Summary
The TrainStack token-budget packing (#42) and the driver-side DP seqlen balancing (#45) are model-agnostic; this PR enables both for the Qwen-VL geo3k recipe. The VL replay is rectangular `[B, P_max + T_max]` and already strips the prompt block to the per-micro max, so the dense 2D cost model applies directly (VL prompt input_ids are processor-expanded, i.e. the cost accounting includes vision tokens).

NOT enabled for VL: the packed-varlen/flex path (#43/#44) — Qwen2.5-VL M-RoPE uses 3D position ids, which the transformers packed-sequence detection (1D position diffs) cannot interpret; porting needs an explicit BlockMask + VL-aware rope indexing.

Stacked on #42 + #45 (their commits are included; will rebase once they land).

## Benchmark
VL ablations in progress (16-GPU, Qwen2.5-VL-7B + geo3k_mc, 5-step windows): base vs +packing vs +packing+balance — numbers will be posted to #40 and here before undraft.

## Test Plan
- runs `vl_base` / `vl_pack` / `vl_packbal` (wandb project unirl-grpo), phase times via perf/*_time_s
- ratio/loss parity across the three configs

### DIFF
--- a/examples/ar/qwen3_drpo_4b_base_dapo_sglang.yaml
+++ b/examples/ar/qwen3_drpo_4b_base_dapo_sglang.yaml
@@ -188,6 +188,11 @@ sync:
 stack:
   _target_: unirl.train.stack.TrainStack
   micro_batch_size: 1
+  # verl ppo_max_token_len_per_gpu parity: pack each mini-batch into micro-batches
+  # under a token budget (length-sorted bin packing) instead of fixed-count
+  # slicing; with mbs=1 every sequence ran its own forward/backward. null keeps
+  # the legacy count-based behavior.
+  micro_token_budget: 10240
   max_grad_norm: 1.0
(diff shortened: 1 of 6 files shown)


leviking98z-rgb:
Closing per the VLM ablation results posted in #40: token-budget packing is memory-capped at the existing mbs=2 on Qwen2.5-VL-7B (OOM at budget 10240 and 6144), and DP balancing is negative on geo3k_mc (uniform ~2-token responses, hydrate/reorder overhead 28.2s -> 37.8s). The #42/#45 knobs stay available for variable-length VL workloads; nothing in the current VL recipe should change.

PR #49, opened 2026-06-12
feat(transport): TensorMeta ref views (HandleView) — selection without hydration (issue #40)

leviking98z-rgb:
Part of the verl performance-parity series tracked in #40. Root-cause follow-up to #45.

## Summary
#45's DP balancing needed `hydrate_track`: a driver-side full hydration of every TensorMeta field before `Batch.select`, because selection had no representation on remote refs (`TensorMeta.select` raised; `_slice_by_refs` only cut at ref boundaries). That workaround broke the transport's zero-copy premise (worker → driver → worker bounce), mutated frozen dataclasses in place, left field types history-dependent, and buried a padding convention in a private helper.

This PR adds the missing primitive — selection over remote refs with **zero driver-side data motion** — in two commits that show the design converging:

1. `0dc87a9` — segment views as a side-band `view_plan` on `TensorMeta` (first iteration).
2. `8c7b4b1` — the view moves down to the **ref layer**: `HandleView(base, start, end)`, a unit-range view of a parent ref. This is the final design; `view_plan` and all its special-casing are deleted.

### Final design (`HandleView`)
- `select` / `select_units` / `select_segments` emit `HandleView` refs — a viewed meta is structurally identical to a plain one (`refs + sizes`), so every transport treats refs uniformly (no viewed/plain branches anywhere).
- **Exact-row routing**: `localize` ships only a view's `[start:end)` rows cross-device (`gpu_store` and `colocate` `_nccl_send` slice by the routing copy's range) — a permuted shard no longer drags whole foreign blocks.
- **GC by construction**: a view holds a Python reference to its parent ref; CPython refcounting aggregates all views onto the parent's single finalizer. No per-view finalizer, no extra decref RPCs.
- `cat_rows` is the single assembly funnel. Trailing-dim CONTRACT documented there: parts from refs padded to different widths are right-padded with zeros (the `TextTokenCondition.concat` convention) — consumers of 2D+ per-shard-padded fields must be mask-driven.
- Boundary-aligned selections pass the original ref objects through untouched — the structural inverse of concat stays the zero-cost case; misaligned `slice` wraps only the boundary refs instead of raising.
- Nested views flatten at construction (no indirection chains).
- `balance_track_for_dp` permutes via native `track.select(perm)`: data stays worker-resident and materializes on the destination worker. `hydrate_track` remains a utility but is off the balance path.

## Test Plan
- `tests/test_tensormeta_views.py`: 10 CPU unit tests — permutation + ragged right-pad, aligned pass-through, misaligned wrap, packed token segments, nested-view flattening, `with_refs` size preservation, empty selection, `HandleView` shape/local, `cat_rows` contract.
- 16-GPU e2e gate (`viewbal_e2e`, Qwen3-4B DRPO + balance on, L2 iteration): 5 steps, `ratio_mean` 0.9995–1.0004, rank token spread 0.06%, rewards nominal.
- 8-GPU paired L2↔L3 ablation on both trainers (same node, back-to-back; details in the PR comment): ARTrainer (geo3k VL trainside, balance on) train 11.13 vs 11.76 s, rollout 22.70 vs 23.57 s; DiffusionTrainer (SD3 trainside, colocate backend) step 111.7 vs 112.7 s — no regression, deltas within noise; ratio 1.0000 on both.

### DIFF
--- a/examples/ar/qwen3_drpo_4b_base_dapo_sglang.yaml
+++ b/examples/ar/qwen3_drpo_4b_base_dapo_sglang.yaml
@@ -37,6 +37,11 @@ adv_normalization_scope: group
 # false = mean-center-only advantage (reward - group_mean), no std division —
 # removes the GRPO difficulty bias (reference norm_adv_by_std_in_grpo=False).
 normalize_adv_by_std: false
+# Reorder the rollout batch driver-side so each DP rank gets a similar token
+# total (verl trainer.balance_batch parity). GRPO groups are dispatched as
+# consecutive siblings, so without this one rank draws a hard prompt's 8 long
+# answers and straggles the whole FSDP step. Measured: train 28.4s -> 15.3s.
+balance_dp_batch: true
 
 logging:
   report_to_wandb: true
--- a/tests/test_tensormeta_views.py
+++ b/tests/test_tensormeta_views.py
@@ -1,13 +1,14 @@
-"""Unit tests for TensorMeta segment views (select/slice without hydration).
+"""Unit tests for TensorMeta ref views (select/slice without hydration).
 
-Pure-CPU: refs are faked with a minimal handle protocol (.local/.shape/.dtype/
-.device), no transport backend required (materialize falls back to per-handle
-fetch when backend is None).
+Selection emits :class:`HandleView` refs — unit-range views of the parent
+refs — instead of moving data. Pure-CPU: parent refs are faked with a minimal
+handle protocol (.local/.shape/.dtype/.device), no transport backend required
+(materialize falls back to per-ref fetch when backend is None).
 """
 
 import torch
 
-from unirl.distributed.tensor.transport import TensorMeta
+from unirl.distributed.tensor.transport import HandleView, TensorMeta, cat_rows
 
 
 class _FakeHandle:
@@ -38,7 +39,8 @@ def test_select_permutation_with_ragged_pad():
     tm = _meta(t0, t1, t2)
     perm = [5, 0, 3, 6, 2, 1, 4]
     v = tm.select(perm)
-    assert v.view_plan is not None and v.batch_size == 7
+    assert v.batch_size == 7
+    assert any(isinstance(r, HandleView) for r in v.refs)
     out = v.materialize(backend=None)
     assert out.shape == (7, 6)  # ragged refs right-padded to the max width
     assert torch.equal(out[0, :5], t2[0])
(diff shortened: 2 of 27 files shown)


leviking98z-rgb:
## Update: the view moved from the meta layer down to the ref layer (`HandleView`)

The first iteration kept selection lazy via a side-band `view_plan` on `TensorMeta`. That worked, but carried three structural smells: every transport needed plan-aware branches (3× `get_batch` + `hydrate`), `with_refs` carried a hidden plan-preservation contract, and `localize` still shipped **whole** source blocks even when a view needed a few rows.

Commit 8c7b4b1 dissolves all three by re-homing the view at the ref layer:

- **`HandleView(base, start, end)`** — a unit-range view of a parent ref. `select`/`slice` emit these, so a viewed meta is structurally identical to a plain one (`refs + sizes`); every consumer treats refs uniformly. `view_plan` and all its special-casing are deleted.
- **Exact-row routing** — `localize` now ships only a view's `[start:end)` rows cross-device (`gpu_store` + `colocate` `_nccl_send` slice by the routing copy's range).
- **GC by construction** — a view holds a Python reference to its parent ref, so CPython refcounting aggregates all views onto the parent's single finalizer. No per-view finalizer, no extra decref RPCs, no new GC machinery.
- `cat_rows` is the single assembly funnel (the documented ragged right-pad contract, unchanged).
- Boundary-aligned selections pass the original refs through untouched — the structural inverse of concat is preserved as the zero-cost case.

### Overhead quantification

Driver-side microbench at production scale (512 samples × 16 refs, scattered LPT permutation): `select` 0.47 ms/field, 4× `slice(128)` 0.28 ms, pickled viewed meta 13 KB/field, 56 B per view object (no finalizer). All negligible against 60–110 s steps.

E2E paired ablation, same node (8× H20), back-to-back, L2 = 0dc87a9 vs L3 = 8c7b4b1, 4 rollouts each, means over steps 2–4 (step 1 = warmup, excluded):

**ARTrainer** — `ar/qwen_vl_grpo_geo3k_mc` trainside, `+balance_dp_batch=true` (exercises select/permute + cross-device localize), wandb `5833mjw8` / `1t351xmc`:

| phase (s) | L2 | L3 | Δ |
|---|---|---|---|
| rollout | 23.57 | 22.70 | −3.7% |
| · generate | 10.23 | 9.95 | −2.7% |
| · reward | 1.52 | 1.57 | +3.3% |
| train | 11.76 | 11.13 | −5.4% |

**DiffusionTrainer** — `diffusion/sd3_trainside` + PickScore (`colocate_store` backend, exercises the aligned-slice pass-through path); this recipe does not emit `perf/*` to wandb, so the comparison is log-timestamp step time:

| | L2 | L3 | Δ |
|---|---|---|---|
| step time (s, mean of steps 2–4) | 112.7 | 111.7 | −0.9% |

Deltas are within run-to-run noise (AR generation lengths vary; DiT steps are deterministic to ~±1%): **no regression on either trainer** — so the deferred-batched-decref fallback considered for GC amplification is not needed; refcount traffic is unchanged by construction.

Correctness: ratio 1.0000±0.0000 on first updates for both trainers, rewards match across revisions (AR ~0.55–0.60, DiT ~0.75–0.77), balance rank token-sums logged tight; unit tests 10/10 (permutation + ragged pad, aligned pass-through, misaligned wrap, packed segment ranges, nested-view flattening, empty selection, `cat_rows` contract).

PR #50, opened 2026-06-13
fix(wandb): restore per-rollout train/* step axis (regressed in #24)

leviking98z-rgb:
## Summary

Makes the AR (and any multi-update) wandb logging emit **one data point per step** with all of a rollout's metrics co-located, so a run overlays cleanly against per-step loggers (e.g. verl) on the default Step axis. Two complementary commits:

**1. Fold per-update train metrics into one row (train/* anchored to rollout).**
Multi-update recipes (`per_update` length N>1, e.g. DRPO/GRPO with 4 updates) used to scatter the N updates as N separate `train/*` points along an optimizer-step counter, so `train/step` ran at ~`rollout x N`. Now the N updates fold into a SINGLE row, each namespaced `update{i}/<metric>` (cross-track: `update{i}/<track>/<metric>`), and `train/step` is anchored to `rollout_id + 1` — shared x-axis with `rollout/*`. `self._optimizer_step` is removed (it was only this x-axis).

**2. Merge the per-rollout log calls into one `_step`.**
Even after (1), each rollout still issued ~3 separate `wandb.log` calls — rollout-stats, train, perf — so wandb auto-incremented its internal `_step` ~3x per rollout: reward landed at `_step` 1, train at 2, perf at 3, … The metrics were therefore NOT co-located in a single history row, and the run stretched ~3x along the default **Step** axis versus a per-step logger, breaking cross-run overlay. Passing the explicit logical step to `wandb.log` in `log_with_step` makes all of a rollout's calls (and the rollout-aligned eval, all at `step == rollout_id + 1`) land in ONE history row at `_step == rollout`.

The two parts are interdependent: commit 2 is only safe because commit 1 folded the N train updates into a single call — otherwise the N same-step train `wandb.log` calls would overwrite each other. All step keys (`rollout/step`, `train/step`, `eval/step` via `log_eval(rollout_id+1)`) share the `rollout_id+1` scale, so `_step` stays monotonic.

### Tradeoff

Lost: the per-update points no longer occupy their own x-axis positions (the on-policy `update0` → off-policy `updateN` drift is now the `update0..N/<metric>` series within each rollout row, not a strung-out curve). Gained: one point per step, all rollout metrics in one row, and clean cross-run overlay on the default axis — the goal for the perf / algorithm-alignment comparisons (#40). An alternative (keep per-update points but anchor to a fractional rollout x-axis) was rejected: not isomorphic to single-row baselines.

## Test Plan

- `yaml`/AST parse + offline logic check (monkeypatched `wandb.log`): single-track 4-update → one row, keys `train/update0..3/<metric>`; multi-track mergeable → `update{i}/<track>/<metric>`; counter trajectory preserved where applicable.
- **Live verification** on a 32-GPU AR GRPO run (UniRL_perf, same logger): raw wandb history shows exactly one row per rollout —
  ```
  _step=1  rollout/step=1  train/step=1  | reward + train(update0..3) + perf  (one row)
  _step=2  rollout/step=2  train/step=2  | reward + train(update0..3) + perf
  ```
  `_step == rollout` (1:1), all metrics co-located, vs the pre-fix `_step ≈ 3x rollout` with reward/train/perf in separate rows.

Relates to #40.

### DIFF
--- a/unirl/utils/wandb_logger.py
+++ b/unirl/utils/wandb_logger.py
@@ -259,7 +259,11 @@ class UniRLWandBLogger:
 
     @property
     def optimizer_step(self) -> int:
-        """Current ``train/`` step-axis value — checkpointed for resume."""
+        """Cumulative optimizer-update count — checkpointed for resume.
+
+        No longer the ``train/`` x-axis (that is now rollout-anchored); this is
+        the running optimizer-update total kept only for resume bookkeeping.
+        """
         return self._optimizer_step
 
     def _handle_init_failure(
@@ -660,7 +664,7 @@ class UniRLWandBLogger:
             rollout_metrics.update(extra_metrics)
         self.log_rollout(step, rollout_metrics)
 
-        self._log_train(results)
+        self._log_train(results, step)
 
         perf: Dict[str, float] = {}
         if step_time_s is not None:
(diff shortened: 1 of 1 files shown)


haonan3:
I'm going to close this on a **diffusion-priority** call, and want the reasoning on the record.

  ### Why this conflicts with diffusion

  The diffusion recipes run one rollout with N>1 inner optimizer updates against a frozen π_old anchor, and we specifically inspect the **per-update importance ratio**:
 `update0` is on-policy (ratio ≈ 1, sampled with the trainable weights), `update1…N-1` drift off 1. Today that is a single `train/ratio_mean` series carrying N points per rollout — the on/off-policy drift read straight off one curve.


---

  The cross-run overlay goal doesn't actually require rewriting the internal `_step`; it needs a shared x-axis metric and a one-time x-axis selection in wandb. UniRL is
  already set up for this:
  
  - `rollout/*`, `perf/*`, `sync/*`, `buffer/*` are all logged **co-located with `rollout/step`** (= `rollout_id + 1`), and declared via
  `define_metric(step_metric="rollout/step")`. So the framework-comparable metrics (reward / advantage / perf / eval) already live on the rollout axis — it's just not the
  *default* `Step` axis.
  
  **Put any panel on the rollout axis (UniRL-only, zero code):**
  - Per panel: open the panel editor → the **X-axis** selector → choose `rollout/step` instead of `Step`.
  - Whole workspace: **Workspace settings → Line plots → X axis** → set the default to `rollout/step`. Every line plot then aligns by rollout count, with no `_step`
  inflation visible.

PR #52, opened 2026-06-13
feat(qwen-vl): packed-varlen replay (packed_replay / padding_replay) (issue #40)

leviking98z-rgb:
## Summary
Adds packed-varlen replay to **Qwen-VL** (verl `remove_padding` parity for M-RoPE). Independent of #42/#43/#44 (touches only `qwen_vl/ar.py`), so it branches off `main`. Part of the verl-parity series (issue #40).

`replay()` is a thin branch — `packed_replay` preferred, else `padding_replay`:
- **`padding_replay`** = the existing dense `[B, max_real + T_max]` path, renamed (it pads).
- **`packed_replay`** = concat each sample's prompt+response into one zero-padded stream with per-stream 4-D M-RoPE positions `[text; t; h; w]` from `get_rope_index`. Text row restarts per stream → transformers builds the packed block-causal mask under **sdpa** (Qwen2.5-VL has no flex) — no monkey-patch.

### Backend-aware gate
- `flex_attention` / `flash_attention_2` (skip cross-seq blocks) → always pack.
- `sdpa`/eager (full `O((ΣL)²)`) → length-variance gate: pack only when packed tokens ≤ 0.75× dense slots.

Qwen-VL gains `flash_attention_2` once flash_attn is installed train-side (#54) → gate no-ops, packed wins unconditionally.

## Test Plan
Real Qwen2.5-VL-7B (sdpa, transformers 4.57.1):
- per-stream `get_rope_index` == dense per-row positions, bit-exact (with images).
- text-only / with-image packed vs dense logits: argmax 100%, no cross-stream leak.
- `packed_replay` vs `padding_replay` logp: max|d| = 0.02 (bf16 noise).
- wall-clock: variable-length 0.53× (packed), equal-length-long 1.36× (gated to padding).

Refs: #40, the Qwen3 packed-replay PR (#44), train-side flash_attn (#54).

### DIFF
--- a/unirl/models/qwen_vl/ar.py
+++ b/unirl/models/qwen_vl/ar.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import functools
+import logging
 from dataclasses import dataclass
 from dataclasses import field as dc_field
 from typing import Any, Dict, List, Optional, Tuple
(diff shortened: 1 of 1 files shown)


leviking98z-rgb:
Folded into #44 (perf/03-flex-attention) via merge + a shared-gate refactor (`unirl/models/_packed_replay.py`), so qwen3 and qwen-vl share one packed-replay support predicate instead of duplicating it. Closing in favour of #44.

PR #55, opened 2026-06-13
feat(qwen-vl): packed-varlen replay (packed_replay / padding_replay) (issue #40)

leviking98z-rgb:
## Summary
Adds packed-varlen replay to **Qwen-VL** (verl `remove_padding` parity for M-RoPE), mirroring Qwen3's `replay` shape. Part of the verl-parity series (issue #40); independent of #42–#44 (touches only `qwen_vl/ar.py`), so it branches off `main`.

`replay()` becomes a thin branch:
```
replay():  out = packed_replay(...);  return out if out is not None else padding_replay(...)
```
- **`padding_replay`** = the existing dense `[B, max_real + T_max]` path, renamed (more apt — it pads).
- **`packed_replay`** = concat each sample's prompt+response into one zero-padded stream with per-stream 4-D M-RoPE positions `[text; t; h; w]` from `get_rope_index`. The text row restarts per stream, so transformers builds the packed block-causal mask from it under **sdpa** (Qwen2.5-VL has no flex support) — **no monkey-patch**.

### Backend-aware gate (pack only when it's a win)
Packed is a win only with a sparse-block kernel. By `attn_implementation`:
- `flex_attention` / `flash_attention_2` (skip cross-seq blocks) → **always pack**.
- `sdpa`/eager (full `O((ΣL)²)`) → **length-variance gate**: pack only when packed tokens ≤ 0.75 × dense slots, else `padding_replay`.

Qwen-VL has no flex but gains `flash_attention_2` once flash_attn is installed train-side (issue: "Install flash_attn train-side") — then the gate no-ops and packed wins unconditionally.

## Test Plan
Real **Qwen2.5-VL-7B** (sdpa, transformers 4.57.1):
- per-stream `get_rope_index` == dense per-row positions, **bit-exact** (with images).
- text-only / with-image packed vs dense logits: argmax **100%**, no cross-stream leak.
- `packed_replay()` vs `padding_replay()` logp: **max|d| = 0.02** (bf16 noise; leading tokens bit-identical); `replay()` routes to packed for B>1.
- wall-clock (7B): variable-length **0.53×** (packed), equal-length-long **1.36×** (now gated to padding).

Refs: #40, the Qwen3 replay split PR (#52), the train-side flash_attn issue.

### DIFF
--- a/unirl/models/qwen_vl/ar.py
+++ b/unirl/models/qwen_vl/ar.py
@@ -87,6 +87,13 @@ def _merge_igt(per_sample_igt: Optional[List[Optional[torch.Tensor]]]) -> Option
     return torch.cat(parts, dim=0) if parts else None
 
 
+# Attention backends with a sparse packed kernel (skip cross-sequence blocks) →
+# packed replay always wins. Qwen2.5-VL has no flex support, but gets
+# flash_attention_2 once flash_attn is installed train-side; on plain sdpa,
+# packed is gated on length variance (see packed_replay).
+_SPARSE_PACKED_ATTN = ("flex_attention", "flash_attention_2")
+
+
 class QwenVLARStage(ARStage[QwenVLARConditions]):
     def __init__(self, *, model: QwenVLBundle) -> None:
         self.model = model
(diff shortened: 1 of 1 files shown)


leviking98z-rgb:
Folded into #52 (VL packed replay now lives there, based on main). Closing as duplicate.

PR #56, opened 2026-06-13
refactor(train): token-budget micro-batching by composition (sort-then-slice)

leviking98z-rgb:
Refactor of the token-budget micro-batching from #42 — no behaviour change to the perf feature, just a cleaner design. **Stacked on #42** (base `perf/01-token-budget-packing`); rebase onto main once #42 lands.

## Summary
Two changes, both pure refactor:

**1. Composition instead of a ctor knob.** `TrainStack` now takes an injected `micro_planner` (default `CountPlanner` = the historical fixed-count behaviour). `TokenBudgetPlanner` carries `token_budget` / `cost_model` and owns the seq-mean loss precondition (`validate`, checked once at construction) — so count-based stacks no longer run a guard that only applies to packing. Recipes opt in with a `micro_planner` block; only `qwen3_drpo` does. Replaces the `micro_token_budget` / `micro_cost_model` ctor args.

**2. Sort-then-slice removes the index-plan invasiveness.** The planner reorders the track **once** up front (length-sort within each update, one `track.select`) so every micro is again a **contiguous `(start, end)` range** — exactly the count-based geometry. The stack therefore only ever slices, and the π_old anchor reassembly is a plain ordered `cat`. Gone: the `MicroPlan` range|index-list union, the slice/select branch, and the per-original-index anchor bookkeeping threaded through `prepare_segment`. All packing-specific logic is isolated to `MicroPlanner.arrange` (a no-op for `CountPlanner`). Reorder cost drops from one gather per packed micro to one gather per rollout.

Also drops the `replay_time_s` / `backward_time_s` timing hooks from `unirl/algorithms/drpo.py` — perf scaffolding that doesn't belong in the algorithm hot path (cross-cutting, DRPO-only).

Invariants preserved: update membership unchanged (each update's permuted block holds its original samples → same per-update gradient set-sum), sample-share loss weighting exact, NCCL micro-count parity.

Stays a single `unirl/train/stack.py` module — carving it into a `stack/` package is a further follow-up.

## Test Plan
- `tests/train/test_packing.py` (new, CPU-only): bin coverage/budget invariants, exact-K partition, sort-then-slice membership preservation, sample-share weighting (1e-12), seq-mean guard
- behaviour-equivalence: the packed plan trains the same sample set per update as before; `ratio_mean` unaffected (geometry identical to #42)

### DIFF
--- a/examples/ar/qwen3_drpo_4b_base_dapo_sglang.yaml
+++ b/examples/ar/qwen3_drpo_4b_base_dapo_sglang.yaml
@@ -188,11 +188,6 @@ sync:
 stack:
   _target_: unirl.train.stack.TrainStack
   micro_batch_size: 1
-  # verl ppo_max_token_len_per_gpu parity: pack each mini-batch into micro-batches
-  # under a token budget (length-sorted bin packing) instead of fixed-count
-  # slicing; with mbs=1 every sequence ran its own forward/backward. null keeps
-  # the legacy count-based behavior.
-  micro_token_budget: 10240
   max_grad_norm: 1.0
   # 4 disjoint mini-batch optimizer steps per rollout, 128 trajectories each —
   # matches the RELEASED verl run script (run_qwen3_4b.sh: ppo_mini_batch_size=16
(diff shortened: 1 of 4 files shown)


leviking98z-rgb:
Closing: the package split is not worth it. The planner strategies are just the OO face of the packing functions and `TrainStack` is their only consumer — they are bound, not a pluggable module others extend. Keeping it all in a single `unirl/train/stack.py` (the composition + sort-then-slice refactor in #42) is more honest than separating into a `stack/` package. No split.

leviking98z-rgb:
Reconsidered — the split IS worth it, but along the planning|driver seam (not planner|packing). Done in #57 (this PR couldn't be reopened after its branch was recreated).

PR #57, opened 2026-06-13
refactor(train): split stack.py into planning + base package

leviking98z-rgb:
Pure structural split of `unirl/train/stack.py` (the composition + sort-then-slice refactor from #42, now ~870 lines). **Stacked on #42** (base `perf/01-token-budget-packing`); rebase onto main once #42 lands. No behaviour change.

(Supersedes #56, which couldn't be reopened after its branch was recreated.)

## Summary
Split along the one real seam — the micro-batch **planning** vs the distributed **driver**, which have different dependency and test profiles:

| module | contents | deps | tests |
|---|---|---|---|
| `stack/planning.py` | `Plan` types, token-budget bin-packing math, `MicroPlanner` / `CountPlanner` / `TokenBudgetPlanner` | pure python (no FSDP/Ray) | CPU unit (`test_packing`) |
| `stack/base.py` | `TrainStack` driver | FSDP / Ray Remote | needs GPU |
| `stack/__init__.py` | re-exports | — | — |

The planner classes stay **with** the packing functions they wrap in `planning.py` — they are bound (a planner is just the OO face of the functions); the split only separates that whole planning unit from the driver, not the planner from the packing.

Public surface unchanged: `__init__` re-exports `TrainStack` / `TrainStepResult` / the planners / `_build_micro_batch_slices`, so recipes (`_target_: unirl.train.stack.TrainStack` / `...TokenBudgetPlanner`) and `from unirl.train.stack import ...` resolve exactly as before. `test_packing` imports the private helpers from the `planning` submodule.

## Test Plan
- no behaviour change — pure file reorganization; identical symbols, identical logic


leviking98z-rgb:
Folded into #42 — the package split now ships together with the composition + sort-then-slice refactor in one PR. Closing.

PR #58, opened 2026-06-14
feat(algorithms): add CPPO (position-weighted cumulative-prefix-budge…

chongqichuizi875:
## Summary

Add **CPPO** (Cumulative Prefix-divergence Policy Optimization, Binary-TV variant;
[arXiv:2606.10968](https://arxiv.org/abs/2606.10968)) as an AR token-level RL algorithm.

CPPO keeps DPPO's token-level ratio-advantage surrogate but replaces DPPO's *uniform*
per-token Binary-TV threshold with a **position-weighted threshold + cumulative prefix
budget** (paper Eq. 8-11). Only the trust-region mask changes; the loss term is DPPO's
`-A_t * r_t` on kept tokens. CPPO is the **hard-mask** sibling of the **smooth-mask**
`DRPO` over the same DPPO Binary-TV lineage, so it mirrors `unirl/algorithms/drpo.py`
exactly: same `StageAlgorithm` plumbing (`stage.replay` for `new_logp`, rollout-anchored
`old_logp = µ` via `old_logp_source="rollout"`, `GRPO._expand_advantages_to_tokens`,
`seq-mean-token-sum-norm` aggregation, `supports_multi_update=True`). The only algorithmic
difference from `DRPO` is the mask math.

The mask is built per packed sequence under `torch.no_grad` in fp32 (it is a trust-region
gate, not part of the loss). fp32 is required because `torch.quantile` rejects bf16/fp16 and
a bf16 `cumsum` over a long response loses the prefix sum to rounding; the recipe pins
fp32 logprobs, but the fp32 mask keeps CPPO correct under a bf16-logprob config too.

Changes:
- `unirl/algorithms/cppo.py` — `CPPO` + `CPPOConfig` + `_cppo_mask` / `_cppo_loss`
- `unirl/algorithms/__init__.py` — export `CPPO`, `CPPOConfig`
- `examples/ar/qwen3_cppo_30b_a3b_base_dapo_sglang.yaml` — Qwen3-30B-A3B-Base on DAPO-Math demo recipe
- `CPPO/README.md` + `CPPO/config.yaml` + `assets/cppo_overview.png` — tutorial (mirrors `DRPO/`)
- `README.md` — add CPPO to the Team-Proposed Algorithms table
- `scripts/check_recipe_targets.py` — scan the `CPPO/` tutorial dir for `_target_` resolution

## Related Issue

N/A

## Test Plan

- Lint / CI gate (all green): `SKIP=no-commit-to-branch pre-commit run --all-files`
  (ruff 0.15.1 check + format, the standard hooks, and the `check-recipe-targets` guard).
- Offline mask correctness: a standalone scalar-vs-vectorized check over 5 scenarios — all PASS:
  (1) vectorized mask == per-sequence scalar reference on a packed batch;
  (2) first-token effective threshold equals `δ`;
  (3) toward-µ updates are always kept;
  (4) once the prefix over-spends the budget, later diverging tokens are rejected;
  (5) with `w_min=1` and a non-binding budget the mask reduces to DPPO Binary-TV.
  Also exercised end-to-end through `_cppo_loss` in fp32 / bf16 / fp16 at 16k sequence length.
- Training / rollout smoke: **Not run; reason: no GPU / ray / sglang on the dev machine.**
  The recipe is a faithful config port to be launched on a training cluster.

## Compatibility / Risk

- Pure addition — no existing algorithm, config, or training path is modified.
- New config dataclass `CPPOConfig` with `cppo_delta` / `cppo_w_min` / `cppo_delta_b`;
  selected only by recipes that set `algorithm._target_: unirl.algorithms.cppo.CPPO`.
- The A3B demo recipe targets a sparse-MoE model (`Qwen3MoeForCausalLM`); its MoE/cluster
  knobs (`num_devices`, `tp_size`, FSDP wrap class, `mem_fraction_static`) are reasonable
  starting points to tune per cluster, not validated numbers, and are flagged as such in the
  recipe header.

## Reviewer Notes

- Implementation mirrors `unirl/algorithms/drpo.py`; reviewing CPPO against DRPO is the
  fastest path — only `_cppo_mask` / `_cppo_loss` differ.
- One intentional divergence from the reference verl implementation: UniRL is packed-varlen
  `[total_tokens]` (no padding), so the position weight and prefix cumsum are computed
  per sequence and `T` is each sequence's own length (the paper's per-sequence `T`), not a
  padded fixed-`T`. Documented in the code.
- AI assistance: this change was prepared with AI assistance; the diff and the offline
  validation were reviewed before submission. Duplicate-work check: no open PR or algorithm
  named CPPO exists in the repo.

## Checklist


zhouxiangxin1998 (approved):
Reviewed and confirmed

haonan3:
Pushed a small fix directly to this branch:

 In _cppo_mask the position weight used a (T - t) numerator, which collapses to w = w_min for a single-token response (T == 1, where T - 1 = 0). That contradicts paper Eq. 9 (the first token should be w_1 = 1) and the code's own comment (T == 1 -> w_t = 1). Switched to the paper's (t - 1) numerator:

  + w_t  = 1.0 - (1.0 - w_min) * (pos - 1) / max(T - 1, 1)

  Algebraically identical for T >= 2, and now correctly gives w_1 = 1 at T == 1. Pure corner-case/clarity fix (only length-1 responses, which don't occur at the 16k-token budget here); ruff clean. The rest of the mask matches the paper (Eq. 8 / 9 / 10 / 22), so merging.

PR #66, opened 2026-06-15
feat: add lora ckpt save & load

CjhHa1:
## Summary

Adds adapter-only LoRA checkpointing to the FSDP backend, plus the offline tooling to publish it.

- New `save_mode=auto`, now the default: resolves to `adapter` when LoRA is active, otherwise `full`. `full` and `adapter` remain valid explicit values.
- Adapter saves now gather **only** the LoRA tensors (`gather_lora_state_dict`) instead of gathering the whole state dict and filtering — no full-model materialization, so it scales to large / meta-init bundles.
- Meta-parameter checks are now mode-aware on save and load: adapter mode only requires the LoRA params to be materialized, full mode keeps the whole-model check. Load strictness is derived from the checkpoint's recorded `save_mode`, so adapter checkpoints resume on meta-initialized base models (e.g. hi3 80B) that previously tripped the unconditional meta check.
- The `_lora_meta` sidecar now records `rank, alpha, target_modules, dropout, bias, task_type`, enough to reconstruct a faithful PEFT `adapter_config.json`.
- New tool `unirl.tools.export_adapter`: turns a UniRL LoRA checkpoint into a standard PEFT adapter folder (`adapter_model.safetensors` + `adapter_config.json`). The merged-model exporter is renamed `export_hf` → `export_full` for clarity; `trainer/README.md` documents both export paths.
- Cleanup: removed dead helpers from `fsdp_utils.py` (`lora_state_dict`, `nft_state_dict`, `_is_lora_key`).

## Related Issue

N/A

## Test Plan

- Lint: `ruff check` and `ruff format --check` over the changed files — clean.
- Checkpoint round-trip / export: Not run; reason: no GPU/torch environment in this workspace. Suggested reproducible validation on a LoRA-enabled run:
  - Resume an adapter checkpoint: `bash examples/run_experiment_single_node.sh diffusion/sd3_trainside +save_interval=200 +save_dir=/ckpts/sd3_run +save_mode=adapter`, then rerun with `+load_dir=/ckpts/sd3_run/checkpoint-<step>`.
  - Export the adapter: `python -m unirl.tools.export_adapter --checkpoint /ckpts/sd3_run/checkpoint-<step> --base stabilityai/stable-diffusion-3.5-medium --output /ckpts/sd3_run/adapter-<step>`.

## Compatibility / Risk

- Default change: `save_mode` flips `full` → `auto`. LoRA runs now write adapter-only checkpoints by default (MBs instead of GBs; the frozen base reloads from the pretrained snapshot on resume). Set `save_mode=full` to keep whole-model checkpoints. Non-LoRA runs are unaffected (`auto` → `full`).
- Checkpoint loading is backward compatible: existing `full` checkpoints have no recorded `save_mode`, default to strict load, and resume unchanged. Adapter checkpoints load non-strict (LoRA keys only).
- Tool rename: `unirl.tools.export_hf` → `unirl.tools.export_full`. External callers of `export_hf` must update; in-repo references and docs are updated here.

## Reviewer Notes

- AI-assisted change; the full diff was reviewed line-by-line and lint-verified before pushing.
- No unit tests are included in this PR — the new save/load paths are covered by lint + manual reasoning only; flagging in case reviewers want coverage added before merge.
- `export_adapter` relies on the extended `_lora_meta` fields; checkpoints written before this change record fewer fields, so exporting a pre-existing checkpoint may need `--lora-alpha` / a manual config.

## Checklist

- [x] I updated docs/configs where needed; no unit tests are included (see Reviewer Notes).


celve (approved):

haonan3:
The core idea is right: an adapter save should only ever touch the adapter tensors and never materialize the frozen base.

But, I don't think this works yet in the case it's meant for (LoRA on a meta-init base like hi3 80B). The PR only specializes the save path (gather_lora_state_dict, so it skips the meta base). Load and the optimizer are unchanged — they still go through full-model DCP (set_model_state_dict / get_optimizer_state_dict), and the _reject_meta_params guard that used to block those on meta-init was loosened. So an adapter checkpoint written on 80B will most likely fail to load back on 80B — the exact problem the save side was rewritten to avoid, just moved to load time.

haonan3 (changes requested):
Please check the comments.

Load and the optimizer are unchanged — they still go through full-model DCP (set_model_state_dict / get_optimizer_state_dict), and the _reject_meta_params guard that used to block those on meta-init was loosened.

zzhuoxin1508 (approved):
LGTM.

Core design is right: adapter save only touches adapter tensors, never materializes the frozen base, so LoRA-on-meta-init (hi3 80B) can save and resume.

Verified load/optimizer won't choke on meta — optimizer is requires_grad-filtered (LoRA only), and load uses set_model_state_dict(strict=False), writing only LoRA keys and skipping meta aux. Backward compatible, dead code removed, and validated on sd3 and hi3/hy3.

haonan3 (approved):

Issue #54, opened 2026-06-13
Install flash_attn train-side → unified varlen packed replay (Qwen3 + Qwen-VL)


leviking98z-rgb:
### Problem
Packed-varlen replay (concatenating each sample's `prompt + response` into one zero-padded stream for the log-prob recompute) is only a speedup when the attention kernel **skips cross-sequence blocks**. Otherwise packed attention is the full `O((ΣL)²)` and *regresses* on low-variance batches.

Today the two AR stages are split:
- **Qwen3** uses `flex_attention` (BlockMask skips masked blocks) → packed wins, but pays a per-shape compile cost worked around by 1024-length bucketing (#44).
- **Qwen-VL** has **no flex support** (`Qwen2_5_VLForConditionalGeneration._supports_flex_attn = False`), so it runs packed on **sdpa**, which computes the full `L×L` and only masks.

Measured on Qwen2.5-VL-7B (sdpa, replay packed vs padded):

| batch shape | packed vs padded (fwd / fwd+bwd) |
|---|---|
| variable-length responses (real GRPO) | **0.53× / 0.49×** (≈2× faster) |
| equal-length, long (resp=1024) | 1.27× / 1.36× (**slower**) |
| equal-length, short (geo3k, resp=4) | 1.03× (neutral) |

PR #52 ships VL packed behind a length-variance gate (pack only when padding waste is large) so it is never a regression — but that leaves the equal-length-long case on the slower padded path, and keeps Qwen3 on the flex+bucketing complexity.

### Proposed fix — a dependency, likely not code
Both models declare `_supports_flash_attn_2 = True`, and transformers ≥4.53 ships a **native** `prepare_fa2_from_position_ids` (`modeling_flash_attention_utils`) that builds `cu_seqlens` from restarting packed position_ids and calls `flash_attn_varlen_func` — exactly verl's trick, but built in. verl only re-implements it (`verl/models/transformers/qwen2_vl.py` + `monkey_patch.py`) to layer Ulysses sequence-parallel on top; **we don't use SP, so the native FA2 path should suffice — no monkey-patch.**

Steps:
1. **Install `flash_attn` in the train venv(s)** — currently absent from both `unirl` and `diffusionrl` venvs. This is the real cost: a compiled CUDA extension, torch/CUDA-version-matched. (← the core decision this issue is about.)
2. Set `attn_implementation: flash_attention_2` for the AR train-side bundles (Qwen3 + Qwen-VL).
[...]

PR #76, opened 2026-06-15
feat(qwen-vl): packed-varlen replay + tf-5.x get_rope_index compat (#40)

haonan3:
## Summary

> **DRAFT — do not merge.** Blocked on a train-side flash backend (#54) and a
> packed-vs-dense logp parity re-run on the target stack. Opened so the Qwen-VL
> half of #44 is tracked separately from the verified Qwen3 half (#75).

Two commits:

1. **`feat`: packed-varlen replay (split from #44).** `packed_replay`
   concatenates each sample's `prompt + response` into one `[1, sum-L]` stream
   with per-stream 4-D M-RoPE positions (`[text_arange; get_rope_index t,h,w]`).
   The restarting text row lets transformers build the packed block-causal mask
   (sdpa) or derive `cu_seqlens` (flash). The gate requires a sparse-block
   backend; Qwen2.5-VL has **no flex_attention**, so the packed path needs a
   flash backend (`flash_attention_2/3/4`) and stays dense on sdpa.

2. **`fix`: `mm_token_type_ids` compat for `get_rope_index`.** transformers ≥ 5.x
   made `mm_token_type_ids` a **required** positional of
   `Qwen2_5_VLModel.get_rope_index`; ≤ 4.57 has no such parameter. Both the
   pre-existing dense `padding_replay` **and** the new `packed_replay` called it
   the old way → `TypeError` on a 5.x stack (this already breaks the dense geo3k
   replay path, independent of packing). `_vision_rope_positions` builds
   `mm_token_type_ids` from the config token ids and passes it only when the
   installed signature accepts it, so 4.57 and 5.x both work.

## Related Issue

Part of #40. Split from #44 (Qwen3 half is #75). Packed enablement depends on #54.

## Test Plan

- **Compat fix verified (transformers 5.8.1, CPU):** `_vision_rope_positions`
  runs `get_rope_index` on a real Qwen2.5-VL for both an image stream and a
  text-only stream (returns `[3, 1, n]`); the old call form raises `TypeError`.
- **Mechanism (shared with Qwen3, verified on 5.8.1):** packed `[1, sum-L]` +
  restarting `position_ids` + `attention_mask=None` → block-diagonal isolation
  (per-sequence hidden states reproduced to ~1e-6).
- **NOT yet run — the reason this is a draft:** end-to-end packed VL replay needs
  a train-side flash backend (#54; VL has no flex). Original #44 logp parity was
  on transformers 4.57; **must re-run packed-vs-dense logp parity on the 5.x +
  flash stack** before this leaves draft.

## Compatibility / Risk

- **Default-safe:** `attn_implementation` defaults to `None` → sdpa → dense
  `padding_replay`. geo3k recipes and existing VL behavior unchanged on a stack
  where `get_rope_index` already worked.
- **The compat fix repairs a current bug:** on transformers 5.x the dense VL
  replay path `TypeError`s today; commit 2 fixes that regardless of packing.
- Packed VL path is **unreachable** until a flash backend is installed
  train-side (#54), so it ships dormant.

## Reviewer Notes

- **Draft, intentionally.** Keep until #54 lands and the parity re-run passes.
- AI-assisted (split from #44 + compat fix + CPU verification with Claude Code);
  a human has reviewed the diff and the checks above.
- Duplicate-work check: this is the Qwen-VL half of #44; Qwen3 half is #75.
- Open question: the `mm_token_type_ids` compat (commit 2) is arguably a
  standalone bug fix for dense VL on transformers 5.x — if geo3k is actively
  trained on a 5.x pin, consider cherry-picking it ahead of #54.

## Checklist


leviking98z-rgb:
## GPU verification update — packed-vs-dense logp parity (incl. the flash kernel)

Ran an end-to-end `packed_replay` vs `padding_replay` logp parity check on a **randomly-initialized tiny Qwen2.5-VL** (head_dim=128, real `mrope_section`, fixed seed so both attention backends share identical weights). Same-weights → the comparison is purely "does packing change the per-token logp".

| config | seg_ok* | packed-vs-dense (max abs) | note |
|---|---|---|---|
| **CUDA / `flash_attention_4`** / text-only, B=3 | ✓ | **0.000** | production-target backend |
| **CUDA / `flash_attention_4`** / text-only, B=6, lens up to 150 | ✓ | **0.000** | scaled up, still bit-exact |
| CUDA / `sdpa` / text-only, B=3 | ✓ | 0.000 | |
| CUDA / `sdpa` / text-only, B=6 long | ✓ | 7.9e-3 | bf16 padding noise (see below) |
| **CPU / `sdpa` / WITH IMAGE**, B=3 | ✓ | 4.8e-7 (fp32) | image-scatter + per-stream M-RoPE |

\* `seg_ok` directly asserts `find_packed_sequence_indices(packed_pos[0])` segments the `[1,L]` stream from the **row-0 (text) line of the 4-D M-RoPE position_ids** into boundaries that exactly match the constructed `cu_seqlens`. True in every config — this was the main VL-specific unknown (does 4-D position packing detection behave like the 2-D Qwen3 case).

### What this validates
- **flash bit-exact (0.000) is the strongest possible evidence of zero cross-sequence leakage.** In the dense path each sample is its own batch row (isolated by construction); in packed everything is one `[1,L]` row and isolation depends *entirely* on `cu_seqlens` detection. Exact equality ⟹ packing reproduces per-row isolation exactly — on the real FA varlen kernel, which is the path this PR is gated on.
- The `sdpa` 7.9e-3 at scale is **expected bf16 noise**, not leakage: dense pads to max-len (150) while packed does not, so the reductions differ in bf16. Leakage would show up as larger, structured error concentrated on the first tokens of later streams.
- The **with-image path** (vision-embedding scatter into the packed layout + per-stream `get_rope_index` M-RoPE rows) matches to 4.8e-7 in fp32 on CPU.

### Relevant to the #54 blocker — flash is currently broken at the transformers integration level
On the only stack here that ships flash (transformers **5.6.0** + `flash-attn-4` 4.0.0b15), **every** flash variant (`flash_attention_2/3/4`) crashes on a plain Qwen2.5-VL forward before any packing is involved:

```
transformers/integrations/flash_attention.py:84, in flash_attention_forward
    s_aux=s_aux.to(query.dtype),  # FA only accepts half precision
AttributeError: 'NoneType' object has no attribute 'to'
```

`s_aux` (attention-sink logits) is `None` for Qwen2.5-VL, and the integration calls `.to()` on it unconditionally — a transformers-5.6.0 bug for any sink-less model on the flash path. I worked around it for the parity run by injecting a `-1e4` sink (`exp(-1e4)≈0` ⇒ no-op). **Implication: "train-side flash backend" (#54) isn't just `pip install flash-attn` — on this tf version the flash path for Qwen2.5-VL needs a transformers bump (5.10.x may fix it) or a patch.**

### Also confirmed
`attn_implementation="flex_attention"` raises `Qwen2_5_VLForConditionalGeneration does not support ... flex_attention` — so the PR's rationale (VL has no flex ⇒ packed must use a flash backend) is correct.

### Not covered
- **flash + image in the same process** — flash is CUDA-only and the *synthetic random* vision tower `munmap`s on CUDA here (a test-harness artifact, not PR code). It is the composition of two independently-verified halves: flash is image-agnostic (it sees an already-scattered, already-RoPE'd `[1,L,H]` stream and isolates purely by `cu_seqlens`), and the scatter+M-RoPE is verified on CPU. A real-weights GPU run would close this last cell.
- This is a tiny random model, not a trained checkpoint — it tests *mechanism/alignment*, not training dynamics. The behavioral logp-parity re-run on the real geo3k stack is still the right gate before leaving draft.

_(AI-assisted verification with Claude Code; tiny-model parity harness + transformers s_aux workaround. Numbers above are reproducible from the harness.)_

leviking98z-rgb (approved):

PR #79, opened 2026-06-16
feat(flux2_klein): condition image-edit generation on the source image + ckpt save

KemingWu:
## Summary

Rebases the Flux2Klein image-edit work onto the latest `main` (which already
contains EditReward via #28), carrying over only the Flux2Klein-specific
changes:

- **Source-image conditioning (image-edit path).** The Flux2Klein pipeline was
  pure text-to-image: `generate()` read `primitives["text"]` but never
  `primitives["image"]`, so for the EditReward image-editing job the source
  image never reached the transformer and outputs ignored the source. Adds the
  reference-image conditioning path mirroring diffusers `Flux2KleinPipeline`:
  a `Flux2KleinVAEEncodeStage` (encode → patchify → BN-normalize → pack to
  `[B,N,128]` tokens + 4-axis RoPE ids), `image_latent`/`image_latent_ids`
  slots on `Flux2KleinConditions` (round-tripped through `from_dict`/`to_dict`
  so replay conditions on the same source and the importance ratio stays
  honest), and `predict_noise` concatenating the condition tokens onto the
  packed noise sequence then slicing back. Pure T2I is unaffected (slots stay
  `None`).
- **LoRA checkpointing** in `flux2_klein_4b_editreward.yaml`
  (`save_interval`/`save_mode`/`save_dir`, env-overridable, default to
  CephFS so adapters survive pod teardown; `save_interval=0` keeps the
  framework default of no saving).
- **Side-by-side wandb preview** in `media_preview.py`: for image-edit (it2i)
  tasks the request carries an input image per sample, so the preview now
  builds an "input | output" PIL via `_hconcat_pil`. Falls back to output-only
  for t2i / when no input image is present and never crashes media logging.

## Related Issue

N/A

## Test Plan

  — all hooks pass (ruff check, ruff format, recipe `_target_` resolution, etc.).
- Training smoke test: not run in this PR; the conditioning path is being
  validated on the live EditReward Flux2Klein job
  (`flux2klein_4b_editreward`, 8×GPU) where wandb input|output previews confirm
  the generation now follows the source image.

Clipboard_Screenshot_1781590211

Clipboard_Screenshot_1781590234

## Compatibility / Risk

- Config change: `flux2_klein_4b_editreward.yaml` gains checkpoint keys; all are
  env-overridable and default to no-op-equivalent behavior, so existing runs are
  unaffected.
- Pure-T2I Flux2Klein generation is unchanged (conditioning slots stay `None`).
- No checkpoint/data-format/API changes.

## Reviewer Notes

- AI-assisted. The diff was reviewed and `pre-commit run --all-files` passes.
- Duplicate-work check: the source-image conditioning fix originally lived on
  `feat/editreward-support` (commit `a769ee4`); EditReward itself is already on
  `main` via #28, so this branch carries only the Flux2Klein-specific delta.
- The conditioning path mirrors the diffusers `Flux2KleinPipeline` reference
  flow (`prepare_image_latents` / `_prepare_image_ids` /
  `cat([latents, image_latents], dim=1)`); worth a sanity check against the
  installed diffusers version.

## Checklist

### DIFF
--- a/examples/diffusion/flux2_klein/flux2_klein_4b_editreward.yaml
+++ b/examples/diffusion/flux2_klein/flux2_klein_4b_editreward.yaml
@@ -19,6 +19,13 @@ num_devices: 8
 batch_size: 64
 num_rollouts: 10000
 
+# Checkpointing (LoRA adapter only — MBs, not GBs). Set SAVE_DIR to a persistent
+# location (e.g. CephFS) so checkpoints survive pod teardown; the default is a
+# repo-relative path. save_interval=0 disables saving (the framework default).
+save_interval: ${oc.env:SAVE_INTERVAL,100}
+save_mode: ${oc.env:SAVE_MODE,adapter}
+save_dir: ${oc.env:SAVE_DIR,checkpoints/flux2klein_4b_editreward}
+
 logging:
   report_to_wandb: true
   project_name: ${oc.env:WANDB_PROJECT,unirl-flux2klein-editreward}
--- a/unirl/models/flux2_klein/__init__.py
+++ b/unirl/models/flux2_klein/__init__.py
@@ -24,7 +24,7 @@ from unirl.models.flux2_klein.schedule import (
     build_flux2_klein_schedule_policy,
 )
 from unirl.models.flux2_klein.text_embed import Flux2KleinTextEmbedStage
-from unirl.models.flux2_klein.vae import Flux2KleinVAEDecodeStage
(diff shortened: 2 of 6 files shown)


unirl/types/media_preview.py
```diff
@@ -127,6 +127,26 @@ def _ref_aligned_prefix_len(decoded: Any, min_items: int) -> int:
     return total
```
CjhHa1, line 127: This change does not belong to this PR it is an additional enhancement, better drop it and move it to another PR

CjhHa1 (changes requested):

CjhHa1 (approved):
LGTM

PR #80, opened 2026-06-16
feat(pe): prompt-level (root) GRPO grouping for the diffusion track

Jayce-Ping:
## Summary

Adds a `diffusion_group_scope` knob to `PETrainer` that selects the diffusion track's
**GRPO grouping** — i.e. the advantage *baseline*, which in critic-free GRPO **is** the
training objective ("better/worse than what?").

- `"rewrite"` (default): group = the **M images of one rewrite** (group by rewrite id).
  Images are compared only to siblings with identical conditioning text; a whole rewrite
  that is systematically better/worse than its sibling rewrites gets ~0 net signal.
  Byte-identical to prior behavior.
- `"prompt"`: group = **all N×M images descended from one ORIGINAL prompt** (group by the
  root prompt id). A rewrite whose images beat the prompt-wide mean earns non-zero
  advantage, so diffusion learns to render well for the **original intent regardless of how
  the (frozen) rewriter phrased it** — robustness to rewording. Larger group ⇒ lower-variance
  baseline too.

Why this is coherent: the image reward is already scored against the **original prompt**
(not the rewrite) in `train_step`/`RewardService`, so all N×M images of a prompt share one
reward target. Grouping them is therefore an apples-to-apples comparison.

Mechanics (no lineage mutation):
- `RolloutTrack.compute_advantages` gains an optional `group_ids` override (default `None`
  → today's `parent_ids`). All reduction math, the group-by-parent contiguity guard, and
  keep-std normalization are unchanged; only the grouping labels differ.
- New `RolloutResp.compute_track_advantages(track, group_key="parent"|"root")`: `"root"`
  resolves each sample's root prompt id via the existing `_root_group_per_sample` and passes
  them as `group_ids`. A prompt's N×M images are already contiguous (`make_root_track` /
  `fork_track`), so `branch = N×M` reshapes cleanly.
- `PETrainer.train_step` routes the diffusion advantage through the root path when
  `diffusion_group_scope="prompt"`; the AR track is unaffected.

The frozen-LLM prompt-group recipe `pe_trainside_pickscore_frozenllm_promptgroup.yaml` sets
`diffusion_group_scope: prompt` — the **objective pairing** (fixed rewriter + prompt-level
grouping). Default stays `"rewrite"` everywhere else.

## Related Issue

N/A

## Test Plan

- Hydra config resolves:
  `python -m unirl.train_pe --config-name=pe/pe_trainside_pickscore_frozenllm_promptgroup --cfg job --resolve`
- Pre-commit: `SKIP=no-commit-to-branch pre-commit run --all-files --show-diff-on-failure`
- Unit (suggested, none in repo today): synthetic 2-prompt × 2-rewrite × 2-image resp;
  assert `group_key="root"` centers on the per-PROMPT mean (size-4 groups) and
  `group_key="parent"` reproduces the size-2 per-rewrite result; N=1 ⇒ root == rewrite.
- GPU smoke (maintainer): with `diffusion_group_scope: prompt`, confirm a rewrite whose
  images beat the prompt-mean gets positive advantage on ALL its images (cross-rewrite
  signal is non-zero), unlike the per-rewrite default.
- Not run in the authoring environment; reason: no GPU and `python`/lint/CLI are
  sandbox-restricted there. The commands above are for the maintainer.
- Verification done in lieu of execution: code review of the diff (root-label contiguity
  via a P=2,N=3,M=2 lineage trace; `group_ids=None` byte-identity; `scope="global"` ignores
  the override; trainer routing + freeze/prompt composition; backward-compat of all existing
  `compute_advantages` callers) — all four checks passed.

## Compatibility / Risk

- Default `diffusion_group_scope="rewrite"` → byte-identical to today for all existing PE
  runs. New behavior is opt-in.
- `group_ids` is a keyword arg with default `None` appended last; all 4 existing
  `compute_advantages` callers use named kwargs and are unaffected. `compute_track_advantages`
  is new (no prior callers, no name collision).
- **Trade-off (deliberate):** per-prompt grouping lets the frozen rewriter's quality
  variance leak into the diffusion advantage — a rich rephrasing's images get credit partly
  because the *text* was better, which diffusion can't act on. Per-rewrite grouping
  differences that out. For the robustness objective this is the right trade; it is why the
  feature is default-OFF and documented.
- No checkpoint / data-format / public-API change (additive ctor kwarg + new RolloutResp
  method + new grouping path).

## Reviewer Notes

- Rebased onto current `main` (which already includes the frozen-LLM `PETrainer` work,
  #78). The `diffusion_group_scope` ctor kwarg and `train_step` routing sit on top of #78's
  `freeze_llm` parameter; the two features are independent and compose (the objective recipe
  sets both `freeze_llm: true` and `diffusion_group_scope: prompt`).
- New recipe `pe_trainside_pickscore_frozenllm_promptgroup.yaml` is added alongside #78's
  `pe_trainside_pickscore_frozenllm.yaml` (distinct filenames, both retained) so the two
  recipes coexist.
- Draft: opened for early review; GPU smoke pending (see Test Plan).
- Duplicate-work check: no other open PR touches `RolloutTrack.compute_advantages` grouping
  or PE advantage scope (re-confirm against the live PR list at review time).

## Checklist


haonan3 (approved):
LGTM

Issue #25, opened 2026-06-10
UniRL Roadmap Q2 2026


celve:
# UniRL Roadmap Q2 2026

*Contributions and feedback are welcome.* 
Pick up any open item by opening a `[Tracking]`/`[RFC]` issue and linking it back here. UniRL post-trains **diffusion** (image / video), **autoregressive VLM / LLM**, and **unified AR+diffusion** models under one RL loop. Sections follow that loop — **rollout → reward → algorithm → training** — plus the **models** they run on and the **tooling** around them.

## Rollout

How samples are generated — engines, multimodal capture, cross-engine parity. *(`unirl/rollout/`)*

- VLM / LLM training efficiency: #40 
- Rollout engine refactors: 
  - #35 
  - #36 

## Reward

How samples are scored — the reward-model zoo, the standalone reward service, and the gradient path for REFL. *(`unirl/reward/`, `unirl-reward-service/`)*

- REFL differentiable rewards
- Image-editing rewards: #28 
- VLM / LLM-as-judge rewards: 
- More rewards

## Algorithm

How advantages are computed and the policy is updated. *(`unirl/algorithms/`)* — five implemented classes: `GRPO`, `FlowGRPO`, `DRPO`, `FlowDPPO`, `DiffusionNFT`. DanceGRPO and MixGRPO are `FlowGRPO` *recipes* (an SDE-kernel / sliding-window swap), not separate classes.

- **Reference-policy KL control**: [DPOK](https://arxiv.org/abs/2305.16381) / Diffusion-DPO / PPO.
- **REFL family (DR-Tune first)**:
  - [DR-Tune](https://arxiv.org/abs/2405.00760) (Deep Reward Supervisions, ECCV 2024) 
  - [ReFL](https://github.com/THUDM/ImageReward) (ImageReward, 2023) 
  - [DRaFT / DRaFT-LV](https://arxiv.org/abs/2309.17400) (2023) 
  - [AlignProp](https://arxiv.org/abs/2310.03739) (2023)
- **Policy-gradient / PPO: value critic + GAE**
- **Multi-track / shared-backbone RL**
## Training Backend & Distributed

The mechanics of the update — sharding, parallelism, checkpoint, weight sync. *(`unirl/train/`, `unirl/distributed/`)*
- Async RL Training for LLM (or even diffusion): #68 
- Parallelism (SP, TP): #71 
- Agentic: #93 

## Models

Route: *(`unirl/models/`)* 

- DiT models
[...]

PR #83, opened 2026-06-16
feat(train): add DCP sharded checkpoints

CjhHa1:
## Summary

Add opt-in PyTorch Distributed Checkpoint (DCP) sharded checkpoint support to `FSDPBackend`.

This keeps the legacy `checkpoint.pt` format as the default while allowing `checkpoint_format="dcp"` to save and load per-rank model and optimizer shards directly from the checkpoint root. The DCP path avoids rank-0 full-state gathering, supports reshard-aware resume, and handles meta-init bundles by omitting never-materialized meta entries.

## Related Issue

N/A

## Test Plan

Not run; reason: the available Python environment does not have `torch` installed, and the DCP/FSDP round-trip requires torch distributed support.

Validation performed:
- `ReadLints` on changed training files: no diagnostics.
- `git diff --check upstream/main...HEAD`: passed.
- Opus review subagent checked DCP API usage, cross-rank load branching, meta-init strictness, and fresh optimizer resume semantics.

## Compatibility / Risk

Default behavior remains `checkpoint_format="torch"`, so existing `checkpoint.pt` checkpoints continue to save and load as before.

DCP checkpoints use the checkpoint directory root as PyTorch DCP `checkpoint_id`, with DCP's `.metadata` at the root and UniRL training metadata in `metadata.pt`. `export_hf.py` does not yet read DCP checkpoints.

## Reviewer Notes

AI-assisted. Checked for overlapping public PRs/issues via web search because the installed `gh` command is not GitHub CLI. No direct overlapping DCP/FSDP sharded-checkpoint PR was found; roadmap issue #25 mentions checkpoint parity directionally.

## Checklist

### DIFF
--- a/unirl/train/backend/fsdp.py
+++ b/unirl/train/backend/fsdp.py
@@ -181,8 +181,7 @@ class FSDPBackend(Remote):
         checkpoint_format = str(getattr(fsdp_cfg, "checkpoint_format", "torch"))
         if checkpoint_format not in ("torch", "dcp"):
             raise ValueError(
-                "FSDPBackend: fsdp_cfg.checkpoint_format must be 'torch' or 'dcp', "
-                f"got {checkpoint_format!r}"
+                f"FSDPBackend: fsdp_cfg.checkpoint_format must be 'torch' or 'dcp', got {checkpoint_format!r}"
             )
         self._checkpoint_format: str = checkpoint_format
         self._checkpoint_async: bool = bool(getattr(fsdp_cfg, "checkpoint_async", False))
--- a/unirl/train/backend/fsdp.py
+++ b/unirl/train/backend/fsdp.py
@@ -511,6 +511,12 @@ class FSDPBackend(Remote):
         meta_path = os.path.join(path, "metadata.pt")
         meta: Dict[str, object] = torch.load(meta_path, map_location="cpu")
         mode = str(meta.get("save_mode", "full"))
+        # Symmetric with _save_dcp: a trainable param on meta would be dropped by
+        # drop_meta_entries below and then left unresolved by the non-strict load
+        # (has_meta_params relaxes strict), silently keeping its pre-load weights.
+        # Reject it here; the remaining meta is the frozen aux that legitimately
+        # relaxes strict.
+        self._reject_trainable_meta_params("load")
         has_meta_params = any(p.is_meta for p in self.model.parameters())
         strict = mode == "full" and not has_meta_params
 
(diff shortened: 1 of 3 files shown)


unirl/train/backend/fsdp.py
```diff
@@ -357,32 +385,94 @@ def save(self, path: str, step: Optional[int] = None, mode: str = "full") -> Non
...
         if missing_on:
             raise FileNotFoundError(
                 f"FSDPBackend.load: checkpoint not visible on rank(s) {missing_on}: {checkpoint_path} "
                 "(save_dir/load_dir must live on storage mounted on every node)"
             )
+        return self._load_torch(checkpoint_path)
+
+    def _load_torch(self, checkpoint_path: str) -> int:
+        """Legacy single-file load: read full state, broadcast from rank 0, reshard."""
+        self._reject_meta_params("load")
```
zzhuoxin1508: Suggest removing the unconditional _reject_meta_params("load") at the top of _load_torch:
Redundant for full mode: the if strict: branch below already calls it, so full mode calls it twice.
The if strict / else branch below already covers both cases, so removing this line restores the original behavior.

unirl/train/backend/fsdp.py
```diff
@@ -357,32 +385,94 @@ def save(self, path: str, step: Optional[int] = None, mode: str = "full") -> Non
...
+        Never gathers a full tensor on any single rank, so meta-init bundles
+        (whose frozen aux stays on meta) are supported — those keys carry no
+        data and are dropped here. Non-tensor metadata (step / save_mode /
+        lora_config / scheduler / optimizer_step_count) is light and rides in a
+        rank-0 ``metadata.pt`` beside DCP's own ``.metadata``.
+        """
+        import torch.distributed.checkpoint as dcp
+
+        os.makedirs(path, exist_ok=True)
+        model_sd = drop_meta_entries(sharded_model_state_dict(self.model))
```
zzhuoxin1508, line 404: _save_dcp doesn't seem to do the pre-save materialization check that _save_torch does (_reject_meta_params / _reject_lora_meta_params); instead it directly uses
  drop_meta_entries to drop all is_meta keys.

  The concern is that there may be two kinds of meta keys — expected frozen aux (should be dropped) and training params that could end up on meta due to a bug (should
  probably raise) — and the two can't easily be told apart by is_meta alone. So the latter might get silently dropped, potentially producing a checkpoint that's
  missing weights without any error

unirl/train/backend/fsdp.py
CjhHa1: you are right, fixed

unirl/train/backend/fsdp.py
CjhHa1: fixed

zzhuoxin1508:
LGTM

zzhuoxin1508 (approved):

PR #84, opened 2026-06-16
refactor(examples): group recipes into per-model folders and prune unread precision fields

zzhuoxin1508:
## Summary

Examples-only refactor in two parts:

1. **Reorganize** diffusion recipes into per-model folders (`examples/diffusion/<model>/`).
2. **Prune precision fields that are never read at runtime**, keeping only the live copies. No behavior change — every removed field was dead config.

## Why these fields are safe to remove

Precision here is a **stage-construction operator knob**, not a per-request value: the diffusion stage parses it once in `__init__` into `self.{autocast,trajectory,logprob}_dtype` (from the `pipeline:` block), and `diffuse()` / `replay()` use **only those** — never the per-request sampling params. The copies on other blocks were duplicates no code path consumes. Verified with grep/AST across the whole repo (incl. the vendored vLLM-Omni fork):

| Removed field | Block | Who reads it | Verdict |
|---|---|---|---|
| `trajectory_precision`, `logprob_precision` | `sampling:` | **nobody** (0 readers) | removed everywhere |
| `autocast_precision` | `sampling:` | only SGLang's `init_same_noise=True` self-noise path (dtype of the engine-generated Gaussian) | removed from trainside / vLLM-Omni (never hit that path); **kept** on SGLang |
| `autocast/trajectory/logprob_precision` | `bundle:` / `model_config:` | **nobody** — the bundle factory reads only `model_precision`; these are read solely by `Pipeline.from_config`, which is never called (the trainer builds the pipeline via `remote_hydra(pipeline_cfg, bundle=...)` → `__init__` / `from_bundle`) | removed, `model_precision` kept |

Kept on purpose (the live copies):

- **Every `pipeline:` block** — the authoritative values fed to the stage. Deleting these would fall back to `__init__` defaults (e.g. `trajectory_precision` → `fp16` while several recipes pin `bf16`), changing behavior.
- **`model_precision`** on every bundle — the weight-load dtype.
- **hi3 `bundle.config`** — its pipeline is `from_bundle(config: ${bundle.config})`, which **does** read those fields, so there the bundle copy is live.
- **bagel `bundle`** — dead, but the fields carry human rationale comments (`fp32` "do NOT lower" / "reward-collapse fix"); left intact.

## Related Issue

N/A

## Test Plan

- Re-parsed every touched config with PyYAML: all OK.
- Repo-wide audit (structural walk + grep): **0** remaining unread precision fields; all live copies intact.

## Compatibility / Risk

Config-only and behavior-preserving — every deleted field had no runtime reader. No checkpoint / API / data-format impact.

## Reviewer Notes

AI-assisted; no overlapping open PR. The reader analysis is the load-bearing claim — worth sanity-checking the `Pipeline.from_config` "never called" point and the SGLang `sampling.autocast` reader.

## Checklist


haonan3 (approved):
LGTM

PR #62, opened 2026-06-14
feat(bagel): BAGEL-7B-MoT multi-mode RL support (it2i/t2t/i2t/it2t/t2ti)

celve:
## Summary

Extends the BAGEL-7B-MoT bundle from **T2I-only** to all generation modes at the **rollout + replay** level — **it2i** (image editing), **t2t / i2t / it2t** (text-out via the und path), and **t2ti** (native think-then-generate). GRPO / rewards / trainer / recipes are out of scope; this adds the per-mode rollout + replay primitives the algorithms consume.

- **it2i** (editing): opt-in und ViT (`enable_vit`); input image prefilled through VAE+ViT; the vendor's 3-context editing CFG recipe. Diffusion stage unchanged.
- **t2t / i2t / it2t**: a new `BagelARStage` on the und path — per-token decode with KV cache (rollout) + one-shot teacher-forced `score_response` (replay), full-softmax logprob, `BagelARConditions` storing raw prompt splits, `BAGEL_UND_LORA_TARGETS`.
- **t2ti**: native think-then-generate — the AR und path plans a `<think>` caption, then diffusion renders conditioned on `[system + prompt + think]`; two linked tracks (`ar` + `image`, `image.parent_track="ar"`), routed via `ComposedSamplingParams`.

Design discipline: `vendor/` stays pristine; all RL glue lives in `rl_ops` via the `__wrapped__` no_grad bypass; `import unirl.models.bagel` stays flash-attn-free; navit bs=1 per-sample loops. Also fixes the `bagel_trainside_lora` recipe `old_logp_source` (`native` → `rollout`; `native` was never a valid FlowGRPO value).

## Related Issue

N/A

## Test Plan

Validated on real `ByteDance-Seed/BAGEL-7B-MoT` weights (8×H20) with an out-of-tree harness (the smoke/CPU scripts are dev-only and intentionally not shipped in this PR):

- **Modality smoke (GPU, all 6 modes) — 7/7 checks PASS**: t2t/i2t/it2t text + a greedy cross-check that `rl_ops.decode_text` emits byte-identical ids to the vendored `generate_text`; t2i/it2i image (diffusion replay ratio 0.00, it2i bit-equal determinism); t2ti (think + image, AR replay median ~2e-5, diffusion ratio 0.00, lineage `parent_track="ar"`). Visual QC of the saved images confirms on-prompt content (red panda for t2i/t2ti; it2i preserves structure and applies the requested edit at an adequate step count).
[...]

PR #88, opened 2026-06-17
feat(bagel): it2i editreward RL — recipe + VAE decode fix + diffusion eval

heguangxin:
## Summary

Enables BAGEL-7B-MoT **image-edit (it2i) editreward** RL end to end, on top of #62. Three pieces:

1. **Recipe** `examples/diffusion/bagel/bagel_editreward.yaml` — text + source image → edited image, scored by a remote EditReward service. Extends the T2I `bagel_trainside_lora.yaml` with the it2i enablement (`enable_vit`, edit pipeline) from #62.
2. **`fix(bagel)` VAE decode dtype** — `BagelVAEDecodeStage.decode` cast the shared VAE to fp32 in place. The it2i path reuses the same VAE to **encode** the source on the next rollout, so a leftover fp32 cast would poison that encode. Snapshot + restore the loaded dtype around decode.
3. **Diffusion-trainer periodic eval** — mirrors the AR trainer's `evaluate()`: on a cadence, generate on the eval set at a deterministic setting, score, and log `eval/reward` (eval prompts are chunked to bound driver memory). Off by default (`eval_interval=0`).

## Related Issue

N/A — but **depends on #62** (it provides the it2i model side this recipe needs). See Reviewer Notes.

## Test Plan

Live it2i editreward run on top of #62 — the recipe **as shipped** (BAGEL-7B-MoT, 8×H20, single node; remote MiMo EditReward; `cfg_text=1` single-forward, `eta=1.0`, group=8):

- **No OOM** through generate + replay + backward across 225+ rollouts; periodic eval + LoRA checkpointing both working.
- **Train reward** climbs −1.04 → ≈−0.35 (EMA); **eval reward** climbs −0.93 → **−0.40 **:

reward_curves

- A few sampled rollout edits (source | output) — object addition / background change:

edit_add_object
edit_background

- Hydra compose check:
  `python -m unirl.train_diffusion --config-name diffusion/bagel/bagel_editreward --cfg job --resolve`
- `eval_interval=0` (default) path unchanged.

## Compatibility / Risk

- **Additive.** Periodic eval is off by default (`eval_interval=0`) → existing diffusion runs are byte-for-byte unchanged. `_build_req` gains an optional `base_sampling` (default `None` = the training path, unchanged).
- The VAE fix is a **no-op for T2I** (T2I only decodes); it only affects the it2i encode-after-decode path.
- The new recipe references `enable_vit` from #62 → **depends on #62**. No file overlap with #62 (this touches only `vae.py`, the diffusion trainer, and the new recipe), so it rebases onto `main` cleanly once #62 lands.
- No checkpoint / data-format / public-API changes. Checkpointing reuses the existing `maybe_save_checkpoint`.

## Reviewer Notes

- **Merge after #62.**
- The recipe ships the **validated single-node config**: `cfg_text=1` (No-CFG, single forward — single-node 7B is memory-tight), `eta=1.0` SDE exploration, group=8. `cfg_text>1` enables CFG (+1 forward/step, more memory) and was not exercised in this run.
- MiMo EditReward is a Bradley-Terry-style head → a **raw, often-negative logit**; only group-relative differences matter for GRPO.
- The periodic-eval addition mirrors the AR trainer's `evaluate()`; happy to split it into its own PR if you'd prefer to review it separately.

## Checklist

### DIFF
--- a/examples/diffusion/bagel/bagel_editreward.yaml
+++ b/examples/diffusion/bagel/bagel_editreward.yaml
@@ -1,0 +1,200 @@
+# @package _global_
+# BAGEL-7B-MoT image-EDIT (it2i) GRPO — trainside + remote EditReward.
+#
+# Text + source image -> edited image: condition generation on (source image +
+# instruction), then score the edit remotely with EditReward. This is the T2I recipe
+# (bagel_trainside_lora.yaml) plus the it2i enablement from the BAGEL multi-mode bundle:
+#   - bundle.config.enable_vit: true  -> loads the SigLIP ViT/und path so the SOURCE
+#       image is ingested into the gen/cfg KV contexts (VAE latents + ViT semantics).
+#   - the pipeline auto-takes the it2i branch when the request carries an input image.
+#   - reward -> RemoteRewardBackend (HTTP to an EditReward service on its own GPU; it
+#       builds the [source, edited] pair from the request's input image — no BAGEL code).
+#   - sampling -> the defaults below reproduce the validated single-node run
+#       (cfg_text=1 single-forward; eta=1.0 SDE exploration; group=8; WindowScheduler).
+#       cfg_text>1 enables CFG (amplifies the instruction) at +1 forward/step (more memory).
+#
+# Deploy (keep EditReward OFF the training GPUs):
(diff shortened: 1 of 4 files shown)


haonan3 (approved):
LGTM

PR #89, opened 2026-06-17
feat: integrate BAGEL-7B-MoT into vLLM-Omni rollout engine

zzhuoxin1508:
## Summary

Integrate BAGEL-7B-MoT text-to-image into the vLLM-Omni rollout engine for GRPO,
with fp32-aligned SDE replay so the rollout↔trainside on-policy ratio stays ≈1.

Root-causes the "reward rises then regresses after ~100 rollouts" issue: it was
**not** a forward-precision mismatch but the per-step SDE noise `z_t` collapsing
across each GRPO group — the worker reseeds the global RNG to the same per-request
seed, so all group samples drew identical `z_t` (frozen exploration). Fixed by
drawing `z_t` from a per-request generator. Also aligns the worker forward to
trainside (fp32 RoPE + RMSNorm).

## Related Issue

N/A

## Test Plan

- LoRA GRPO run (rank 64, gen experts, fp32 master): reward 0.75→0.87,
  grad-norm flat (~1e-4), stable past 400 rollouts, no regression.
- On-policy parity: rollout↔replay logp ratio = 1.0000 (max-abs-diff ~1e-5).

## Compatibility / Risk

Additive — new BAGEL pipeline/adapter under the vLLM-Omni engine; existing
engines and recipes are unaffected.

## Reviewer Notes

AI-assisted.

## Checklist

- [x] Reviewed the changed code; removed diagnostic/generated artifacts.
- [x] Recipe `examples/diffusion/bagel/bagel_vllmomni.yaml` included.


heguangxin:
**The "reward regresses after ~100 rollouts" is the per-step SDE noise z_t, not the forward-precision mismatch.** I aligned the rollout↔trainside forward (RoPE→torch fp32, un-fused QKV, fp32 VAE) — it still regressed.

Cause: `pipeline_bagel.py` reseeds the global torch RNG to a fixed `sampling_params.seed` per request, and the SDE scheduler draws z_t from that global RNG (`generator=None`):

```python
if req.sampling_params.seed is not None:
    torch.manual_seed(req.sampling_params.seed); torch.cuda.manual_seed(req.sampling_params.seed)
```

Every GRPO-group sample is a separate bs=1 request reseeded to the **same** value → identical per-step z_t across the group → frozen exploration noise → breaks GRPO's sample independence → grad-norm spikes → reward regresses. (Trainside never reseeds per request, so its z_t is independent — that's why only the vLLM rollout regresses.)

**Fix in `bagel_flow_match_sde_scheduler.py` — draw z_t from a per-request generator instead of the global RNG.**

This is correct for `cfg_parallel_size=1` (the current recipe). Note `cfg_scale>1` alone is fine — when CFG runs on a single rank the branches are combined there and z_t is drawn once per step. **Only `cfg_parallel_size>1` needs care**: with CFG branches on separate ranks, `os.urandom` gives each rank a *different* z_t, so the branches diverge. There, draw z_t deterministically so all ranks agree — either seed by `(base_seed, group_id, step)` (every rank computes the same seed), or draw on rank 0 and `cfg_group.broadcast` it (mirroring the existing x_t broadcast).

Result (single variable = this fix): reward 0.75→0.87, grad-norm flat ~1e-4, stable past r400, no regression.

vomni_reward_curve

unirl/rollout/engine/vllm_omni/pipelines/bagel/bagel_flow_match_sde_scheduler.py
celve, line 1: It's not shared. Put it into bagel dir.

unirl/train/backend/fsdp/backend.py
```diff
@@ -70,6 +71,13 @@ def __init__(
         self._rank = int(rank)
         self._device = device if device is not None else torch.device("cuda" if torch.cuda.is_available() else "cpu")
 
+        # FSDP compute dtype (MixedPrecisionPolicy.param_dtype) = the wire dtype for
+        # weight sync. With master_dtype=fp32 the trainable LoRA params live in fp32
+        # (the reward-collapse fix), but the rollout engine's vLLM punica kernel
+        # hard-asserts bf16/fp16 — so LoRA extraction casts to this dtype at the
+        # all-gather (also halves sync bandwidth). Read via the ``wire_dtype`` property.
+        self._wire_dtype: torch.dtype = parse_torch_dtype(fsdp_cfg.param_dtype, field_name="training.fsdp.param_dtype")
```
celve: Maybe rename to `sync_dtype` or `weight_sync_dtype`

celve (changes requested):

unirl/rollout/engine/vllm_omni/pipelines/bagel/bagel_flow_match_sde_scheduler.py
zzhuoxin1508: done

unirl/train/backend/fsdp/backend.py
zzhuoxin1508: Done — renamed to weight_sync_dtype

celve (approved):

PR #91, opened 2026-06-17
feat(ltx2): add LTX-Video-2 model bundle + T2V trainside recipes

KemingWu:
## Summary

Add support for the LTX-Video-2 family (`Lightricks/LTX-2`) — a ~2.4B
audiovisual video DiT with flow matching and Gemma3 text encoding — as a
diffusion model that composes with the existing diffusion RL algorithms
(DanceGRPO / FlowGRPO) via the trainside rollout engine.

**New model subpackage `unirl/models/ltx2/`:**
- `bundle.py` — load the VAE (`AutoencoderKLLTX2Video`), the audiovisual
  transformer, the Gemma3 text encoder, and the `LTX2TextConnectors` projection.
- `config.py` — `LTX2PipelineConfig` + latent geometry constants (32× spatial /
  8× temporal compression, 128 latent channels).
- `pipeline.py` — trainside `generate`: regenerates `x_T` from the driver
  `NoiseRecipe` (5D noise → pack → diffuse → unpack → denormalize → VAE decode);
  exposes `latent_shape` and a `LTX2Pipeline.from_bundle` factory classmethod.
- `diffusion.py` — `predict_noise` drives the audiovisual transformer forward,
  co-denoising a real audio latent stream (`isolate_modalities=False`) so the
  trained audio→video cross-attention residual is preserved; `generate`/`replay`
  follow the framework SDE step contract and store the audio trajectory sparsely
  for correct log-prob replay.
- `schedule.py` — `LTX2SchedulePolicy`: dynamic exponential shift with constant
  μ = 2.05 and `shift_terminal = 0.1`, matching the checkpoint
  `scheduler_config.json` and diffusers' `image_seq_len`-pinned `calculate_shift`.
- `vae.py` / `text_embed.py` / `conditions.py` / `__init__.py`.

**Supporting infra (additive, default-off for other models):**
- `types/segments/latent.py` — `LatentSegment.aux_latents` (+ `aux_latents_at`)
  to carry the co-denoised audio trajectory for replay; defaults `None`, so
  every other model is unaffected.
- `trainer/diffusion.py` — resolve a pipeline `_target_` with `get_object` so a
  factory classmethod (`LTX2Pipeline.from_bundle`) works, not just a class.
  `get_class` rejected the bound-method form.

**Recipes `examples/diffusion/ltx2/`:**
- `ltx2_t2v_trainside.yaml` — T2V DanceGRPO trainside (1×8, LoRA r32, 10-step).
- `ltx2_t2v_trainside_20step_sde1.yaml` — 20-step / 1-SDE sampling variant.

The `shift_terminal` plumbing this work originally needed already landed on
`main` (the dynamic σ schedule supports it, and `FlowMatchSchedulePolicy` has
the field), so this PR reuses it and does **not** touch `unirl/sde/runtime.py`.

## Related Issue

N/A

## Test Plan

  — all hooks pass (ruff check, ruff format, recipe `_target_` resolution, etc.).
- Hydra compose-check for both recipes resolves cleanly (the recipe `_target_`
  pre-commit hook exercises this), including the `LTX2Pipeline.from_bundle`
  factory `_target_`.
- Training: validated on a live LTX-2 T2V DanceGRPO trainside job
  (`ltx2_t2v_dancegrpo_trainside`, 1×8 H20, LoRA r32, 33 frames). Reward rises
  and frames sharpen over training; sampling-budget tuning (10 vs 20 steps) is
  the `*_20step_sde1` variant.

Clipboard_Screenshot_1781682297

Clipboard_Screenshot_1781682417

Clipboard_Screenshot_1781682440

## Compatibility / Risk

- Purely additive: a new model subpackage plus two opt-in, default-`None`
  extension points (`LatentSegment.aux_latents`, the `get_object` pipeline
  resolution which is a strict superset of the previous `get_class` behavior).
- No changes to existing models, checkpoints, data formats, or public APIs.
- `unirl/sde/runtime.py` is intentionally untouched (its `main` version already
  supports `shift_terminal`).

## Reviewer Notes

- AI-assisted. Full diff reviewed; `pre-commit run --all-files` passes.
- Scope was carefully separated from sibling work on the same source branch:
  this PR excludes all `flux2_klein` and `hunyuan_image3` changes and the HI3
  `stage_params` routing (LTX-2 does not use `stage_config`). Only the
  `get_object` hunk of `trainer/diffusion.py` is included.
- The audio co-denoise path (`predict_noise` + `aux_latents`) is the trickiest
  area — it co-processes a real audio latent so the video forward's audio→video
  residual matches training. Worth a focused look.
- T2AV (LTX-2.3, audio output) recipes are intentionally **not** included here;
  the model code supports audio, but the T2AV reward setup is still being
  worked out separately.

## Checklist


unirl/models/ltx2/pipeline.py
```diff
@@ -0,0 +1,306 @@
...
+
+        # Determine mode
+        has_image = isinstance(images, Images)
+
+        # 1. Text embedding. CFG empty-negative: LTX-2's diffusers pipeline
+        # defaults negative_prompt to "" when guidance is on, so the model sees
+        # its trained unconditional embedding. Without this, predict_noise would
+        # skip the CFG branch entirely (negative_text is None) and guidance_scale
+        # would be a silent no-op.
+        negative_texts = req.primitives.get("negative_text")
```
CjhHa1, line 238: image conditioning is silent failed

unirl/models/ltx2/diffusion.py
```diff
@@ -0,0 +1,463 @@
...
+        autocast_ctx = (
+            torch.autocast("cuda", dtype=self.autocast_dtype) if self.autocast_dtype != torch.float32 else nullcontext()
+        )
+
+        with autocast_ctx:
+            for step_idx in target:
+                sigma = sigmas[step_idx].to(dtype=torch.float32)
+                sigma_next = sigmas[step_idx + 1].to(dtype=torch.float32)
+                sample = segment.latents_at(step_idx).to(device=device, dtype=self.autocast_dtype)
+                prev_sample = segment.latents_at(step_idx + 1).to(device=device, dtype=self.autocast_dtype)
```
CjhHa1: this auto cast may cause mismatch

unirl/models/ltx2/diffusion.py
```diff
@@ -0,0 +1,463 @@
...
+                    latent_num_frames=latent_t,
+                    latent_height=latent_h,
+                    latent_width=latent_w,
+                    audio_sample=a,
+                    audio_num_frames=audio_t,
+                )
+
+                # Video: SDE (RL) step. strategy.denoise →
+                # (prev_sample, log_prob, prev_sample_mean); log_prob None on ODE.
+                x_next, log_prob, _ = self.strategy.denoise(
```
Jayce-Ping, line 318: SDE is only applied on video modality. Should we also include audio?

xshrz (commented):
Reviewed the LTX-2 bundle + T2V trainside recipes. Overall well-structured and the T2V rollout → replay → GRPO skeleton mirrors WAN21 cleanly. I found **two must-fix issues** (one affects RL correctness), plus a few minor / discussion points.

## 🔴 Must fix

**1. rollout/replay model-input dtype mismatch → biased FlowGRPO ratio**

- `generate()` keeps `x` in `trajectory_dtype` (fp16) and feeds it directly to both `predict_noise` and `strategy.denoise` (`x = x_next.to(dtype=self.trajectory_dtype)`).
- `replay()` casts the sample/audio to `autocast_dtype` (bf16) **at the call site**:
  ```python
  sample      = segment.latents_at(step_idx).to(device=device, dtype=self.autocast_dtype)
  prev_sample = segment.latents_at(step_idx + 1).to(device=device, dtype=self.autocast_dtype)
  audio_sample = segment.aux_latents_at(step_idx).to(device=device, dtype=self.autocast_dtype)
  ```
- WAN21 (which this PR mirrors) keeps the two paths consistent: it casts the model input to the embeds dtype **inside** `predict_noise`, and does **not** cast at the replay call site — so `strategy.denoise` always sees the same (trajectory) dtype in both generate and replay.
- Effect: the step-0 importance ratio should be ≈1 but won't be (fp16 vs bf16 residual stream), injecting a systematic bias into the FlowGRPO gradient. Please align the dtype fed to the model/strategy across `generate` and `replay`. (matches @CjhHa1's "this auto cast may cause mismatch")

**2. I2V image conditioning is a silent no-op**

- `pipeline.generate()` sets `conditions.image_latent` when an image is present, but `diffusion.predict_noise()` never reads `conditions.image_latent` — there's no channel-concat and no image kwarg to the transformer. So I2V silently degrades to T2V.
- Only T2V recipes ship here, so it's not blocking, but please at least raise `NotImplementedError` on the `has_image` path (or actually wire the conditioning) to avoid a silent failure later. (matches @CjhHa1's "image conditioning is silent failed")

## 🟡 Discussion / minor

3. Audio is ODE-only (no SDE / log-prob). For pure T2V where audio is just the a2v-residual conditioning signal this is reasonable; if T2AV later optimizes audio it'll need SDE. Could you confirm the intent in a comment? (re: @Jayce-Ping)
4. video & audio share one `strategy` instance, calling `denoise` twice per step. Safe for the stateless `FlowSDEStrategy`, but a stateful multistep strategy (DPM) would have its `model_outputs` history corrupted by the interleaved audio call. A note/guard that LTX-2 requires a stateless SDE strategy would help.
5. The in-pipeline sigma fallback uses `config.shift = 1.0` (identity) when `req.sigmas is None`; this is only correct when the engine pins `req.sigmas`. Consider falling back to `build_ltx2_schedule_policy()` for self-containment.

## ✅ Looks good

- `LatentSegment.aux_latents` is a clean additive extension (CONCAT field, same kind as `latents`, default `None`; `aux_latents_at` mirrors `latents_at`) — other models are unaffected.
- `schedule.py` correctly reuses `FlowMatchSchedulePolicy` (overriding only `compute_mu` → constant `max_shift`), consistent with leaving `runtime.py` untouched.
- `trainer/diffusion.py` `get_class` → `get_object` is backward-compatible (recovers the owning class from a factory classmethod via `__self__`).
- replay requiring `aux_latents` and reusing the stored per-step audio so the video forward matches `generate` is the right call.

xshrz:
image

haonan3 (approved):

PR #100, opened 2026-06-19
feat(bagel): trainside it2t (understanding) GRPO RL on BagelARStage

heguangxin:
## Summary

Builds on the #62 BAGEL multi-mode RL primitives to make BAGEL-7B-MoT run **image+text -> text (understanding / it2t) GRPO RL**: it wires the `BagelARStage` rollout/replay into the trainside GRPO trainer and adds the pieces that path needs to run under FSDP/HSDP sharding.

- **Lockstep rollout** (`rl_ops.decode_text`): decode a fixed `max_new_tokens` forwards (no early EOS break). Under FSDP every forward triggers an all-gather, so a fixed step count keeps the collective schedule identical across ranks once the rollout runs sharded. See Reviewer Notes for the early-stop alternative.
- **Single-forward replay** (`rl_ops`): two replay scorers, each a single grad-capable forward to fit the FSDP parameter lifecycle (params reshard after each forward, so one grad forward per backward):
  - `score_response_with_prompt` — `no_grad` image prefill + one `forward_inference` over `[prompt + response]`, kernel-matched to the rollout (ratio ~1, image frozen).
  - `pack_und_forward_inputs` / `und_replay_logits` — one `forward_train` over the packed `[image | prompt | response]` (image also trained).
- `BagelARStage.replay_mode` (`"train"` | `"inference"`) picks which replay scorer computes `new_logp`; `BagelPipeline` passes it through.
- Recipe `examples/ar/bagel_grpo_arxivqa_mc_2x8_lora.yaml` (2x8 HSDP trainside LoRA on the und path) + `unirl/utils/prepare_arxivqa_mc.py` to build the arxivqa-MC dataset (same on-disk layout as `prepare_geo3k_mc`).

## Related Issue

N/A (builds on #62, already merged).

## Test Plan

> **Still running — numbers below are a mid-run snapshot (rollout 188/1000) and will be updated as the run continues.**

Live 2-node HSDP run of the shipped recipe (BAGEL-7B-MoT, 2x8 H20, b128 / g8 / T1.0, `replay_mode=train`, num_updates=2, und LoRA r64a128, clip 1e-2, lr 3e-5), ~4 min/rollout, stable throughout (no hangs/OOM); periodic full-val eval (166x8) + LoRA checkpoint every 20 rollouts both working.

**Train reward (EMA) ~0.67 -> ~0.74; eval accuracy 0.6084 -> ~0.70:**

pr_reward_curve

- **Hydra compose check** (no GPU): `PYTHONPATH=$PWD python -m unirl.train_ar --config-name ar/bagel_grpo_arxivqa_mc_2x8_lora --cfg job --resolve`

## Compatibility / Risk

- Additive: `replay_mode` defaults to `"train"`; the #62 rollout/replay primitives and the diffusion path are unchanged.
- `micro_batch_size` must be 1 on this AR path (replay scores one sample per grad forward); the recipe sets it and documents why.
- Requires FSDP `root_wrap: false` (the vendored BAGEL forward calls `embed_tokens` / `lm_head` / `forward_inference` outside a root forward). Documented in the recipe.

## Reviewer Notes

- Thanks to #62 for the `BagelARStage` rollout/replay primitives this builds on.
- **Fixed-length rollout decode, by design for now:** `rl_ops.decode_text` runs a fixed `max_new_tokens` forwards rather than breaking at EOS, so every FSDP rank stays on an identical all-gather schedule with zero extra collectives. This MC task uses a small `max_new_tokens` (8), so always decoding to the cap is cheap and the wasted tail is negligible. For long-output tasks (larger `max_new_tokens` / CoT), we plan to migrate to the rank-synchronized early-stop that the Qwen AR stages use — a per-step `dist.all_reduce(min)` on the done flag — which skips the long tail at the cost of one small collective per token.
- Drive-by: aligns the wandb logging block in `examples/diffusion/bagel/bagel_trainside_lora.yaml` with the repo default (`report_to_wandb: false`, `project_name: unirl`, `run_name/entity/tags: null`).

## Checklist


haonan3 (approved):
LGTM

PR #101, opened 2026-06-19
feat(rollout): WAN 2.1 T2V sglang_diffusion rollout adapter

leviking98z-rgb:
## Summary

Adds a **WAN 2.1 T2V video rollout path on the `sglang_diffusion` engine**, so the `wan21` family can roll out via SGLang instead of only the trainside FSDP sampler. This is the **first true video-output consumer** of the engine (image families + image-path video stubs existed; WAN was trainside-only).

## What's included

- **`adapters/video.py`** — new `VideoAdapter` base: 6-D `[B, T+1, C, F, H, W]` video-form trajectory (vs the image path's 5-D) and decoded media packed as a ragged `Videos` instead of dropped. `Wan21T2VAdapter` registered under `model_family=wan21`. `mochi`/`hunyuan_video` stay on the legacy image-path parity stubs (no behavior change).
- **`utils/tracks.py`** — `stack_decoded_videos`: packs each result's `[C, T, H, W]` decode into `Videos` with frame-major `[T, C, H, W]` (what `video_pickscore` consumes).
- **`_patches/patch_wan_scheduler.py`** — swaps `WanPipeline`'s hardcoded `FlowUniPCMultistepScheduler` (a multistep ODE solver lacking `SchedulerRLMixin`, so the GRPO single-step SDE rollout + driver-pinned σ schedule can't run on it) for `FlowMatchEulerDiscreteScheduler`. Mirrors how `wan_dmd_pipeline` already builds it; single-step flow-match also matches what the trainer replays.
- **`models/wan21/config.py`** — adds `use_lora` / `lora_target_modules` (mirrors `SD3PipelineConfig`) so the engine's `server_intent` can boot SGLang in LoRA mode for `wan21`.
- **`__init__.py`** — opt-in global cuDNN disable (`UNIRL_DISABLE_CUDNN=1`) to work around a cuda-compat-13 / driver-535 `munmap_chunk(): invalid pointer` crash in conv paths (sglang VAE decode **and** the reward models' CLIP convs, which run in unirl actors that `patch_vae_decode_safe` doesn't cover). No-op otherwise.
- **`examples/diffusion/wan21/wan21_t2v_sglang.yaml`** — 1×8 GRPO recipe (480×832 — the 1.3B native res; `num_frames=5`; `old_logp_source: replay`).

## Validation

1×8 on `Wan-AI/Wan2.1-T2V-1.3B-Diffusers`: full loop runs — **sglang rollout → video decode → `video_pickscore` reward → train step → LoRA weight-sync → wandb**. Rollout-1 reward is **bit-identical to the trainside `wan21_t2v` reference (0.6722)** under matched settings, confirming the sglang rollout is numerically consistent with trainside.

## Known caveat (not introduced by this PR)

`grad_norm ≈ 1e-5` at this config (advantages are healthy — `advantage_std ≈ 0.55`, `zero_std_group_ratio = 0`). The **trainside `wan21_t2v` reference shows the same `gn ≈ 0`**, so this is a pre-existing property of the WAN FlowGRPO recipe at these settings, not the rollout adapter. Flagging for a separate follow-up; the rollout path itself is verified.

## Draft

Opening as draft pending review of the scheduler swap and the cuDNN workaround, and the `gn≈0` recipe follow-up.

### DIFF
--- a/examples/diffusion/wan21/wan21_t2v_full.yaml
+++ b/examples/diffusion/wan21/wan21_t2v_full.yaml
@@ -1,0 +1,136 @@
+# @package _global_
+# GRPO WAN 2.1 T2V trainside, FULL fine-tuning (no LoRA). Curve-alignment twin
+# of wan21_t2v_sglang_full.yaml — IDENTICAL config except the rollout engine
+# (trainside FSDP sampler here vs SGLang there). Trains the whole transformer
+# (no lora_cfg); trainside uses the live model so there is no weight sync.
+
+num_devices: 16
+batch_size: 24
+adv_use_global_std: true
+num_rollouts: 60
+
+model_config:
+  _target_: unirl.models.wan21.config.WAN21PipelineConfig
+  pretrained_model_ckpt_path: ${oc.env:PRETRAINED_MODEL,Wan-AI/Wan2.1-T2V-1.3B-Diffusers}
+  model_precision: bf16
+  autocast_precision: bf16
+  trajectory_precision: fp16
+  logprob_precision: fp32
+  shift: 5.0
+  max_sequence_length: 512
+  use_lora: false
+
+strategy:
+  _target_: unirl.sde.kernels.FlowSDEStrategy
+
+bundle:
+  _target_: unirl.models.wan21.bundle.WAN21Bundle.from_config
+  config: ${model_config}
+
+pipeline:
+  _target_: unirl.models.wan21.pipeline.WAN21Pipeline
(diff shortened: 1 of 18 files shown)


leviking98z-rgb:
Superseded by the split: #102 (general weight-sync rename fix) + #103 (WAN 2.1 T2V sglang rollout, full-FT + LoRA, without the env-specific CPU-stage workaround). Closing in favor of those.

Issue #72, opened 2026-06-15
🤖 Code Audit: 14 potential issue(s) found


asmit25805:
##  Code Audit Report

> All findings are reviewed for confidence before posting.
> Please verify each finding before acting on it.

**Repository:** `Tencent-Hunyuan/UniRL`
**Findings:** 14 issue(s) found — 🟠 3 high · 🟡 7 medium · 🔵 4 low

---

### 1. 🐛 Premature return prevents fallback to earlier module splits

| Field | Details |
|---|---|
| **Severity** | 🟠 High |
| **Type** | Bug |
| **File** | `scripts/check_recipe_targets.py` |
| **Location** | _resolve |
| **Confidence** | 96% |

**Problem:**
The `_resolve` function returns immediately after finding the first existing module file, even if the first attribute after the module is not present in the module's top‑level names. This stops the loop from trying shorter module prefixes, causing false negatives for valid dotted paths where the attribute chain is deeper than one level.

**Suggested Fix:**
Replace the early `return` with a conditional check that continues the loop when the attribute is not found. Only return `True` when a matching attribute is confirmed; otherwise, let the loop iterate to try shorter module prefixes. Example:
```python
if names is not None and attr_parts[0] in names:
    return True
# else continue searching with a shorter module prefix
```
After the loop, return `False`.

---

### 2. 🐛 Incorrect absolute file path for test dataset

| Field | Details |
|---|---|
| **Severity** | 🟠 High |
| **Type** | Bug |
| **File** | `datasets/pickscore/prpocess.py` |
| **Location** | with open("/dataset/pickscore/test.txt", "w", encoding="utf-8") as file: |
| **Confidence** | 96% |

**Problem:**
The code writes the test split to "/dataset/pickscore/test.txt", which is an absolute path starting at the filesystem root. On most systems this directory does not exist or is not writable, causing a runtime error and preventing the test file from being saved.

**Suggested Fix:**
Change the path to a relative one consistent with the train file, e.g., "dataset/pickscore/test.txt", and ensure the target directory exists before writing.

---

### 3. 🐛 Assumes exactly one score result from client.score

| Field | Details |
|---|---|
| **Severity** | 🟠 High |
| **Type** | Bug |
| **File** | `unirl-reward-service/scripts/smoke_client.py` |
| **Location** | (result,) = client.score([req]) |
| **Confidence** | 91% |

**Problem:**
[...]

PR #106, opened 2026-06-22
fix(datasets): write pickscore test split to relative path

mvanhorn:
## Summary

The PickScore preprocessing script wrote the test split to an absolute path while the train split used a relative one. This makes the test write target the filesystem root instead of the working directory, so the script crashes before saving the test file.

This change makes the test output path relative and consistent with the train write, and creates the output directory up front so a fresh checkout does not fail on a missing directory.

## Related Issue

Refs #72 (finding #2 of the code audit: the absolute test-split path). The other audit findings are out of scope here.

## Test Plan

- `python -m py_compile datasets/pickscore/prpocess.py` (passes)
- `ruff check datasets/pickscore/prpocess.py` (All checks passed)
- `ruff format --check datasets/pickscore/prpocess.py` (already formatted)

Not run: the full preprocessing job, which downloads `yuvalkirstain/pickapic_v1` and needs the dataset to be available; reason: large dataset dependency.

## Compatibility / Risk

Output location for the test split changes from `/dataset/pickscore/test.txt` (filesystem root) to `dataset/pickscore/test.txt` (relative to the working directory), matching the train split. No API, config, or data format changes.

## Reviewer Notes

The absolute path on line 26 of `datasets/pickscore/prpocess.py` was inconsistent with the relative train write two lines above and would not be writable on a normal machine. AI assistance was used; the diff was reviewed line by line. Checked the open issue and PRs for overlapping work before submitting.

## Checklist

### DIFF
--- a/datasets/pickscore/prpocess.py
+++ b/datasets/pickscore/prpocess.py
@@ -20,11 +20,11 @@ unique_text_dataset = unique_dataset[:test_size]
 train_dataset = unique_dataset[test_size:]
 
 # Save the datasets with shuffling
-os.makedirs("dataset/pickscore", exist_ok=True)
-with open("dataset/pickscore/train.txt", "w", encoding="utf-8") as file:
+os.makedirs("datasets/pickscore", exist_ok=True)
+with open("datasets/pickscore/train.txt", "w", encoding="utf-8") as file:
     for line in train_dataset:
         file.write(line + "\n")
 
-with open("dataset/pickscore/test.txt", "w", encoding="utf-8") as file:
+with open("datasets/pickscore/test.txt", "w", encoding="utf-8") as file:
     for line in unique_text_dataset:
         file.write(line + "\n")
--- a/datasets/pickscore/prpocess.py
+++ b/datasets/pickscore/prpocess.py
@@ -1,3 +1,4 @@
+import os
 import random
 
 from datasets import load_dataset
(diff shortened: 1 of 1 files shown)


datasets/pickscore/prpocess.py
```diff
@@ -19,10 +20,11 @@
 train_dataset = unique_dataset[test_size:]
 
 # Save the datasets with shuffling
+os.makedirs("dataset/pickscore", exist_ok=True)
```
CjhHa1: better change to `datasets/`, align to the settings in this repo

datasets/pickscore/prpocess.py
```diff
@@ -19,10 +20,11 @@
...
 with open("dataset/pickscore/train.txt", "w", encoding="utf-8") as file:
     for line in train_dataset:
         file.write(line + "\n")
 
-with open("/dataset/pickscore/test.txt", "w", encoding="utf-8") as file:
+with open("dataset/pickscore/test.txt", "w", encoding="utf-8") as file:
```
CjhHa1: same as above

CjhHa1: done

datasets/pickscore/prpocess.py
CjhHa1: done

CjhHa1 (approved):

CjhHa1:
I have fixed the issues in your commit, and this will be merged @mvanhorn. Thanks for your contributing!

PR #113, opened 2026-06-23
feat: add WAN 2.2 V2V trainside support

xshrz:
## Summary
Adds trainside WAN 2.2 video-to-video support, including video condition loading, V2V latent preparation, WAN22 V2V pipeline/config, and example recipes for 14B trainside runs.
## Related Issue
N/A
## Test Plan
- ReadLints on changed Python files: no linter errors found.
- Verified required local WAN 2.2 and PickScore model caches are available in the training environment.
- Not run; reason: current host Python is too old for the project, and the Taiji training task was not reachable from this shell. Hydra compose and 8-GPU smoke should be run in the training container before marking ready.
## Compatibility / Risk
Adds support for `(video, condition)` media refs in `MultimodalRLDataSource`. V2V recipes require a manifest with one condition video per prompt and use WAN 2.2 T2V A14B weights.
## Reviewer Notes
Draft PR for early review. Main areas to inspect: `unirl/models/wan22_v2v/`, V2V schedule trimming, and `DiffusionSamplingParams.strength` plumbing.
## Checklist


unirl/models/wan22_v2v/pipeline.py
```diff
@@ -0,0 +1,212 @@
...
+                base_seed=int(params.seed),
+            )
+        if int(noise.shape[0]) != batch_size or tuple(noise.shape[1:]) != tuple(video_latents.shape[1:]):
+            raise ValueError(
+                f"WAN22V2VPipeline.generate: noise shape {tuple(noise.shape)} incompatible with "
+                f"video latents {tuple(video_latents.shape)}."
+            )
+
+        x_start = (1.0 - sigma_start) * video_latents + sigma_start * noise
+        sde_indices = [int(i) for i in (params.sde_indices or []) if int(i) < t_eff]
```
leviking98z-rgb: **`sde_indices` is in the full-schedule frame but is not remapped to the trimmed frame.** `params.sde_indices` is resolved against the *full* `num_inference_steps` (trainer `resolve_sde_indices` → `AllSDEScheduler(num_timesteps=num_inference_steps)`), so with `timestep_fraction=[0,0.2]` it is `⊂{0,1,2}`. V2V then trims the schedule to start at `t_start` (=3 for `strength=0.8`, `num_inference_steps=16`). This filter keeps `i < t_eff` **without subtracting `t_start`**, so full-frame indices {0,1,2} are reinterpreted as *trimmed*-frame indices {0,1,2}. That lands at the start of the trimmed schedule only because `timestep_fraction` starts at 0; for any fraction not starting at 0 (e.g. `[0.5,0.7]`) the SDE steps would land on the wrong steps. Resolve sde_indices against `t_eff` (decide the V2V trim *before* resolving), or subtract `t_start` — and reconcile with the WAN21 path, which does the opposite.

unirl/models/wan21/pipeline.py
```diff
@@ -251,5 +292,38 @@ def generate(self, req: RolloutReq) -> RolloutResp:
...
+        effective_steps = int(effective_schedule.shape[0]) - 1
+        if effective_steps < 1:
+            raise ValueError(
+                f"WAN21Pipeline.generate: V2V strength={strength} leaves no denoising steps "
+                f"from num_inference_steps={total_steps}."
+            )
+
+        sde_indices = None
+        if params.sde_indices is not None:
+            sde_indices = [int(i) - t_start for i in params.sde_indices if t_start <= int(i) < total_steps]
```
leviking98z-rgb: **This is the principled remap (`i - t_start`), but it's inconsistent with `WAN22V2VPipeline` and degenerates for the shipped recipes.** For `timestep_fraction=[0,0.2]` + `strength=0.8` (`t_start=3`), `{0,1,2}` all fail `t_start <= i`, so `sde_indices` becomes `[]` → **zero SDE steps → no RL gradient / empty replay log-probs**. Meanwhile `WAN22V2VPipeline` (the path the recipes actually use) keeps them. The two V2V implementations disagree here; this one is unexercised by any recipe and broken for the natural config. Root cause: `timestep_fraction` is resolved over the *full* `num_inference_steps` but V2V only denoises the trimmed tail — resolve SDE indices in the effective (trimmed) frame instead.

leviking98z-rgb (commented):
## Review (head 2cd9bcc) — my earlier review of a5f94a2 + the 3 new commits

The 3 new commits (OOM/data fix, edit-delta CLIP reward, ruff format) don't touch the two main issues — both still stand. New-code notes at the end.

### Main
1. **sde_indices frame handling is wrong + inconsistent between the two V2V pipelines** — see the two inline comments. Net: `WAN22V2VPipeline` (the one the recipes use) only "works" because `timestep_fraction` starts at 0; the principled `WAN21` remap drops *all* SDE steps for the shipped `[0,0.2]`+`strength=0.8` config. Fix: resolve SDE indices in the **effective (trimmed)** frame, and unify the two implementations.
2. **Two parallel V2V implementations, only one wired up.** Both `wan21/pipeline.py` (+`WAN21VideoLatentEncodeStage`) and the new `wan22_v2v/` package implement V2V; both recipes use only `WAN22V2VPipeline`. The WAN21 path is unexercised, untested, and (per #1) buggy. Drop it from this PR, or add a recipe + share one encode stage + one strength/sde helper.

### Minor
3. `data_source._load_condition_videos`: `torch.load(uri)` has no `weights_only=True` — an arbitrary-code-execution vector for untrusted manifests (low risk if data is internal).
4. **New edit-delta reward (`video_clip_delta.py`)**:
   - The two terms aren't on the same scale by construction: `text_align = logit_scale*cos/26` (~0.8) vs `source_sim` = raw image-image cosine (~0.9). `lambda_source=1.0` "nets ~0" only for this PickScore model's specific `logit_scale` — fragile if the model/scale changes.
   - Scores **only the first frame** of both edited and source → the edit reward is a single-frame proxy, ignoring temporal content.
   - Relies on `RewardService` `repeat_interleave`-ing `primitives['video']` to per-sample (docstring asserts it); load-bearing and unverified (PR is "Not run"). Guarded by a count check that raises if wrong.
5. `Videos.from_list` pads ragged H/W with zeros, then the encoder `F.interpolate`s the padded frame → black-border distortion for mixed-resolution condition videos in one batch (homogeneous batches unaffected). Prefer resize-to-target before packing.

### Still good
The flow-match add-noise math, strength→σ_start mapping, `(num_frames-1)%4` checks, and defensive shape guards are correct.

**Status**: still "Not run" — given #1, a smoke must verify SDE steps are non-empty and replay log-probs align.

leviking98z-rgb (approved):

xshrz:
image

NUM_NODES=4 GPUS_PER_NODE=8 \
bash examples/run_experiment_multinode_taiji.sh diffusion/wan22_v2v/wan22_v2v_14b

PR #119, opened 2026-06-24
build(deps): raise  bagel transformers dependency to 5.6 to match engine stacks 

CjhHa1:
## Summary

Raise the `transformers` floor `>=4.41` → `>=5.6,<5.7` across `pyproject.toml`,
`setup.py`, `requirements.txt`, and patch BAGEL's vendored Qwen2 code for the two
transformers-5.x construction differences found by smoke testing.

Both engine stacks already pin transformers 5.6.x (sglang 0.5.12.post1 →
`==5.6.0`, vllm 0.20.0 → resolves `5.6.2`), so the 4.41 floor was dead — every
engine install path already runs 5.6.x. BAGEL was the last 4.x-era holdout.

## Related Issue

N/A

## Test Plan

H20 pod (`unirl-v1d`), transformers upgraded 4.56 → 5.6.0 in place (torch 2.7.1 /
flash_attn 2.8.1 unchanged), staged BAGEL smoke using real
`BAGEL-7B-MoT` from `/apdcephfs/private_aimicahchen/models/BAGEL-7B-MoT`:

- **A versions** — PASS (`transformers 5.6.0`, `torch 2.7.1`, `flash_attn 2.8.1`, H20)
- **B import** — PASS (vendor modeling + bundle + pipeline import, after installing pre-existing missing `opencv-python-headless`)
- **C config** — PASS (`Qwen2Config.from_json_file` + `rope_config_validation`; expected tf5 FutureWarning only)
- **D load real 7B bundle** — PASS (`BagelBundle.from_config`, 14.17B params loaded on cuda; first cold CephFS load took ~1478s, later cached loads ~9s)
- **E direct vendored inferencer T2I** — initially reached final VAE decode then failed with dtype mismatch (`Input type (float) and bias type (c10::BFloat16) should be the same`). This PR now applies the same fp32 VAE decode policy used by `BagelVAEDecodeStage`; not re-run end-to-end after that small fix because the UniRL pipeline path below is the supported path.
- **E2 UniRL pipeline-path T2I** — PASS (`BagelPipeline.generate`, `num_inference_steps=2`, `256x256`, no SDE steps): `decoded pixels (1, 3, 256, 256) torch.float32`, `segment latents (1, 1, 256, 64) torch.float32`, `PIPELINE_SMOKE_PASS`.
- **F replay-backward smoke** — PASS (`BagelPipeline.generate` with `sde_indices=[0]`, `eta=1.0`, then `pipe.diffusion.replay(...).log_probs.mean().backward()` with one gen-expert parameter unfrozen): `replay.log_probs (1, 1) torch.float32 requires_grad True`; target `model.layers.0.self_attn.q_proj_moe_gen.weight`; grad dtype `bf16`, max abs `6.63e-05`, norm `9.47e-04`; `BACKWARD_SMOKE_PASS`.

Additional small construction smoke on transformers 5.6.0:

- Tiny `Qwen2ForCausalLM(Qwen2Config(... rope_type=default ...))` — PASS with `pad_token_id` absent (`padding_idx=None`) and default RoPE fallback (`inv_freq_shape=(8,)`).

## Compatibility / Risk

- The main package is intentionally capped to the tested transformers 5.6.x window (`>=5.6,<5.7`); engine extras already resolve/pin inside that window.
- BAGEL vendored modeling now preserves transformers 4 behavior when `pad_token_id` is absent (`None`) and keeps a local default RoPE initializer because transformers 5 no longer exposes `ROPE_INIT_FUNCTIONS["default"]`.
- `rope_config_validation` is still deprecated in transformers 5 but functional. Future cleanup can replace it with `standardize_rope_params()` / `validate_rope()` if we want to silence the warning.
- The vendored upstream inferencer is retained as a reference/debug path; UniRL's supported inference/training path is still `BagelPipeline`.
- Pre-existing runtime dep gap: `unirl-v1d` lacks `opencv-python-headless` and `diffusers`, both required by the BAGEL pipeline smoke.

## Reviewer Notes

AI-assisted; checked open PRs — none touch the transformers floor. UniRL pipeline forward and replay-backward pass on transformers 5.6. The direct vendored inferencer was kept usable with the same VAE dtype policy as the pipeline, but it is not the supported training path.

## Checklist


zzhuoxin1508 (approved):

heguangxin (approved):
Image

Ran a quick PickScore experiment on this branch. rollout reward_mean shows a stable upward trend over 100+ steps (~0.74 → ~0.81). LGTM.

PR #122, opened 2026-06-25
Adopt a unified Sample data structure for the rollout engine

celve:
## Summary

Introduces a unified `Sample`/`Part` type model and converts the rollout engine to operate on it end-to-end.

### Types (LIN-446)
- Add `Sample`/`Part` endomorphism types, with `sample_id` lineage derived from the sample_ids path.
- New: `unirl/types/sample.py`, `unirl/types/sample_id.py`.

### Rollout (LIN-454)
- Convert the rollout engine to a `Sample → Sample` interface across the `sglang`, `sglang_diffusion`, and `vllm_omni` adapters/engines.
- Add smoke scripts and a sample-roundtrip checker under `scripts/`.

## Commits
- `types: add Sample/Part endomorphism types with sample_id lineage (LIN-446)`
- `rollout: convert the rollout engine to Sample → Sample (LIN-454)`

## Diffstat
39 files changed, +2098 / -744.

---
_Opened as a draft for review of the Sample-based rollout interface._

PR #123, opened 2026-06-25
chore(gitignore): keep ignoring reward/local bytecode caches

CjhHa1:
## Summary

The force-include rules `!unirl/reward/local/` and `!unirl/reward/local/**` re-add
every file under that Python package, which overrides the global `__pycache__/`
ignore. As a result, compiled `.pyc` files under `unirl/reward/local/__pycache__/`
keep surfacing as untracked entries in `git status`.

This adds `unirl/reward/local/**/__pycache__/` right after the force-include so the
package sources stay tracked while bytecode caches are excluded again (gitignore
gives precedence to the later rule).

## Related Issue

N/A

## Test Plan

- `git check-ignore -v unirl/reward/local/__pycache__/base.cpython-312.pyc` → matched by the new rule (now ignored)
- `git check-ignore -v unirl/reward/local/__pycache__/` and a nested `.../sub/__pycache__/x.pyc` → both ignored
- `git check-ignore -v unirl/reward/local/base.py` and `.../__init__.py` → no match (sources still tracked)
- `git ls-files unirl/reward/local/` → package `.py` sources remain tracked
- `pre-commit run --files .gitignore` → all hooks pass

## Compatibility / Risk

N/A — ignore-only change. No source, config, checkpoint, data format, or runtime
behavior is affected. Already-tracked files are unchanged.

## Reviewer Notes

AI-assisted change. Checked open PRs (#89–#122); none touch `.gitignore`, so no
duplicate work.

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,8 +67,6 @@ logs/
 local/
 !unirl/reward/local/
 !unirl/reward/local/**
-# the force-include above re-adds bytecode caches; keep ignoring them
-unirl/reward/local/**/__pycache__/
 **/rollout_data/
 **/buffer_stats/
 *.out
@@ -104,3 +102,6 @@ settings.json
 !.cursor/rules/**
 REFACTOR_LOG.md
 uv.lock
+
+# Re-assert bytecode ignore after the force-includes above (last match wins)
+__pycache__/
--- a/.gitignore
+++ b/.gitignore
@@ -67,6 +67,8 @@ logs/
 local/
 !unirl/reward/local/
 !unirl/reward/local/**
+# the force-include above re-adds bytecode caches; keep ignoring them
+unirl/reward/local/**/__pycache__/
 **/rollout_data/
 **/buffer_stats/
 *.out


celve:
Works as written. One note: the same `__pycache__` override affects all four `!…/**` force-includes in this file (`reward/local`, `.claude/skills`, `.cursor/skills`, `.cursor/rules`), not just `reward/local`. A single rule at the end of the file covers them all:

```gitignore
__pycache__/
```

Last-match-wins re-asserts the global ignore after the force-includes. Fine to merge as-is — just simpler this way.

celve:
Pushed a follow-up commit that swaps the `reward/local`-scoped rule for a single `__pycache__/` at end-of-file, so the net diff is now just that one rule. It covers all four `!…/**` force-includes (current and future) via last-match-wins, rather than only `reward/local`.

Verified locally: bytecode caches at all four sites + nested + root are ignored, all 18 `reward/local` sources stay tracked, and `git ls-files --ignored --cached` is identical to `main` (no tracked file masked).

celve (approved):

PR #126, opened 2026-06-25
fix(ltx2): make audio x_T driver-authoritative (salted sibling of video recipe)

leviking98z-rgb:
## Summary

LTX-2's unified transformer always co-denoises an audio latent stream alongside video (the audio→video cross-attention residual is kept even for T2V). But while the **video** x_T is driver-authoritative — regenerated from the shipped `NoiseRecipe` (CPU-fp32, per-group seeded → reproducible, shard-safe, cross-engine-aligned, `init_same_noise`-aware) — the **audio** x_T was a bare global-RNG `torch.randn(...)` inside `LTX2DiffusionStage.generate()`. That audio init noise was:

- unseeded / non-reproducible and not cross-engine-aligned;
- drawn on the global CUDA generator, so it perturbed the video SDE noise stream;
- blind to `init_same_noise` / `noise_group_ids`, so grouped rollouts that share the video x_T did **not** share the audio x_T.

This makes the audio x_T a first-class sibling of the same recipe:

- **`NoiseRecipe.resolve(..., salt="", latent_shape=None)`** — new optional args. With a `salt`, `resolve()` forks an *independent but reproducible* sibling stream: group ids are salted (`{gid}::{salt}` → a different per-group seed, so it is **not** a prefix-alias of the base stream), any Path-1 `initial_latents` is bypassed, and `latent_shape` overrides the recipe's own. Default `salt=""` is byte-for-byte identical to before, so the **video path and every existing caller are unchanged**.
- **`LTX2Pipeline.generate`** resolves the audio x_T from the same video recipe — `video_recipe.resolve(salt="audio", latent_shape=audio_latent_shape(params))`, symmetric with the video line right above it — and passes it to the stage via a new `initial_audio_latents` arg.
- **`LTX2DiffusionStage.generate`** uses the supplied audio x_T; falls back to the old `torch.randn` only when none is supplied (non-pipeline / test callers).

Scope is the audio **initial** noise source only. The per-step audio SDE draw still uses the global RNG via `StepStrategy.denoise(generator=None)` by deliberate framework design (diversity/perf), matching the video SDE path — left as-is.

## Test Plan

- `python -m py_compile` on the three changed files — passes.
- `ruff check` + `ruff format --check` (the lint CI gate) — passes.
- Numerical test of `NoiseRecipe.resolve(salt=...)` on torch 2.7.1, 5 assertions, all pass:
  - `resolve(salt="audio")` is reproducible (two calls byte-identical) and **stable across processes** (same sha256 in a fresh interpreter) → cross-engine determinism;
  - the salted audio stream is **not** a prefix-alias of the video stream (independence), while a **control** using the same recipe *without* salt **is** a prefix-alias — exactly the aliasing this fix prevents;
  - `salt` bypasses a Path-1 `initial_latents` tensor and regenerates; plain `resolve()` still returns the shipped Path-1 tensor.
- **Not run:** full GPU LTX-2 model smoke (load + rollout) — needs the checkpoint + GPU. The change is confined to noise-recipe plumbing and leaves the video path byte-identical.

## Compatibility / Risk

- `resolve()`'s new args are keyword-only with defaults; the **video x_T path and all existing callers are byte-for-byte unchanged** (default `salt=""`).
- `LTX2DiffusionStage.generate` gains an optional `initial_audio_latents`; omitted → previous `torch.randn` fallback, so non-pipeline callers are unaffected.
- Behavior change is limited to the LTX-2 audio stream's initial noise: deterministic/grouping-aware and no longer perturbs the video RNG stream. No checkpoint/data migration.


KemingWu (approved):
LGTM

PR #70, opened 2026-06-15
refactor(sampling): modality-keyed sampling dict in RolloutReq, drop ComposedSamplingParams

celve:
## Summary

`RolloutReq.sampling_params` becomes a modality-keyed `Dict[str, BaseSamplingParams]`
(keys `"diffusion"` / `"ar"`), matching the other per-modality fields (`primitives`,
`request_conditions`, `RolloutResp.tracks`). This removes the bespoke
`ComposedSamplingParams` dataclass and the scattered `isinstance` / hand-rolled `N*M`
special-casing in favor of plain dict access.

- A `build_sampling_dict()` normalizer at the trainer boundary wraps flat
  single-modality recipes under their modality key and instantiates composed recipes
  per entry — so the flat recipes need no change; only the **6** composed
  (PE / unified-model) recipes drop their `_target_: ComposedSamplingParams` line.
- `total_samples_per_prompt()` replaces the per-prompt `N*M` math and drives the
  reward expansion-factor guard.
- The `get_diffusion_params` / `get_ar_params` accessors are removed; call sites read
  `sampling_params.get("diffusion")` / `.get("ar")` directly.
- BAGEL / HunyuanImage3 `t2ti` mode dispatch keys on dict membership
  (`"ar"`+`"diffusion"` ⇒ `t2ti`, `"ar"` ⇒ text-out) instead of `isinstance`.

## Related Issue

N/A

## Test Plan

No automated suite covers this area; validated functionally with the repo Python env:
- `python -m compileall unirl` passes; all touched modules import.
- `build_sampling_dict` resolves the 6 composed recipes to `{"ar", "diffusion"}` and
  the 56 flat recipes to a single-key dict.
- `total_samples_per_prompt` returns `N*M` for composed, the value for flat, `1` for
  empty; the BAGEL/HI3 dict-dispatch reproduces the old `isinstance` routing.
- `grep -rn "ComposedSamplingParams\|get_diffusion_params\|get_ar_params" unirl/ examples/`
  returns nothing.

Not run; reason: no GPUs available here, so no training/rollout smoke test.

## Compatibility / Risk

Config: the 6 composed sampling recipes are reshaped (drop the
`_target_: ...ComposedSamplingParams` line); flat recipes are unchanged.
`RolloutReq` is an in-memory type (not a serialized checkpoint format), so no
checkpoint impact. Any external caller constructing a `RolloutReq` with a bare
sampling-params object must now pass a `{"diffusion": ...}` / `{"ar": ...}` dict.

[...]

PR #130, opened 2026-06-26
fix(bagel): replace removed get_diffusion_params with modality-keyed dict access

NancyFyong:
## Summary

`origin/main` currently fails to import `unirl.rollout.engine.vllm_omni.adapters`, which breaks the entire vllm-omni adapter registry for every user:

```
ImportError: cannot import name 'get_diffusion_params' from 'unirl.types.sampling'
```

PR #89 (BAGEL integration, `ae2693a`) imported `get_diffusion_params` from `unirl.types.sampling`, but that helper was already removed by PR #70 (`30654a8` — *"refactor(sampling): modality-keyed sampling dict in RolloutReq, drop ComposedSamplingParams"*).

The same issue was already noted in `unirl/models/ltx2/pipeline.py:213-215`:
> `get_diffusion_params` helper only exists on the in-flight ComposedSamplingParams branch, not on main, so it broke import here.

…but PR #89 missed it.

## Fix

Replace `get_diffusion_params(req.sampling_params)` with `req.sampling_params.get("diffusion")` — the same accessor every sibling adapter uses (`qwen_image.py`, `sd3.py`, …). 3 lines changed.

## Test Plan

Verified import on clean `origin/main` + this patch:

```python
import unirl.rollout.engine.vllm_omni.adapters as a
adapters = a.registered_adapters()
print("count:", len(adapters))         # 9
print("bagel_t2i" in adapters)         # True
```

Imports cleanly; all 9 modalities registered including `bagel_t2i`.

Not run; reason: I don't have BAGEL weights handy to confirm the BAGEL recipe still runs end-to-end. Maintainers please confirm.

### DIFF
--- a/unirl/rollout/engine/vllm_omni/adapters/bagel.py
+++ b/unirl/rollout/engine/vllm_omni/adapters/bagel.py
@@ -58,7 +58,6 @@ from unirl.rollout.engine.vllm_omni.utils.sigmas import sigmas_list_from_req
 from unirl.sde.runtime import FlowMatchSchedulePolicy
 from unirl.types.rollout_req import RolloutReq
 from unirl.types.rollout_resp import RolloutResp
-from unirl.types.sampling import get_diffusion_params
 
 
 class BagelInputAdapter(DitInputAdapter):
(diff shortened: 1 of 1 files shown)


celve:
cc @zzhuoxin1508

celve (approved):

PR #136, opened 2026-06-27
feat(rollout): WAN 2.1 T2V sglang_diffusion rollout (full+LoRA, curve-aligned)

leviking98z-rgb:
## Summary

Adds **WAN 2.1 T2V** to the `sglang_diffusion` rollout engine (full-FT + LoRA), curve-aligned with the trainside sampler.

- `Wan21T2VAdapter` — rides the `VideoAdapter` base (single UMT5 text encoder; no CFG negative branch when `guidance_scale <= 1`); the sglang server resolves the WAN pipeline from `model_path`.
- WAN `param_names_mapping` + **sglang-leaf-name LoRA targets** (`to_q/to_k/to_v/to_out`) so all 240 attention projections wrap (vs 90 with diffusers paths) — otherwise sglang ran ~base weights and the curve went flat.
- Configs: `wan21_t2v_sglang{,_loramerge,_full,_full_nccl}.yaml`.

## Related Issue

Engine support matrix — #127. Depends on the foundation #135.

## Test Plan

- `py_compile` + import smoke (registers as `wan21`; video module imports) — passes.
- GRPO rollout<->replay reward-curve alignment vs trainside (`old_logp_source=replay`, cross-engine safe): tracks the trainside curve with minimal deviation.

## Compatibility / Risk

Additive — new adapter + configs; depends only on the foundation. No change to existing engines.

## Reviewer Notes

Independent of the other model PRs (#137–#139). Diff vs `main` includes the foundation commit (#135) until that merges.

## Checklist

- [x] I reviewed the changed code and removed unrelated/generated artifacts.
- [x] I updated tests, docs, and configs where needed, or explained why not.

## EXP
WAN21 TRAINSIDE LORA
截屏2026-07-01 11 54 22

WAN21 SGLANG LORA
截屏2026-07-01 11 54 44

WAN21 SGLANG FULL
截屏2026-07-01 11 56 16

WAN21 TRAINSIDE FULL
截屏2026-07-01 11 55 30

PR #135, opened 2026-06-27
feat(sglang-rollout): shared video-rollout infra (VideoAdapter base + WAN scheduler + weight-sync)

leviking98z-rgb:
## Summary

Adds the **shared `sglang_diffusion` video-rollout foundation** that the four video-model PRs build on (#136 WAN 2.1, #137 HunyuanVideo, #138 LTX-2, #139 WAN 2.2-A14B). Those four are **independent of each other** and depend only on this PR.

This extracts the model-agnostic plumbing so adding a video family becomes an adapter + config swap:
- **`VideoAdapter` base** — 6-D video trajectory (`[B, T+1, C, F, H, W]`) -> ragged `Videos`; the response-shape base every video family subclasses.
- **`patch_wan_scheduler`** — swaps WAN's hardcoded `FlowUniPCMultistepScheduler` (ODE, no RL mixin) for the RL-mixin `FlowMatchEulerDiscreteScheduler` (per-step SDE log-prob + external sigma); shared by both WAN models.
- **Weight-sync** — `_apply_fused_param_mapping` now applies a model's `param_names_mapping` string-renames (`attn1.to_q -> to_q`) and empty-string drops, so rename-based models sync real weights instead of stale base; `SafeUnpickler` allowlist for unirl tensor-rebuild during full-weight sync; `UNIRL_SYNC_CPU_STAGE` CPU-staged colocate sync for kernels without `pidfd_getfd`.

## Related Issue

Engine support matrix — #127.

## Test Plan

- `py_compile` + import smoke of the `sglang_diffusion` package on this branch (adapters registry, hijack patch suite, weight-sync) — passes.
- Exercised end-to-end by the four model PRs that depend on it (see their Test Plans).

## Compatibility / Risk

Additive — touches only the `sglang_diffusion` rollout path; existing engines/trainers unchanged. The weight-sync rename/drop is gated on a model defining `param_names_mapping` (no-op otherwise). No new dependencies.

## Reviewer Notes

Base of a foundation + 4-independent-models stack. Merge this first, then #136–#139 in any order.

## Checklist

- [x] I reviewed the changed code and removed unrelated/generated artifacts.
- [x] I updated tests, docs, and configs where needed, or explained why not.

Issue #127, opened 2026-06-25
Video RL Roadmap


leviking98z-rgb:
A video-focused slice of the master roadmap (#25). Contributions and feedback are welcome.

Current support (models / tasks / engines / rewards) is tracked in the matrices in the comments below. This roadmap lists what to build next.

## Tasks
- I2V for the remaining models (LTX-2, HunyuanVideo 1.5)
- V2V / editing, first-last-frame, video extension
- Control (pose / depth / reference), TI2V
- Audio / S2V, animate (later; need separate checkpoints)

## Rollout
- Engine coverage: **sglang** across video models, **FastVideo** (new), **vllm_omni** beyond HunyuanVideo 1.5

## Reward
- Learned video reward (temporal consistency, motion quality, text-video alignment)
- Differentiable reward (REFL / DRaFT-K)
- Multi-reward composition

## Performance
- Sequence / context parallelism across video models and long clips
- VAE decode tiling and offload
- Disaggregated / asynchronous rollout and training
- FP8 rollout (later)

leviking98z-rgb:
### Task support matrix

| Task | LTX-2 | WAN 2.1 | WAN 2.2 | HunyuanVideo | HunyuanVideo 1.5 |
|---|:---:|:---:|:---:|:---:|:---:|
| T2V | ✅ | ✅ | ✅ | ✅ | ✅ |
| I2V | 🔲 | ✅ | ✅ | 🔲 | 🔲 |
| TI2V | 🔲 | – | 🔲 | – | 🔲 |
| First / last frame | 🔲 | 🔲 | 🔲 | – | – |
| V2V (editing) | 🔲 | 🔲 | 🟡 #113 | – | – |
| Video extension | 🔲 | 🔲 | 🔲 | 🔲 | 🔲 |
| Control (pose / depth / ref) | 🔲 | 🔲 | 🔲 | – | – |
| Audio / S2V | 🟡 #115 | – | 🔲 | 🔲 | 🔲 |
| Animate | – | – | 🔲 | – | – |

**Legend:** ✅ supported (recipe on main) · 🟡 in progress (PR) · 🔲 not yet supported (planned) · – not native / needs a separate checkpoint

**Currently supported:** T2V (all five), I2V (WAN 2.1, WAN 2.2).

leviking98z-rgb:
### Engine support matrix

[...]

PR #139, opened 2026-06-27
feat(rollout): WAN 2.2-A14B T2V sglang_diffusion rollout (dual-expert MoE, full+LoRA)

leviking98z-rgb:
## Summary

Adds **WAN 2.2-A14B T2V** (dual-expert MoE) to the `sglang_diffusion` rollout engine (full-FT + LoRA); rollout aligned at **ratio = 1.0000**.

WAN 2.2-A14B runs two experts (high-noise / low-noise) switched at a sigma boundary (`boundary_ratio = 0.875`). sglang's multimodal_gen **already serves this** (auto-loads `transformer_2` when the checkpoint `model_index.json` carries `boundary_ratio` + both transformers; `DenoisingStage._select_and_manage_model` routes per-step), so the unirl side is thin:
- `Wan22T2VAdapter` — a thin `VideoAdapter` subclass forwarding `guidance_scale_2` for the low-noise CFG branch; the dual-expert lives entirely engine-side.
- **`patch_wan_scheduler` eta-drop guard** — the dual-expert denoise wraps `scheduler.step` with `**kwargs`, defeating sglang's kwarg filtering so the diffusers-style `eta` leaked into the RL-mixin `step` (which has no `eta`); drop the stray `eta` (matches the single-transformer path).
- Config: `wan22_t2v_14b_sglang.yaml` (layerwise expert offload for the 27B colocate).

## Related Issue

Engine support matrix — #127. Depends on the foundation #135.

## Test Plan

- `py_compile` + import smoke (registers as `wan22`; `Wan22T2VAdapter` subclasses `VideoAdapter`) — passes.
- GRPO smoke (256x448x5, layerwise offload): both experts load + LoRA, 10/10 dual-expert denoise, VAE decode, reward, replay -> **rollout 1 ratio = 1.0000 +/- 0.0000**, reward 0.78.

## Compatibility / Risk

Additive. The scheduler `eta`-drop is a no-op everywhere `eta` was already filtered. **Known limitation:** multi-rollout weight-sync OOMs on a single 8-GPU node (the 27B dual-expert + sync staging after the first optimizer step) — needs sync-time HBM headroom (smaller LoRA rank / more offload / multi-node); rollout alignment itself is proven.

## Reviewer Notes

Independent of #136/#137/#138. `Wan22T2VAdapter` subclasses `VideoAdapter` directly (NOT Wan21), so this PR does not depend on the WAN 2.1 PR.

## Checklist

## EXP
WAN22 LORA
截屏2026-07-03 07 23 08

WAN22 FULL
截屏2026-07-03 07 25 14

### DIFF
--- a/examples/diffusion/ltx2/ltx2_3_t2av_audioreward_trainside.yaml
+++ b/examples/diffusion/ltx2/ltx2_3_t2av_audioreward_trainside.yaml
@@ -1,0 +1,180 @@
+# @package _global_
+# LTX-Video-2.3 T2AV GRPO with AUDIO REWARD — trainside (in-process FSDP).
+#
+# Extends the T2AV recipe (ltx2_3_t2av_trainside) by scoring the jointly-
+# generated AUDIO in addition to the video. The pipeline decodes the audio
+# waveform (audio_vae + vocoder) and carries it on the track as a parallel
+# stream; the reward service injects it into the reward request so a composite
+# scorer can blend video + audio signals.
+#
+# Reward = T2AVCompositeScorer:
+#   - videopickscore (video quality / image-text alignment on first frame)
+#   - clap           (audio-text alignment, LAION CLAP; zero extra deps)
+# ImageBind (audio-video alignment) is also available as an inner scorer but is
+# NOT enabled here: it requires the facebookresearch/ImageBind package and is
+# CC-BY-NC-SA 4.0 (NonCommercial). To enable, add `imagebind: <weight>` to the
+# composite weights (and install the package).
+#
+# With audio_joint_sde=true, audio is part of the SDE policy AND now receives a
+# direct reward signal, so it is optimized toward prompt alignment (not just
+# passively coupled through cross-attention).
+#
+# Launch (1 node × 8 GPUs):
+#   bash examples/run_experiment_multinode_taiji.sh diffusion/ltx2/ltx2_3_t2av_audioreward_trainside
+
+num_devices: 8
+batch_size: 8
+num_rollouts: 10000
+
+logging:
+  report_to_wandb: true
+  project_name: ${oc.env:WANDB_PROJECT,unirl-ltx2.3-t2av-audioreward}
+  run_name: null
+  entity: ${oc.env:WANDB_ENTITY,null}
+  tags: [ltx2.3, t2av, grpo, trainside, dancegrpo, audio, audioreward]
+  log_media: true
+  media_max_items: 4
+  # How often (in ROLLOUTS) to log generated media to wandb. One rollout =
+  # one sampling batch (batch_size prompts x samples_per_prompt), NOT one
+  # optimizer step. Set high to keep media logging sparse / cheap.
+  media_log_interval: 500
+
+bundle:
+  _target_: unirl.models.ltx2.bundle.LTX2Bundle.from_config
+  config:
+    _target_: unirl.models.ltx2.config.LTX2PipelineConfig
+    pretrained_model_ckpt_path: ${oc.env:PRETRAINED_MODEL,dg845/LTX-2.3-Diffusers}
+    model_precision: bf16
+    autocast_precision: bf16
+    trajectory_precision: fp16
+    logprob_precision: fp32
+    shift: 1.0
+    max_sequence_length: 512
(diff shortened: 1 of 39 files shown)


leviking98z-rgb:
## Review + rebase/fixups (self-review before handoff)

Rebased onto latest `main`, stripped cluster-specific env workarounds, and fixed a hidden cross-PR dependency. Summary + verification below.

### Rebased onto `main` (was CONFLICTING → now clean)

`main` moved far ahead of this PR's base and the shared foundation (VideoAdapter, `stack_decoded_videos`, `patch_wan_scheduler`, weight-sync rename/drop) landed independently, so the branch was stale/conflicting. Resolved all 6 foundation conflicts by taking `main`'s version, keeping only the WAN 2.2 delta. Diff collapsed from **143 files** to **5 files / +290 −3**: the wan22 sglang YAML, `Wan22T2VAdapter` (+registration), the `patch_wan_scheduler` eta-drop guard, and the config fix below.

Also fixed a stale `Ltx2T2VAdapter` ghost entry in `__all__`, and added `Any, Dict` to `video.py`'s typing import (main only imported `List, Optional`, but `Wan22T2VAdapter.build_sampling` annotates `Any`/`Dict` → would `NameError` otherwise).

### Fix: removed the hidden dependency on #136 (commit `e17a59f`)

The PR claims "Independent of #136/#137/#138", but it was **not**: `WAN22PipelineConfig(WAN21PipelineConfig)` inherits from WAN 2.1's config, and the `use_lora` / `lora_target_modules` fields the wan22 sglang YAML sets (and `engine.py` reads) are added by **#136**, not present on `main`'s `WAN21PipelineConfig`. Merging #139 before #136 would `AttributeError` / reject the kwarg.

Fixed by declaring `use_lora` / `lora_target_modules` **directly on `WAN22PipelineConfig`**. Verified this is merge-order-independent: if #136 later adds the same fields to the parent, the child's redeclaration is a harmless same-default override (confirmed both dataclass inheritance orders construct without field-ordering errors).

### Stripped cluster-specific env workarounds (commit `2efa2b4`)

Removed `_maybe_disable_cudnn` (`unirl/__init__.py`) and the `UNIRL_SYNC_CPU_STAGE` CPU-staging path (`TensorWeightSync` + `sgl_compat` guard) — cluster-specific environment workarounds (cuDNN forward-compat crash / kernel-5.4 missing `pidfd_getfd`) that belong in launch scripts, not the upstream PR. All three files restored to `main`.

### Verified

- `Wan22T2VAdapter.build_sampling` override signature is byte-identical to `ImageAdapter.build_sampling`; `guidance_scale_2` is forwarded via `getattr(..., None)` and omitted when unset, so a `guidance_scale=1.0` smoke is unaffected.
- Dual-expert mechanism is entirely engine-side (sglang auto-loads `transformer_2` from `model_index.json` `boundary_ratio`; `DenoisingStage._select_and_manage_model` routes per-step) — the unirl adapter is genuinely thin. Comments are accurate.
- All YAML `_target_`s resolve (`WAN22PipelineConfig`/`Bundle`/`Pipeline`, `VideoPickScoreScorer/Spec`, `FlowGRPO`, `LoraConfig`); `boundary_ratio` and `guidance_scale_2` exist on both `WAN22PipelineConfig` and `DiffusionSamplingParams`.

### Note on the eta-drop guard (not blocking)

`patch_wan_scheduler` is applied **unconditionally in every sglang_diffusion engine process** (not WAN-gated), and it patches `FlowMatchEulerDiscreteScheduler.step` — the same scheduler class SD3/Flux/Qwen/Z-Image roll out on. The guard only `kwargs.pop("eta", None)`, and those families already filter `eta` out before `step`, so it is a genuine no-op for them (idempotent via the `_unirl_drop_eta` sentinel). The claim holds; just flagging that the blast radius is "all families using this scheduler", not "WAN only" — worth a one-line mention in the (already detailed) comment.

The known limitation in the PR description (single-node 8-GPU multi-rollout weight-sync OOM on the 27B dual-expert) is real, not a bug — it needs sync-time HBM headroom.

leviking98z-rgb:
Superseded: re-opening from the same (history-cleaned) branch to reset the PR timeline. Content is byte-identical.

PR #140, opened 2026-06-28
feat(hi3): veomni expert parallelism  for HunyuanImage-3

zzhuoxin1508:
## Summary

Add **Expert Parallelism (EP)** to the VeOmni FSDP2 training backend, enabling
LoRA RL training of the 80B, 64-expert MoE **HunyuanImage-3.0 (HI3)**.

The frozen 64-expert MoE is the memory bottleneck. EP shards the experts across
`ep_size` ranks (each rank owns `E/ep_size` experts) on a **separate `(ep, ep_fsdp)`
device mesh, orthogonal** to the existing `dp_shard × ulysses` FSDP mesh; tokens are
routed via `all_to_all` and computed with a fused grouped-GEMM kernel. Experts stay
frozen — only attention LoRA is trained.

Driven by a **single switch**: `backend.fsdp_cfg.ep_size > 1`. At `ep_size = 1` the
path is a true no-op and byte-identical to the prior VeOmni backend (every other
VeOmni-backed model is unaffected).

What's included:
- Meta-time swap of HI3's per-expert `ModuleList` → a fused `FusedHunyuanMoE`
  (`[E, 2I, H]` tensors, `Shard(0)`), wired via a model-agnostic bundle hook.
- EP-sharded weight load that bypasses DCP (per-block broadcast to keep load-time
  peak memory at `[E/ep, …]`, not the full `[E, …]`).
- Root-param all-gather hooks (`wte`/`ln_f`/`lm_head`) for calls outside the FSDP
  forward.
- Fixes surfaced while bringing the path up: activation-checkpointing was a silent
  no-op under FSDP2 (class-name match), and the fused-`qkv_proj` LoRA adapter was
  mis-sliced when synced to the AR rollout engine (GQA de-interleave).

**Result:** end-to-end single rollout drops ~25.5 min → ~18 min (**≈ -29%**); the
training step alone is >2× faster (see Test Plan).

## Related Issue

N/A

## Test Plan

**End-to-end rollout timing** — HI3-Instruct, 8×GPU, `ep_size=8`, EP vs FSDP backend:

| Phase | FSDP backend | EP backend | Delta |
| --- | --- | --- | --- |
| ① AR generation | `16:54:57→17:04:08` = 9 m 11 s | ~9 m 30 s | Nearly identical (vLLM engine, backend-agnostic) |
| ② DiT image gen | `17:04:08→17:06:07` = 2 m 00 s | ~2 m | Same |
| ③ reward + train step | `17:06:07→17:20:30` = ~14 m 23 s | ~6 m 40 s | **FSDP ~2.2× slower** ⚠️ |
| **e2e per rollout** | `16:54:57→17:20:30` = ~25 m 33 s | ~18 m 10 s | **EP ~7.5 min faster** |

- `pre-commit run --files <changed files>` — **green** (ruff check/format, body checks).
- LoRA de-interleave correctness verified against HI3's HF `modeling_hunyuan_image_3.py`
  qkv layout and the rollout-side `_split_qkv_weight`, plus a CPU reproduction of the
  channel mapping.
- Not run: dedicated CPU unit tests for the new EP helpers (`fuse_expert_state_dict`,
  `_deinterleave_gqa`) — see Reviewer Notes. Full multi-step convergence run: not in
  scope of this PR (single-rollout smoke + timing only).

## Compatibility / Risk

- **No-op at `ep_size=1`**: other VeOmni-backed models (qwen_image, sd3, qwen3-SP) are
  unchanged. `ep_size>1` requires a `veomni` build that supports `extra_parallel_*`.
- **Activation checkpointing now actually applies** for any VeOmni recipe with
  `activation_checkpointing: true` (it was a silent no-op before). Three existing
  recipes set it (qwen_image, qwen3 DRPO/GRPO-SP); expect a throughput/memory change,
  and the AC + Ulysses-SP combination is newly exercised — **re-validate those runs**.
- **LoRA de-interleave** assumes HI3's GQA-interleaved fused `qkv_proj`; it is
  size-guarded and falls back to a plain block split otherwise.
- **Load-time host memory**: rank 0 still materializes the full filtered checkpoint
  (~80B) in host RAM before broadcast; on a shared cgroup memcg this can OOM — keep the
  checkpoint off `tmpfs` and size the limit accordingly.

## Reviewer Notes

N/A

## Checklist


unirl/train/backend/veomni/wrap.py
CjhHa1, line 120: if activation_checkpointing=True and len(block_instances)==0, need fast failure

unirl/models/hunyuan_image3/bundle.py
```diff
@@ -449,6 +484,26 @@ def materialize(
...
+            # DTensors; HI3's ForCausalMM wrapper calls them OUTSIDE the decoder
+            # forward (wte -> inputs_embeds for ViT scatter; ln_f/lm_head -> logits),
+            # hitting `aten.* got mixed Tensor and DTensor`. Hook those three to
+            # all-gather their weights for such direct calls. Pass the whole
+            # transformer so lm_head (on the outer ForCausalMM) is reachable too.
+            from unirl.train.backend.veomni.ep import register_unsharded_param_hooks
+
+            n_hooked = register_unsharded_param_hooks(self.transformer)
+            if _current_rank() == 0:
+                print(f"expert-parallel: hooked root params for direct all-gather: {n_hooked}", flush=True)
```
CjhHa1: same as above

unirl/models/hunyuan_image3/bundle.py
```diff
@@ -449,6 +484,26 @@ def materialize(
             ),
         )
 
+        if getattr(self, "_ep_enabled", False):
+            from unirl.train.backend.veomni.ep import load_ep_experts
+            from unirl.train.backend.veomni.ep.models.hi3 import is_fused_expert_key
+
+            n_exp = load_ep_experts(self.transformer, expert_sd, is_fused_expert_key)
+            if _current_rank() == 0:
+                print(f"expert-parallel: loaded {n_exp} EP-sharded expert param(s)", flush=True)
```
CjhHa1: better use logger?

CjhHa1:
Also we need a smoke test for EP+SP, You can reply me here

zzhuoxin1508:
Phase | EP=16 (VeOmni) | non-EP (FSDPBackend) | Delta
-- | -- | -- | --
Rollout (AR+DiT, sequential dp) | ~23 min | ~24 min | ~same
Reward + Train | ~18 min | ~28 min | non-EP ~10 min slower
Step total | ~41 min | ~52 min | EP saves ~11 min/step (~21%)
Reward (step 1) | 0.88 | 0.8774 | ~identical ✅

<br class="Apple-interchange-newline">

zzhuoxin1508:
> Also we need a smoke test for EP+SP, You can reply me here

HY3 doesn't support SP (Ulysses) yet — the existing SP patchers only cover HF causal-LMs and diffusers transformers, and HY3 is a custom unified MoT with its own attention, so neither applies.maybe should open a separate PR to add SP support for HY3.

unirl/models/hunyuan_image3/bundle.py
zzhuoxin1508: fixed

leviking98z-rgb:
## Review + CPU/dist test results (ran what I could, static-checked the rest)

Follow-up on correctness verification. The pure functions that previously had only GPU-smoke backing (and no CPU unit tests, per your own Test Plan) — I ran real unit + distributed tests on an 8×GPU node (torch 2.7.1) and cross-checked the layouts against `modeling_hunyuan_image_3.py` in the HI3 checkpoint. Bottom line: **everything statically verifiable is green**; the only thing I couldn't reproduce locally is the veomni-kernel end-to-end (no veomni in the test env).

### ✅ Pure-tensor unit tests — 26/26 PASS (CPU, no veomni)

`_swap_gate_up_halves` / `fuse_expert_state_dict` / `_deinterleave_gqa`:

- **half-swap is both correct and non-vacuous**: reproduced HI3's real MLP math `x1*silu(x2)` vs veomni merged-fc1 `silu(g)*u`, and verified `HI3(orig) == veomni(swapped)` **and that WITHOUT the swap veomni diverges** — numerically confirming the swap is required and its direction is right.
- **fuse**: `[E,2I,H]`/`[E,H,I]` shapes, expert index ordering, gate_up half-swapped while down is left untouched, non-expert keys passthrough, non-contiguous indices raise — all pass.
- **de-interleave**: tagged each head-slot with a unique id and verified the GQA-interleaved → block `[q;k;v]` remap is element-wise identical to HI3's own qkv split, plus the size-guard fallback.

### ✅ EP-load sharding — dist test PASS (gloo, 4 ranks)

The per-block broadcast/select loop in `load_ep_experts`: every rank's transient is always `[E/ep,...]` (never the full `[E,...]` — the load-time memory-saving invariant), each rank gets its own contiguous expert block, and concatenating all ranks reconstructs the full tensor with no gaps/overlap/reorder.
> Note: the `distribute_tensor(my_block, param.device_mesh, param.placements)` step depends on veomni's EP DTensor topology, which can't be faithfully reproduced without veomni — that step is out of scope here (not tested, not faked).

### ✅ Static ground-truth check (against the HI3 checkpoint source)

- MLP `modeling_hunyuan_image_3.py` L1080-1081: `x1,x2 = chunk(2); down(x1*silu(x2))` → silu on the **2nd half**, confirming the half-swap direction.
- attn L1327-1329: `qkv.reshape(B,S,n_kv,groups+2,head).split([groups,1,1],dim=3)` → confirms qkv is GQA-interleaved, and `_deinterleave_gqa` mirrors it exactly.
- gate `topk_impl='easy'` is a real path; `easy_topk` returns `(weight, index)`, matching `FusedHunyuanMoE`'s unpack order.

### ❌ Couldn't reproduce locally

End-to-end EP training smoke (grouped-GEMM kernel / all_to_all / ep mesh) needs a veomni build with `extra_parallel_*` support, which isn't installed in the test env — that part still relies on your GPU runs under jobs/hi3_ep/.

### Suggestions (non-blocking)

1. Both tests above are pure CPU/gloo with zero veomni dependency — worth landing as regression tests (covers the CPU unit tests your Test Plan flagged as missing). Happy to paste the scripts or open a PR if useful.
2. In `load_ep_experts`, rank0's `full = expert_state_dict[name].to(device=param.device)` materializes the full `[E,...]` on the rank0 GPU (only `block` is `[E/ep,...]`). The PR body only mentions the host-RAM memcg OOM; consider documenting the extra `[E,...]` GPU headroom on rank0, or keeping `full` on CPU and moving each block to GPU per-broadcast.

**Overall**: the no-op guarantee, GQA de-interleave, half-swap, the AC fix, and the EP-block memory invariant are all verified sound and safe for existing paths. Leaning approve.

leviking98z-rgb (approved):

PR #145, opened 2026-06-29
perf(bagel): cache T2I prompt KV contexts across GRPO siblings

CjhHa1:
## Summary

BAGEL runs navit `bs=1`, so the rollout engine sends each of the `N`
`samples_per_prompt` GRPO siblings to `BagelPipeline.generate` one at a time. For
pure text-to-image every sibling shares the prompt and rebuilds the identical
`(gen, cfg_text, cfg_img)` KV contexts, re-running the text prefill `N` times per
prompt.

This PR memoizes `_build_contexts(prompt)` in a bounded per-pipeline LRU
(`cache_t2i_contexts`, default on; `context_cache_size`, default 32). The three
contexts route entirely through the und/shared prefill path, so for gen-only
LoRA (only the `*_moe_gen` experts train) a prompt's contexts are bit-stable for
the whole run and are already reused verbatim by `replay` — sharing one
reference across the `N` siblings is a pure speedup with zero behavior change.

To stay correct when that assumption does not hold (und LoRA targets, full / EMA
training), the cache is **self-guarding**: `_t2i_cache_enabled()` lazily inspects
trainable params on first use (after the backend injects LoRA / sets
`requires_grad`) and auto-disables if any non-gen prefill param is trainable —
otherwise stale contexts would silently pass the `ratio` check, since rollout and
replay reuse the same cached conditions. `it2i` is excluded (its input-image
prefill rides the trained gen experts).

Measured on `bagel_trainside_lora`, 1x8 H20: **-11.4% step time** (generate
-9.4%), with PPO `ratio` unchanged at `1.0000`.

## Related Issue

N/A

## Test Plan

- Static checks in this PR's rebased worktree:
  - `python3 -m py_compile unirl/models/bagel/pipeline.py unirl/models/bagel/config.py` -> OK
  - pre-commit on push (`ruff check`, `ruff format`, python-ast, merge-conflict, recipe `_target_` resolve, ...) -> all passed
- Empirical validation on the source branch (identical code):
  - `bagel_trainside_lora`, 1x8 H20: step time **-11.4%** (generate -9.4%); PPO `ratio` stays `1.0000` (on-policy invariant intact)
  - Cache correctness: cached prompt KV context is **bit-identical** to an independent fresh build (`max|diff| = 0`); a velocity forward on cached vs fresh context matches exactly
  - Self-guard: real gen-only LoRA -> 392 trainable params, all `*_moe_gen`, 0 non-gen -> cache stays ON; und / full-train -> cache auto-disabled (unit-tested)
- Not run in this isolated worktree: full training/rollout smoke test (no GPU env here); behavior is unchanged vs cache-off by construction + the bit-identity check above.

## Compatibility / Risk

- Adds two optional fields to `BagelPipelineConfig`: `cache_t2i_contexts: bool = True`, `context_cache_size: int = 32`. Backward compatible (existing configs/checkpoints unaffected).
- Default-on, but runtime self-guard auto-disables the cache whenever the prefill path is trainable, so und-LoRA / full / EMA recipes stay correct. Opt out with `cache_t2i_contexts=false`.
- Only affects the BAGEL T2I rollout path; `it2i` and all other models are untouched. No data-format or API breakage.

## Reviewer Notes

- This is the focused, upstream-ready slice of a larger local profiling branch: only the core context-cache + self-guard ship here (the profiling scripts/reports and an fbs>1 prototype are intentionally excluded).
- Rebased onto current `main`; resolved a purely textual overlap with #89's `prompts=` threading through `_diffuse_and_decode` (kept both changes). Independent of the open transformers-5.6 bump (#119) — no shared files.
- Duplicate-work check: scanned open PRs; #119 is a separate deps bump and no open PR covers this caching change.
- AI assistance: drafted with the Cursor agent; all changed lines were reviewed.

## Checklist


heguangxin (approved):

CjhHa1:
指标 | 缓存关闭 | 缓存开启 | 变化
-- | -- | -- | --
整步时间 | 182.47s | 161.58s | −11.4%
Generate | 125.96s | 114.14s | −9.4%
Train | 48.80s | 46.11s | −5.5%
Reward | 7.67s | 1.32s | 噪声较大，不应归因于缓存
峰值显存 | 58.46GB | 56.29GB | −2.18GB
PPO ratio | 1.0000 | 1.0000 | 无变化

Issue #13, opened 2026-06-09
Question about RL for Unified Models


Wang-Xiaodong1899:
HunyuanImage3 | Unified AR + diffusion | Text → Image | ✅
-- | -- | -- | --
Bagel | Unified AR + diffusion | Text → Image | ✅

Thanks for the great job!

Can this RL framework support these two models on Image+Text → Text, that is, the image understanding task?

haonan3:
Thanks for your interest!
 
Yes, the framework already supports the image understanding task (Image + Text → Text). We currently ship Qwen2.5-VL with full RL support — see the `unirl/models/qwen_vl/` pipeline and the example recipe `examples/ar/qwen_vl_grpo_geo3k_mc_4x8.yaml` (GRPO on Geometry3K, with SGLang and LoRA variants as well; we have tested the training and the reward curve increases well).

Wang-Xiaodong1899:
> Thanks for your interest!
> 
> Yes, the framework already supports the image understanding task (Image + Text → Text). We currently ship Qwen2.5-VL with full RL support — see the `unirl/models/qwen_vl/` pipeline and the example recipe `examples/ar/qwen_vl_grpo_geo3k_mc_4x8.yaml` (GRPO on Geometry3K, with SGLang and LoRA variants as well; we have tested the training and the reward curve increases well).

Thanks a lot. Qwen2.5-VL is ok. How about the RL for image understanding task on Bagel?

yanghlll:
> Thanks for your interest!
> 
> Yes, the framework already supports the image understanding task (Image + Text → Text). We currently ship Qwen2.5-VL with full RL support — see the `unirl/models/qwen_vl/` pipeline and the example recipe `examples/ar/qwen_vl_grpo_geo3k_mc_4x8.yaml` (GRPO on Geometry3K, with SGLang and LoRA variants as well; we have tested the training and the reward curve increases well).

Hi， great work. Can this RL framework support think mode, like UniGRPO.

haonan3:
@Wang-Xiaodong1899 Hi! — Bagel image understanding (**Image + Text → Text**) RL is now supported on main.

We just landed GRPO on the BAGEL understanding (AR) path (#100):
- Model: BAGEL-7B-MoT, run through BagelARStage with the understanding ViT tower enabled (`enable_vit: true`).
- Task / reward: multiple-choice VQA with exact-match reward (`MCExactMatchRewardScorer`), on ArxivQA.
[...]

PR #148, opened 2026-06-29
feat(bagel): reproduce UniGRPO on a single BAGEL trainside (unified-model), rebased onto main 

Ideny42:
## Summary

Reproduce UniGRPO (Prompt → Thinking → Image joint GRPO) on a single BAGEL-7B-MoT
backbone in the framework's unified-model mode: one shared MoT transformer drives
both halves on the same weights — the AR (und/text) experts generate the reasoning
"thinking" text, then the diffusion (gen/image) experts render an image conditioned
on prompt+thinking. Trained jointly — GRPO on text + a new `BagelFlowUniGRPO`
(FlowGRPO + velocity-MSE, optional GRPO-Guard RatioNorm) on the image — two backward
passes into one optimizer step. The HI3 two-engine path is untouched.

**This PR is the rebase of #95 onto current `main`.** Since #95 was opened, `main`
independently grew the BAGEL primitives the original branch had added
(`BagelARStage` / `BagelARConditions` / the `rl_ops` und path, a native t2ti
pipeline, `BAGEL_UND_LORA_TARGETS`) plus a modality-keyed sampling dict (#70), and
renamed `train/factories.py` → `train/optim.py` and `train/backend/fsdp.py` →
`train/backend/`. A literal commit replay was therefore an add/add cascade, so this
re-applies only the genuine UniGRPO delta on top of main's native primitives, then
layers the correctness/memory fixes found during GPU bring-up.

UniGRPO delta (re-applied on main):

- **dual-LR in one optimizer** — `OptimizerConfig.param_group_lrs` +
  `optim._build_named_param_groups`; the backend forwards named params so the und
  experts step at `learning_rate` and `moe_gen` (image) experts at the group LR.
- **multi-update PPO** — `UnifiedModelTrainStack.num_updates_per_batch`: split each
  rollout shard into disjoint mini-batches, freeze both π_old anchors once, one
  optimizer step per mini-batch (so `clip_range` / RatioNorm actually engage on the
  off-policy update).
- **single-engine (M=1) trainside mode** in `UnifiedModelTrainer`: one TRAINSIDE
  rollout on the shared pipeline (samples the live FSDP modules — no weight sync, no
  base offload); the image track shares the AR's prompt-level advantage (M=1);
  reward metadata is expanded 1:1 onto the image track. Selected by a single optional
  `rollout:` block in the entrypoint (vs HI3's `ar_rollout`/`dit_rollout`).
- **`BagelFlowUniGRPO`** — FlowGRPO clipped surrogate + unweighted velocity-MSE
  (`v_θ` vs frozen `v_ref`) replacing the latent-KL; optional GRPO-Guard RatioNorm;
  full-FT decoder-block unfreeze in the bundle; `build_forward_kwargs` /
  `predict_velocity_at` + `sde_means` recording on the diffusion stage; chunked VAE
  decode to CPU.
- **thin `BagelUniPipeline`** — P → P\*N → P\*N\*M (M=1) think→image fan-out reusing
  main's `BagelARStage` / `_build_think_contexts` / `diffuse` / `vae_decode`;
  returns the `{"ar","image"}` 2-track resp.
- **flash-attn-free SDPA varlen shim** (`sdpa_varlen.py` + a 2-line vendor import
  repoint) so the BAGEL generation/inference path runs without a `flash_attn`
  dependency (the train/replay path already used SDPA).
- **recipe** `examples/unified_model/bagel_trainside_unigrpo.yaml` — a self-contained
  example (PickScore local reward, no external service) of the full Prompt→Thinking→Image
  full-FT setup.

Correctness/memory fixes on top of the re-port (the "adjust to run on main" half):

- align RatioNorm μ_old (`sde_means`) with `sde_logp`: record `sde_means` only on
  SDE steps and pass `sigma_max=schedule[1]` to the recenter std, so the on-policy
  ratio stays ≈1 instead of over-correcting (~1.37) and clipping out the gradient.
- FSDP2-safe full-FT `v_ref`: swap the frozen base in by in-place local-shard copy
  (a `.data` pointer swap is silently ignored under `fully_shard`), and hoist all
  `v_ref` forwards under a single swap to bound peak memory.
- keep the diffusion SDE forward in `eval()` for the grad replay, so
  activation-checkpointing's recompute also takes `forward_inference` (the shared MoT
  is left in `train()` by the AR teacher-force).
- bind `train/ar/*` and `train/image/*` wandb metrics explicitly to `train/step`
  (the `train/*` glob does not match across the extra `/`).

## Related Issue

Related: #95 — the original "reproduce UniGRPO on a single BAGEL trainside" PR; this
PR is its rebase onto current `main` and is intended to land in its place. Background
discussion: #13.

## Test Plan

Static / config (rebased onto current `origin/main` — clean, no conflicts):

- `SKIP=no-commit-to-branch pre-commit run --files $(git diff --name-only origin/main..HEAD)`
  — all hooks green: check-yaml, ruff-check, ruff-format, detect-private-key,
  trailing-whitespace, end-of-files, python-AST / debug-statements. The
  `check-recipe-targets` hook confirms the committed recipe's `_target_` paths
  resolve. (Its only failures are the untracked, not-committed GenEval2 `*_4node*`
  recipes, which never reach CI.)
- `PYTHONPATH=$PWD python -m unirl.train_unified_model \
  --config-name unified_model/bagel_trainside_unigrpo --cfg job --resolve` — recommended
  config-resolution check; not run in the review env (no `ray`/`torch` there).

Live training (full config) — **completed** a 32-GPU / 4-node BAGEL UniGRPO trainside
full-FT + GenEval2 run (G=24, M=1, 25 steps, η=0.8, dual LR und 1e-6 / gen 3e-5,
RatioNorm, `num_updates_per_batch=2`). reward_mean trends up and the image + AR
ratio_mean stay ≈1 — curves below.

Clipboard_Screenshot_1782742754
Clipboard_Screenshot_1782742872
Clipboard_Screenshot_1782742905

The GenEval2 4-node recipe + the reward sidecar used for this run are kept local-only
(not committed); the checked-in `bagel_trainside_unigrpo.yaml` uses the in-repo
PickScore scorer so the example runs without a reward service.

## Compatibility / Risk

- **Additive** — the HI3 two-engine path is unchanged. Single-engine mode is gated on
  the new optional `rollout:` block; `train_unified_model` reads the rollout cfgs via
  `cfg.get(...)`, so both recipe shapes work from the one entrypoint.
- New `OptimizerConfig.param_group_lrs` and `UnifiedModelTrainStack.num_updates_per_batch`
  default to the previous behavior when unset.
- The `flash_attn` → `sdpa_varlen` repoint only affects BAGEL's vendored
  generation/inference attention; it removes a hard `flash_attn` import from that path.
- No checkpoint/data format changes. Full fine-tuning shards params+grads+optimizer
  state via FSDP; a 32-GPU/4-node class machine is the validated footprint.

## Reviewer Notes

- **Relationship to #95.** Same feature; this is the landable rebase onto current
  `main`, not a new capability. The per-component delta vs #95 (algorithm, trainer
  M=1 wiring, dual-LR, multi-update stack, pipeline fan-out) is logically equivalent
  — the file-set differs only because `main` reorganized the BAGEL module and renamed
  `factories.py`/`fsdp.py`.
- **Deliberate adaptation vs #95:** the per-prompt rollout runs navit **bs=1** (the
  branch's pack-B `forward_batch_size` packing is dropped — it conflicts with main's
  KV-in-conditions diffusion). With bs=1 the rollout and replay geometry match, so
  the image side uses `old_logp_source=rollout`. Re-adding pack-B for rollout
  throughput is a follow-up.
- **Net-new vs #95:** the flash-attn-free SDPA varlen shim and the diffusion
  `eval()` / FSDP2 `v_ref` fixes are not in #95 — they exist purely to make the
  feature run on current `main`. Worth a focused look.
- **Inherits #95's caveats:** no unit tests (behavior is validated by the live run);
  full-FT `v_ref` assumes a fresh run (a resume-from-checkpoint snapshot would
  capture resumed, not base, weights).
- AI-assisted. Duplicate-work check: distinct from #89 (vLLM-Omni BAGEL rollout,
  merged) and #62 (BAGEL multi-mode rollout); this is the trainside training core +
  recipe.

## Checklist


unirl/models/bagel/sdpa_varlen.py
```diff
@@ -0,0 +1,119 @@
+"""SDPA-based ``flash_attn_varlen_func`` drop-in for the vendored BAGEL modeling.
```
heguangxin, line 1: The sdpa_varlen.py drop-in looks correct (packed varlen, GQA expand, bottom-right causal for KV-cache decode) and is a reasonable way to drop the hard flash_attn dependency. 

One concern: it's effectively a patch hardcoded into the vendored modeling files — qwen2_navit.py / siglip_navit.py are edited away from pristine upstream just to repoint the import. That makes future re-vendoring noisier and unconditionally gives up the fused kernel even where a correct flash_attn is installed.

Suggestion: keep the stock files byte-identical to upstream and move the fallback into the vendoring layer (vendor/__init__.py), resolving the symbol only when it's actually missing:

### vendor/__init__.py — runs before any vendored submodule imports flash_attn
import importlib, sys, types
try:
    fa = importlib.import_module("flash_attn")
except Exception:
    fa = types.ModuleType("flash_attn"); sys.modules["flash_attn"] = fa
if not hasattr(fa, "flash_attn_varlen_func"):
    from unirl.models.bagel.sdpa_varlen import flash_attn_varlen_func
    fa.flash_attn_varlen_func = flash_attn_varlen_func

This keeps the model code unmodified, uses real flash when present (faster), and falls back only when the symbol is absent.

heguangxin (approved):

PR #149, opened 2026-06-29
feat(rollout): FastVideo VideoGenerator rollout engine (colocate)

Zcchill:
## Summary

Adds a FastVideo-backed WAN2.1 video rollout engine for colocated GRPO training. The engine adapts FastVideo's `VideoGenerator` to UniRL's typed rollout interface, synchronizes merged LoRA weights through checkpoint files, and supports both trainer-replayed and rollout-native old log-probabilities.

The rollout now follows the trainer's exact shift-applied sigma schedule and resolved SDE window. Replay and native modes therefore generate through the same Flow/Dance SDE transitions and deterministic low-sigma tail; native mode additionally returns FastVideo's transition log-probabilities. The implementation also preserves synced weights across sleep/wake and retries transient TCPStore port collisions.

FastVideo dependency: [`Zcchill/FastVideo`](https://github.com/Zcchill/FastVideo) at or after commit `7fe1d7db9a0b8aebb46679e7924f597431f23665`.

## Related Issue

N/A

### Dance YAML — baseline vs replay vs native

Clipboard_Screenshot_1784546103

### Flow YAML — baseline vs replay vs native

Clipboard_Screenshot_1784546093

## Test Plan

- Static checks on all files in this change:
  `SKIP=no-commit-to-branch python3 -m pre_commit run --show-diff-on-failure --files examples/diffusion/wan21/wan21_t2v_dancegrpo_fastvideo.yaml unirl/distributed/weight_sync/full/checkpoint.py unirl/rollout/engine/fastvideo/__init__.py unirl/rollout/engine/fastvideo/config.py unirl/rollout/engine/fastvideo/engine.py`
  - YAML, AST, whitespace, private-key, Ruff check, and Ruff format hooks passed.
- Recipe target validation:
  `python3 scripts/check_recipe_targets.py` — passed.
- Python compilation:
  `PYTHONPYCACHEPREFIX=/tmp/unirl-pycache python3 -m py_compile unirl/distributed/weight_sync/full/checkpoint.py unirl/rollout/engine/fastvideo/__init__.py unirl/rollout/engine/fastvideo/config.py unirl/rollout/engine/fastvideo/engine.py` — passed.
- End-to-end validation: WAN2.1-T2V-1.3B, single node with 8 GPUs, 720p × 5 frames, 14 denoising steps, VideoPickScore reward. Trainside baseline, FastVideo replay, and FastVideo native all completed rollout and optimization steps; native emitted/replayed log-probabilities aligned with importance ratio approximately 1 and zero alignment clipping.
- Hydra resolve was not rerun in the lightweight local checkout because `hydra` is not installed there; the same recipe was exercised by the end-to-end runs above.

- Post-fix GPU smoke: WAN2.1-T2V-1.3B, one node with 8×H20, FastVideo `7fe1d7db9a0b8aebb46679e7924f597431f23665`, Torch SDPA, 256×256 × 5 frames, 2 denoising steps, 1 prompt × 8 samples, 3 rollouts, replay log-probs, and checkpoint sync every rollout — passed.
  - All three rollout/optimization steps completed with finite non-zero grad norms (`0.0023`, `0.0029`, `0.0021`) and ratio `1.0000`.
  - The second wake rebuilt FastVideo, re-applied `weights_v0.pt`, then loaded `weights_v1.pt` successfully on all 8 workers (`825` keys, `0` missing/unexpected).
  - A real `EADDRINUSE` during wake was recovered by the port-retry path and the run continued.
  - Trainer teardown removed the run-scoped checkpoint namespace; no weight files remained.
- Follow-up regression validation after review fixes:
  - CPU-only checkpoint publication test covering stale-file replacement, per-run namespace isolation, and two-version cleanup — passed.
  - CPU-only SDE-window test covering `None`, explicit `[]`, deduplication, and out-of-range rejection — passed.
  - CPU-only wake failure/success test confirming failed weight restoration keeps the engine offloaded and discards the pretrained generator — passed.
  - `SKIP=no-commit-to-branch pre-commit run --all-files --show-diff-on-failure` — passed after merging the latest `main`.
  - `python -m py_compile` on the FastVideo engine/config and checkpoint sync modules — passed.

## Compatibility / Risk

- Requires the pinned FastVideo revision above and `FASTVIDEO_PATH` pointing to that checkout.
- The included recipe is a 1×8 colocated configuration and enables FSDP activation checkpointing for 720p memory use.
- On kernels without `pidfd_getfd` CUDA-IPC support, do not set `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True`.
- Native mode requires both `rollout.config.native_logprob=true` and `algorithm.old_logp_source=rollout`; the example defaults to replay.
- Current scope is WAN2.1 and local-mode colocate. FastVideo still regenerates initial noise from the derived per-sample seed rather than consuming UniRL's `x_T` tensor byte-for-byte.

## Reviewer Notes

The highest-risk areas are sigma pre-image conversion, SDE-column alignment, and sleep/wake weight preservation. The published FastVideo dependency contains the matching Flow/Dance transition implementation and deterministic tail behavior.

AI assistance was used to investigate alignment failures and prepare the implementation; the production diff was isolated from probe scripts and generated artifacts.

## Checklist


unirl/distributed/weight_sync/full/checkpoint.py
```diff
@@ -0,0 +1,112 @@
...
+        """
+        import torch
+
+        state_dict: Dict[str, torch.Tensor] = {}
+        for name, tensor in self._iter_full_tensors():
+            if self._my_rank == 0:
+                state_dict[name] = tensor.detach().to("cpu", copy=True)
+
+        version = int(self.weight_version)
+        path = os.path.join(self._dir, f"weights_v{version}")
```
leviking98z-rgb: **Blocking:** this publication name is only `weights_v{weight_version}`, while `weight_version` restarts at 0 for every run and the `.ready` files are never removed. A stale marker from a previous or concurrent run lets nonzero ranks leave `_wait_for_marker` immediately and load an old or concurrently overwritten checkpoint before rank 0 publishes the new one. Separately, every sync permanently leaves a full checkpoint plus marker behind; the example syncs a 1.3B model every rollout for 1000 rollouts, so `/tmp` grows by multiple GB per step and will fill. Please use a run-unique publication/manifest and bounded cleanup (while retaining the latest checkpoint needed by `wake_up`).

unirl/rollout/engine/fastvideo/engine.py
```diff
@@ -0,0 +1,590 @@
...
+        _g = _s / (_f - _s * (_f - 1.0))
+        sp.sigmas = [float(x) for x in _g.tolist()[:-1]]
+
+        # SDE window handed to FastVideo's denoiser so it injects exploration
+        # noise ONLY on the trainer's SDE steps and runs the rest as a
+        # deterministic Euler step (clean low-sigma tail). ``params.sde_indices``
+        # is stamped per rollout by the trainer (resolve_sde_indices); it matches
+        # the columns the trainer replays. Empty/None => every step is SDE.
+        _sde_idx = getattr(params, "sde_indices", None)
+        sde_step_indices = [int(x) for x in _sde_idx] if _sde_idx else None
```
leviking98z-rgb: **Blocking:** an empty `sde_indices` list is the framework's explicit forward-process/no-SDE value (`num_sde_steps=0`), but this truthiness check converts `[]` to `None`. FastVideo interprets `None` as 'every step is SDE', so a deterministic request becomes stochastic on all steps. `_build_segment` repeats the mismatch by replacing an empty set with `range(T)`. Preserve the distinction (`None` vs `[]`) end to end; for zero SDE steps native mode must also not require a log-prob tensor.

unirl/rollout/engine/fastvideo/engine.py
```diff
@@ -0,0 +1,590 @@
...
+                port_in_use = "EADDRINUSE" in str(exc) or "address already in use" in str(exc).lower()
+                if not port_in_use or attempt == max_port_attempts:
+                    raise
+                logger.warning(
+                    "fastvideo wake_up: master_port=%s busy (attempt %d/%d); retrying",
+                    self._ports.master_port,
+                    attempt,
+                    max_port_attempts,
+                )
+        self._is_offloaded = False
```
leviking98z-rgb: Set `_is_offloaded = False` only after the cached checkpoint has been successfully re-applied. If `update_transformer_weights_from_path` raises (missing/corrupt checkpoint, worker failure, etc.), the method currently exits with a freshly rebuilt **pretrained** generator while reporting `is_offloaded=False`; a retry then makes `wake_up()` return immediately and can serve stale pretrained weights. Keep the flag true on failure (and preferably tear down the partially restored generator), matching the fail-closed wake contract used by the other engines.

leviking98z-rgb (changes requested):
Found three correctness/lifecycle issues that should be addressed before merge; details are inline. I also verified that the branch auto-merges with current main and that the merged Python files compile. Non-blocking follow-up: current main now supports `WAN21PipelineConfig.load_vae=false` for dedicated-engine recipes, which would avoid loading an unused trainer-side VAE in this colocated FastVideo recipe.

leviking98z-rgb (approved):
The requested fixes are now applied: checkpoint publications are isolated per Ray job and bounded to the latest two versions; explicit empty SDE windows remain deterministic with no native log-prob requirement; wake-up fails closed if cached-weight restoration fails; and the dedicated FastVideo recipe skips the unused trainer-side VAE. Targeted CPU regressions, merged-tree Python compilation, full pre-commit, recipe target validation, lint, CodeQL, and the open-source scan all passed.

leviking98z-rgb (approved):
Re-reviewed latest head after the follow-up memory/cleanup changes. A 3-rollout 1×8 H20 FastVideo smoke completed end-to-end (WAN2.1-1.3B, 256×256×5, 2 denoising steps, 8 samples/rollout): finite non-zero grad norms, ratio 1.0000, successful checkpoint sync and wake-time weight reapply on all workers, real EADDRINUSE recovery, and teardown removal of the checkpoint namespace. Full pre-commit, lint, CodeQL, and open-source scan also pass.

PR #144, opened 2026-06-29
perf(pe): batched SD3 replay for PE diffusion training (ratio=1 exact)

CjhHa1:
## Summary
PE trainside (Qwen3 rewrite + SD3 diffusion + PickScore) was dominated by the **SD3 training phase** `diffusion_train` — **62% of the step (29.5s)**. Root cause is **not compute**: SD3.5-medium is ~compute-bound even at batch=1 (batching ceiling ~1.4×). It's **per-forward overhead** across **~192 batch=1 transformer forwards/step** (FSDP all-gather + Python dispatch), because replay walks the SDE steps one-by-one at `micro_batch_size=1`.

This PR adds **`batch_replay_steps`** to `SD3DiffusionStage`: it collapses the per-SDE-step replay loop into **one batched `[S*B]` transformer forward** (+ a single vectorized SDE transition), cutting S model forwards / FSDP all-gathers per replay to 1. Together with the existing `micro_batch_size` and `torch.compile` knobs it drops the per-step forward count from ~192 to as few as **8**.

- Gated, **default off** (code default `False`, stateless SDE strategies only — Flow/Dance/CPS; DPM2 excluded). Threaded through `SD3Pipeline` / `SD3PipelineConfig.from_config` and exposed on the PE recipe (ships `false`; enabled via override).
- **Preserves on-policy `ratio=1` exactly**: the π_old anchor (`prepare_segment`) and the train forward both run through the same `replay`, so under `old_logp_source=replay` they share an identical `[S*B]` geometry → bit-identical → ratio=1.

## Results
8×H20, PE trainside, 6 rollouts (stable r4–6), GPU occupier suppressed, **every config `ratio=1.0000±0.0000`**:

| config (override on the PE recipe) | step | Δstep | diffusion_train | Δ | generate | peak VRAM |
|---|--:|--:|--:|--:|--:|--:|
| baseline (`micro_batch_size=1`) | 48.3s | — | 29.5s | — | 15.3s | 61.6 GB |
| `reshard_after_forward=false` | 42.8s | −11% | 24.0s | −19% | 15.3s | 65.7 GB |
| `use_torch_compile=true` | 41.0s | −15% | 25.8s | −13% | 11.7s | 61.5 GB |
| **`batch_replay_steps=true`** (this PR) | 32.6s | −32% | 13.7s | −54% | 15.5s | **61.3 GB** |
| `micro_batch_size=8` | 30.1s | −38% | 11.3s | −62% | 15.4s | 78.3 GB |
[...]

PR #156, opened 2026-07-01
perf(diffusion): extend batched-step replay to qwen_image, z_image, flux2_klein

CjhHa1:
## Summary

Extends the SD3 batched-step replay optimization (#144) to the remaining **image** diffusion stages (`qwen_image`, `z_image`, `flux2_klein`), and de-duplicates the implementation into one shared mixin.

During diffusion-RL training replay, the `S` SDE steps of a stored trajectory were recomputed in a **serial loop** — `S` transformer forwards, i.e. `S` FSDP all-gathers of the full sharded transformer per `replay()` call. `_replay_batched_steps` instead stacks all `S` steps **step-major** on the batch dim (`sample`/`prev_sample` -> `[S*B, ...]`, conditioning tiled `S`x, sigmas ride as `[S*B]` vectors) and runs **one** forward + one vectorized SDE transition, collapsing `S` all-gathers -> 1.

Why this is safe: the transformer has no cross-sample interaction, so per-sample log-probs are unchanged up to floating-point batch-shape rounding (numerically equivalent, *not* claimed bit-identical). Under `old_logp_source='replay'` the π_old anchor and the train forward both take the *same* batched path, so the measured on-policy ratio stays exactly `1.0000` — verified in every configuration reported below.

Changes:

- New `BatchedStepReplayMixin` (`unirl/models/types/batched_replay.py`): the previously SD3-only `_replay_batched_steps` extracted verbatim, with two per-model hooks — `_tile_conditions(conditions, repeats)` and `_batched_step_kwargs(segment, params)` (default `{}`; only `qwen_image` overrides it, for `latent_h`/`latent_w`/`distilled_guidance_scale`).
- `sd3`, `qwen_image`, `z_image`, `flux2_klein` all inherit the mixin and implement their own `_tile_conditions`. SD3 drops its private copy; net effect is one implementation instead of four near-identical ones.
- Per-model tiling is the only real work: `qwen_image` replicates the per-call variable-length text trim + `img_shapes`/`txt_seq_lens`; `z_image` extends its list-based per-sample inputs to `S*B`; `flux2_klein` rebuilds 4-axis RoPE ids and preserves the replay `.eval()` fence.
- `batch_replay_steps` threaded through the Qwen-Image, Z-Image and Flux.2-klein pipeline configs and constructors (SD3 already had it from #144).

## Related Issue

N/A — direct follow-up to #144.

## Test Plan

Measured on H20 with real checkpoints, using out-of-tree profiling/validation harnesses (see "Reviewer Notes" — they are deliberately not part of this diff). `ratio = 1.0000` exact in every row.

- **Correctness, Qwen-Image on real 20B weights** (`Qwen-Image-Edit` transformer, B=2, S=3): batched==serial parity `max rel 1.03e-6`; `max|ratio-1| = 0.000`; grad flows. PASS.
- **Distributed FSDP2 replay A/B** (per-block `fully_shard`, `reshard_after_forward=True`, B=1) — serial (`S` all-gathers) vs batched (1):
  - `qwen_image` (20.4B): w2·S3 **3.08x**, w8·S3 **2.95x**, w8·S6 **8.26x**
  - `sd3` (2.24B): w2·S3 1.59x, w8·S3 2.74x, w8·S6 3.69x
  - `z_image` (6.15B, Z-Image-Turbo): w2·S3 1.18x, w8·S3 1.55x
- **End-to-end recipe A/B**, `qwen_image_trainside_veomni.yaml` on 2xH20, `+pipeline.batch_replay_steps=false/true`: total step ~94 s -> ~72 s (**~1.3x**), `ratio=1.0000+-0.0000` in both arms, reward comparable. Small-scale (8 samples/step, 3 rollouts) — the batched replay only accelerates the `diffusion_train` fraction of the step, so the isolated microbench above is the clean measure of the replay win.
- **Single-GPU control**: 1.05–1.14x only (and <1 at large `S*B`). This is expected and is the point — the win is the FSDP all-gather-count reduction, not single-GPU compute; a single-GPU bench understates it for every model.
- `flux2_klein`: implemented and correctness-reasoned, **GPU runtime validation not done** (single-file checkpoint + slow network mount at the time). Default off.
- Full `pre-commit` run passes; branch is merged up to date with `main` and conflict-free. Revalidated 2026-07-26 with an out-of-tree inline harness covering shared step-major ordering, output reshaping, every model's condition tiler (including Flux image latents/IDs), and the Qwen-specific replay kwargs. No test file was added.

Full measurement write-up, including per-row numbers and the reasoning about which models should and should not get this, is posted as a comment below.

## Compatibility / Risk

- Additive `batch_replay_steps` config field (default `False`) on Qwen-Image, Z-Image and Flux.2-klein. **No behavior change unless opted in.** No checkpoint or data-format changes.
- The fast path is gated on all three of: the flag, `S > 1`, and `isinstance(self.strategy, SDEStrategy)` (stateless Flow/Dance/CPS). Anything else takes the untouched serial loop. The stateful `DPM2Strategy` is an ODE strategy, not an `SDEStrategy`, so it can never enter this path.
- `ratio = 1` is preserved because the anchor and the train pass share the batched path, so numerics and the optimization trajectory are unchanged.
- Costs extra replay activation memory (the `[S*B]` stack). **Video and unified models are intentionally not touched** — their much larger per-step activations need a memory-aware (chunked-`S`) design, which is follow-up work.

## Reviewer Notes

- The commit messages on this branch reference `docs/diffusion_replay_batching_cross_model.md` and `scripts/profiling/{fsdp_replay_microbench,validate_batched_replay_qwen_image}.py`. Those were **intentionally removed** from the final diff as temporary profiling artifacts (@celve — this is why you could not find the doc). The doc's measured conclusions are reproduced verbatim-in-substance in a comment below so the numbers stay reviewable from the PR itself.
- On reward curves: the end-to-end A/B was a short smoke (3 rollouts, 8 samples/step, 2 GPUs), so its reward numbers overlap but are not a curve. See the comment below for the exact numbers and for why reward dynamics cannot change (identical log-probs up to fp rounding, `ratio = 1` in both arms, default off). Happy to run a longer 8-GPU reward-curve A/B if that is a merge requirement.
- `flux2_klein` is default-off and not GPU-validated; it cannot activate without both the strategy gate and the flag.
- Current diff is production code only: 12 files, no profiling script, no doc artifact, no test file.
- AI-assisted. Duplicate-work check: direct continuation of #144; no overlapping open PR (checked the open PR list).

## Checklist


celve:
LGTM. Do you have reward curves with batched replay turning on? I didn't find the doc mentioned in the pr: `docs/diffusion_replay_batching_cross_model.md`

CjhHa1:
@celve Thanks — answering both questions.

**On the missing doc:** `docs/diffusion_replay_batching_cross_model.md` was intentionally dropped from the final diff (commit `76d1d5d`, "exclude profiling artifacts from PR") together with the two profiling/validation scripts, so this branch would carry production code only. That is why you could not find it. Its measured content is reproduced below so the numbers are reviewable from the PR itself. One wording correction versus the original doc: it said "bit-identical" in places; the honest claim is **numerically equivalent** (parity `1e-7`/`1e-6` relative), while `ratio = 1.0000` is exact because the anchor and the train pass go through the same batched path.

**On reward curves:** no long reward curve was run. The end-to-end A/B was a small smoke run (3 rollouts, 8 samples/step, 2xH20) — reward ranges overlap but that is not a curve; details and the argument for why reward dynamics cannot change are in §5/§6 below. If a real reward-curve A/B is a merge requirement, say so and I will run a longer 8-GPU one; the flag is default-off, so it does not block anything else.

---

## 1. Central finding: this is a *distributed* (FSDP all-gather) optimization, not a single-GPU compute win

Measured on **1xH20** with the real SD3.5-medium transformer (2.24B, 24 layers), transformer-only, no FSDP:

| microbench (1 GPU, no FSDP) | result |
|---|---|
| forward throughput B=1 -> B=32 | 26.8 -> 34.3 samples/s = **1.28x** (per-sample) |
| fwd+bwd throughput B=1 -> B=16 | 6.9 -> 11.7 samples/s = **1.69x** |
| **serial S=3 forwards (bs=B) vs 1 batched (bs=S*B)** | **1.14x @B=1, 1.04x @B=4, 1.03x @B=8, 0.50x @B=16** |
| batched==serial parity | `max rel 1.1e-7` |
| batched-replay determinism -> ratio=1 | `max|ratio-1| = 0.000` (exact) |

On a single GPU the batched collapse is worth ~1.0–1.14x, and is **negative** once `S*B` is large (bs=48 is past SD3's compute-saturation point). Yet #144 measured −54% `diffusion_train` end-to-end on 8xH20. The entire delta is the **FSDP all-gather count reduction** (`S` -> 1 per replay) plus Python/launch overhead — not single-GPU compute. Consequence: a single-GPU microbench *understates or hides* the win for every model; the benefit has to be judged in the distributed FSDP setting.

Per `replay()` call, ignoring compute (invariant — same total FLOPs):

```
saved_allgather_time ~= (S - 1) * (param_bytes / effective_NVLink_BW)   [reshard_after_forward=true]
extra_activation_mem ~= (S - 1) * (per-step activation footprint at bs=B)
```

So a model is a good candidate when replay is **all-gather-bound**: large params relative to per-forward compute, `reshard_after_forward=true` (the default — re-gathers every forward), many SDE steps `S`, and small enough per-step activations that `[S*B]` fits.

## 2. SD3 single-GPU anchor (1xH20)

```
Claim 1 (batched==serial, bf16 tol + mapping): PASS   (max rel 1.1e-7)
Claim 2 (batched replay deterministic ratio=1): PASS   (max|ratio-1| = 0.000)
Claim 3 (grad flows through batched replay):    PASS
forward (no_grad):  serial=396.6ms  batched=378.5ms   speedup=1.05x
fwd+bwd:            serial=1202.6ms batched=1089.9ms  speedup=1.10x
```

## 3. Distributed FSDP2 replay sweep (the right lens)

Per-block `fully_shard` with `reshard_after_forward=True` (the training-path wrap), **B=1** (the `micro_batch_size=1` replay geometry). `ratio = 1` exact in every row.

**sd3 (2.24B)**

| world | S | serial (S all-gathers) | batched (1) | speedup |
|---:|---:|---:|---:|---:|
| 2 | 3 | 185 ms | 117 ms | **1.59x** |
| 8 | 3 | 1736 ms | 634 ms | **2.74x** |
| 8 | 6 | 3416 ms | 925 ms | **3.69x** |

**qwen_image (20.4B, real weights)**

| world | S | serial | batched | speedup |
|---:|---:|---:|---:|---:|
| 2 | 3 | 4365 ms | 1416 ms | **3.08x** |
| 8 | 3 | 8580 ms | 2909 ms | **2.95x** |
| 8 | 6 | 12183 ms | 1475 ms | **8.26x** |

**z_image (6.15B, Z-Image-Turbo)**

| world | S | serial | batched | speedup |
|---:|---:|---:|---:|---:|
| 2 | 3 | 1669 ms | 1419 ms | **1.18x** |
| 8 | 3 | 2589 ms | 1668 ms | **1.55x** |

**Cross-model summary** (B=1, `ratio = 1` exact everywhere)

| model (arch) | params | w2·S3 | w8·S3 | w8·S6 |
|---|---:|---:|---:|---:|
| sd3 (MMDiT) | 2.24B | 1.59x | 2.74x | 3.69x |
| z_image (list S3-DiT) | 6.15B | 1.18x | 1.55x | — |
| qwen_image (MMDiT) | 20.4B | 3.08x | 2.95x | **8.26x** |

The win grows with world size and `S`, and tracks the all-gather/compute balance: Qwen-Image's per-forward all-gather is ~10x SD3's (20B vs 2.24B), so Qwen at w2 (3.08x) already beats SD3 at w8 (2.74x), and at S=6 collapsing 6 all-gathers -> 1 hits 8.26x. z_image's win is *smaller* than SD3's despite ~3x the params — a real architecture effect: its list-based single-stream S3-DiT forward (plus refiner layers) is more compute-heavy per sample, so the all-gather share is smaller. That is the §1 model behaving as predicted, in both directions.

## 4. Qwen-Image correctness on real 20B weights (1xH20)

`QwenImageTransformer2DModel` (60 layers, 3072 hidden, ~20B), B=2, S=3, `guidance_scale=1.0`:

```
Claim 1 (batched==serial, bf16 tol + mapping): PASS   (max rel 1.03e-6)
Claim 2 (batched replay deterministic ratio=1): PASS   (max|ratio-1| = 0.000)
Claim 3 (grad flows through batched replay):    PASS
```

This is the part that matters for Qwen: it proves the tricky tiling (per-call variable-length text trim + `img_shapes` / `txt_seq_lens` replication) is correct on real weights, not just on random init.

## 5. End-to-end recipe A/B, including reward (this is the reward answer)

`qwen_image_trainside_veomni.yaml` on 2xH20 (`num_devices=2 +devices_per_node=2`), scoped small (batch_size=4, samples_per_prompt=2, 3 rollouts), reward pointed at local PickScore/CLIP, `+pipeline.batch_replay_steps` off vs on:

| run | step 1->2 | step 2->3 | ratio | reward |
|---|---:|---:|---|---|
| off (serial replay) | 97.8 s | 90.7 s | 1.0000+-0.0000 | 0.82–0.86 |
| on (batched replay) | 62.4 s | 82.0 s | 1.0000+-0.0000 | 0.78–0.84 |

Total step ~94 s -> ~72 s (**~1.3x**; per-step 1.1–1.5x, noisy at this scale because batched replay only accelerates the `diffusion_train` fraction while rollout-gen and reward are fixed cost).

**Why the reward ranges are not a curve, and why I do not expect one to differ:** 3 rollouts at 8 samples/step is inside run-to-run noise, so 0.82–0.86 vs 0.78–0.84 shows "nothing broke", not "same learning curve". The structural argument is stronger than the measurement here: the batched path computes the *same* per-sample log-probs as the serial path up to floating-point batch-shape rounding (§1/§4 parity `1e-7`/`1e-6`), the π_old anchor and the train forward both use it so `ratio = 1.0000` exactly (measured in both arms), and therefore the loss and its gradients are the same function of the same quantities. There is no mechanism by which reward dynamics change — the only real risk in this PR is a *wrong tiling*, which is what §4 tests. A longer 8-GPU reward-curve A/B is doable on request.

## 6. What was *not* measured, and the video/unified caveat

- `flux2_klein`: implemented and correctness-reasoned, **no GPU runtime validation** (single-file safetensors checkpoint + slow network mount at the time). Default off, and it cannot activate without both the SDE-strategy gate and the flag.
- Video models (`wan21/22`, `ltx2`, `hunyuan_video*`) and unified models (`bagel`, `hunyuan_image3`) are **deliberately not modified**. They pass the structural gate (identical serial loop, same `step_with_logp` primitive, stateless `SDEStrategy` in the RL recipes), but video forwards are already compute-bound at 10^4–10^5 latent tokens, so the relative all-gather share is smaller *and* stacking `[S*B]` multiplies an already-large activation footprint -> likely OOM at full `S`. The right design there is a **chunked** batched replay (group <=k steps) behind a memory check; unified models additionally have AR/diffusion and (Bagel) MoT + context-cache interplay. Both are follow-up work, not this PR.
- Practical note for anyone reproducing the 20B numbers: the runs were load-bound off the network checkpoint mount; copying the checkpoint to local disk and pointing the reward model at a local PickScore/CLIP with `HF_HUB_OFFLINE=1` is what made the end-to-end A/B in §5 finish at all.

CjhHa1:
Ran an independent validation of this branch on 4×8 H20, with the code at the PR head (`f18bf594`, clean worktree) rather than at the commits the original numbers were taken on. This covers the reward-curve question, the missing `flux2_klein` validation, and the memory cost the PR states but never measured. Harness was kept out of tree, same as yours.

Headline: **the optimization holds up — `ratio = 1` is exact in 24/24 FSDP configurations and the reward curves are indistinguishable across seeds.** Three things in the PR text should be corrected, and one precondition should be promoted from a detail to a documented requirement.

---

## 1. Reward-curve A/B (@celve's question)

`qwen_image_trainside` (native `FSDPBackend`), 8×H20 per arm, `batch_size=16`, `samples_per_prompt=8` (128 samples/rollout), `S=3`, 60 rollouts, PickScore. Two seeds, arms placed on different nodes to control for node effects.

| arm | median step | reward mean | rollouts with clipping |
|---|---|---|---|
| seed 42 off (serial) | 90 s | 0.8249 | 1/60 (max 0.02) |
| seed 42 on (batched) | 70 s | 0.8113 | 10/60 (max 0.15) |
| seed 43 off (serial) | 93 s | 0.8196 | 2/60 (max 0.02) |
| seed 43 on (batched) | 68 s | 0.8245 | 7/60 (max 0.04) |

**The reward gap changes sign between seeds** — `on − off` = −0.0136 (seed 42), +0.0048 (seed 43). Both arms rise from ~0.81 to ~0.83. So the difference is seed noise, not an effect of the flag. That is the empirical version of your structural argument.

End-to-end speedup at 8 GPUs: **1.29x / 1.37x** on total step time, consistent with the ~1.3x you measured at 2 GPUs.

## 2. `flux2_klein` — now GPU-validated

The stated blocker does not apply on this cluster: `FLUX.2-klein-4B` and `-9B` are both standard diffusers layouts with a `transformer/` subdir, so `from_pretrained` loads them directly.

FLUX.2-klein-4B, B=2, S=3, single H20:

| variant | parity max abs | ratio=1 | grad |
|---|---|---|---|
| dance (stage default) | 6.83e-04 | 0.0e+00 | PASS |
| dance + CFG 4.5 | 9.11e-04 | 0.0e+00 | PASS |
| flow | 9.95e-06 | 0.0e+00 | PASS |
| cps | 1.60e-05 | 0.0e+00 | PASS |
| dance + image-edit tiling | 2.11e-04 | 0.0e+00 | PASS |

The last row exercises `image_latent` / `image_latent_ids` tiling, which #242 makes live for Klein. FSDP: w2·S3 2.28x, w2·S6 2.80x, w4/w8 1.22–1.30x.

Dance's *relative* error looks large only because its log-prob is dominated by a per-step constant (|logp| ≈ 0.19); the absolute gap is < 1e-3. Step-major ordering is verified separately (cross-step discriminator 223–27551× the matched gap in every variant).

Also re-confirmed on the PR head, since all published numbers predate `6a03d0e` (mixin extraction), `faaff540` (z_image/flux2 config plumbing) and two `main` merges: qwen_image parity `2.38e-07` (tighter than the reported 1.03e-06), sd3 `2.98e-08`, z_image `1.81e-04`.

## 3. The memory cost is stated with the wrong sign

> Costs extra replay activation memory (the `[S*B]` stack).

True under `no_grad` (the π_old anchor). But the training replay runs **with gradients**, and there the ordering reverses: the serial loop keeps `S` separate autograd graphs alive, the batched path keeps one.

Single H20, B=1, peak replay memory (MiB), grad mode:

| model | S | serial | batched |
|---|---|---|---|
| qwen_image (20.4B) | 2 | 46717 | **38841** |
| qwen_image | 3 | **OOM** | **38858** |
| qwen_image | 4 | **OOM** | **38984** |
| flux2_klein | 12 | 54955 | 52131 |
| z_image | 6 | 69937 | 69237 |
| sd3 | 12 | 48268 | 48251 |

Batched is cheaper in **every** model at every `S` where both fit, and on the 20B model serial OOMs at `S≥3` while batched stays flat at ~38.9 GB. Where OOM does happen it happens for both paths at the same point. Under FSDP the per-rank delta is small (qwen +10 MiB, +0.5%).

This also undercuts the stated reason for excluding video/unified models ("stacking `[S*B]` multiplies an already-large activation footprint → likely OOM at full S"). At least for image models the grad-path footprint moves the other way, so that exclusion is worth re-measuring rather than assuming.

## 4. The world-size trend does not reproduce

Per-block `fully_shard`, `reshard_after_forward=True`, B=1, `ratio = 1` exact in all 24 rows:

| model | w2·S3 | w4·S3 | w8·S3 | w2·S6 | w4·S6 | w8·S6 |
|---|---|---|---|---|---|---|
| qwen_image (20.4B) | **2.96x** | 1.67x | 1.69x | **5.82x** | 2.03x | 2.04x |
| z_image (6.15B) | 2.09x | 1.20x | 1.21x | 2.55x | 1.28x | 1.28x |
| flux2_klein (3.88B) | 2.28x | 1.22x | 1.23x | 2.80x | 1.29x | 1.30x |
| sd3 (2.24B) | 2.60x | 1.60x | 1.56x | 3.15x | 1.78x | 1.83x |

The **S** trend is confirmed and is the cleanest evidence for the mechanism: qwen batched time is essentially flat in S (1083 ms at S=3 → 1101 ms at S=6) while serial doubles (3211 → 6411 ms).

The **world-size** trend is the opposite of "the win grows with world size" (sd3 w2 1.59x → w8 2.74x in the PR). Here w2 is the largest for every model, with w4 ≈ w8 below it. A plausible reading: on a single node the w2 all-gather pulls half the model across one peer link — the most expensive case — while at w4/w8 the transfer spreads over more NVLink peers and each all-gather gets cheaper, leaving less to save. The gain is real everywhere (1.20x – 5.82x), just not monotone in world size. Worth reconciling with your run before the claim ships.

Grad-mode replay at the recipe geometry (qwen, w8, S=3, B=1): serial 16468 ms vs batched 9862 ms → **1.67x**.

## 5. `old_logp_source='replay'` is a hard precondition, not a detail

Comparing replay log-probs against the **rollout-time** log-prob — i.e. what the ratio would be if the anchor were not sourced from replay:

| model / strategy | serial vs rollout | batched vs rollout |
|---|---|---|
| sd3 | 0.0e+00 | 0.0e+00 |
| qwen_image (flow) | 0.0e+00 | 2.4e-07 |
| flux2_klein (dance) | 0.0e+00 | 6.8e-04 |
| z_image (dance) | 0.0e+00 | 3.3e-03 |

Serial replay is bit-exact against the rollout for every model; batched is not, and under DanceSDE the deviation is 7–33× the recipe's `clip_range: 1.0e-4`. Since DanceSDE is `flux2_klein`'s stage default, this combination deserves an explicit note in the config docs rather than being implied by the gate.

Related, and visible in the A/B above: with `clip_range = 1e-4` the batched path clipped in 10/60 and 7/60 rollouts versus 1/60 and 2/60 for serial, once reaching 15% of samples (`ratio=1.0000±0.0001 clip=0.15`). This replicates across both seeds, unlike the reward difference. It does not change the reward outcome and the flag is default-off, but "the measured on-policy ratio stays exactly `1.0000`" is not literally true over a long run — it holds in every microbench configuration, and holds in the mean but not the spread in real training.

---

## Unrelated: Qwen-Image is broken on diffusers 0.39.0

Not caused by this PR, but I hit it while validating and it affects `main`. `unirl/models/qwen_image/diffusion.py` passes `txt_seq_lens=` to `QwenImageTransformer2DModel.forward()`; diffusers deprecated it and removed it in 0.39.0, while `pyproject.toml` pins `diffusers>=0.38.0` with no upper bound. A fresh install today resolves to 0.39.0 and Qwen-Image fails with `TypeError: ... got an unexpected keyword argument 'txt_seq_lens'` on **both** the serial and batched paths. I validated Qwen-Image on a 0.37 env instead. Filing this separately.

## Verdict

Nothing here blocks the merge. The correctness and `ratio = 1` claims hold on the current head across four models, three SDE strategies, CFG on/off, and the Klein image-edit path; the reward question is answered. I would suggest editing three things in the PR text before merge: the memory-cost direction (§3), the world-size claim (§4), and promoting `old_logp_source='replay'` to an explicit precondition (§5).

CjhHa1:
Follow-up: extended the reward-curve A/B to the other three image models, so the question is now answered for everything this PR touches rather than for Qwen-Image alone. **No measurable reward effect on any of them.** One claim from my previous comment does not survive the wider data and I am correcting it below.

Same protocol as before: the model's own `*_trainside` recipe (native `FSDPBackend`), 8×H20 per arm, `batch_size=16`, `samples_per_prompt=8` (128 samples/rollout), `num_sde_steps=3`, PickScore, recipe-default seed, both arms identical except `batch_replay_steps`.

## Reward

| model | n | reward off | reward on | diff (on−off) | per-rollout sd |
|---|---|---|---|---|---|
| sd3 (2.24B) | 60 | 0.8260 | 0.8255 | **−0.0005** | 0.027 |
| z_image (6.15B) | 60 | 0.8336 | 0.8327 | **−0.0009** | 0.015 |
| flux2_klein (3.88B) | 49 | 0.8403 | 0.8410 | **−0.0001** | 0.016 |
| qwen_image (20.4B) | 60 | 0.8249 | 0.8113 | −0.0136 | 0.017 |

For the three new models the gap is one to two orders of magnitude below the
per-rollout noise, and the smoothed curves are visually indistinguishable. sd3
is the most informative row because it actually learns over the window
(0.760 → 0.845) and the two arms track each other the whole way, so this is not
just "two flat curves agreeing".

Qwen-Image is the −0.0136 from before, which flipped to +0.0048 at seed 43 —
noise, as reported. Nothing here changes that reading.

flux2_klein's arms were cut short at 49 and 59 rollouts by Ray raylets being
OOM-killed on this cluster (unrelated to the flag — it hit both arms); the
comparison is aligned over the 49 rollouts both completed.

## End-to-end speedup (median step time)

| model | off | on | speedup |
|---|---|---|---|
| sd3 | 19 s | 13 s | **1.46x** |
| qwen_image | 90 s | 70 s | 1.29x |
| flux2_klein | 41 s | 36 s | 1.14x |
| z_image | 56 s | 55 s | 1.02x |

The spread is worth a note for anyone deciding whether to enable this. z_image
gets almost nothing end-to-end despite a 1.21x replay microbench at w8·S3,
because its step is dominated by rollout generation (12 inference steps) and
reward — the replay fraction it accelerates is small. The isolated replay
speedup is the right number for judging the change; the end-to-end number is
recipe-dependent and can be near 1.

## Correction to my previous comment

I wrote that with `clip_range = 1e-4` the batched path clips far more often
than serial, and framed it as a general consequence of the batched path not
being bit-reproducible. **That generalisation was wrong** — it was Qwen-Image
data extrapolated to a mechanism. Rollouts with any clipping:

| model | clip_range | off | on |
|---|---|---|---|
| qwen_image | 1.0e-4 | 1/60 | 10/60 |
| sd3 | 1.0e-4 | 0/60 | 1/60 |
| z_image | 1.0e-4 | 1/60 | 0/60 |
| flux2_klein | 5.0e-3 | 0/49 | 0/59 |

sd3 and z_image run at the **same** `clip_range = 1e-4` and show no difference
at all. So the effect is specific to Qwen-Image, not a property of batched
replay. Qwen-Image is both the largest model here and the only one whose text
conditioning is trimmed per call (the `txt_seq_lens` / RoPE-slice path), either
of which could plausibly explain it — I have not isolated which, and it may be
worth a look given that recipe's very tight clip range, but it is not a general
caveat and should not be read as one.

The other statements in my earlier comment are unaffected: `ratio = 1` exact in
24/24 FSDP configurations, the grad-mode memory ordering (§3), the world-size
trend (§4), and the `old_logp_source='replay'` precondition (§5) all still
stand. `ratio` printed as `1.0000` in every rollout of all eight arms here as
well.

## Net

The reward question is now closed for all four models this PR touches: enabling
`batch_replay_steps` does not change the learning curve on any of them, at a
1.02x–1.46x end-to-end saving depending on how replay-bound the recipe is.

celve (approved):

CjhHa1:
Follow-up to the validation above. I ran the experiment that isolates the clipping asymmetry, and the result was clear enough that I pushed a guard onto this branch — details at the bottom, please revert it if you'd rather handle it differently.

## The clipping was the anchor, not the batched path

My earlier A/B used `qwen_image_trainside`, which does **not** set `old_logp_source` — so it inherited `FlowGRPO`'s default of `'rollout'`. That means those runs were in exactly the configuration this PR's safety argument excludes: the π_old anchor came from the rollout (serial) forward while the train pass went through the batched one. Re-running with `+algorithm.old_logp_source=replay`, same recipe and scale (8×H20, 128 samples/rollout, S=3, 60 rollouts, `clip_range=1e-4`):

| anchor | arm | median step | reward | rollouts with clipping | max clip |
|---|---|---|---|---|---|
| `rollout` (default), seed 42 | serial | 90 s | 0.8249 | 1/60 | 0.02 |
| `rollout` (default), seed 42 | batched | 70 s | 0.8113 | **10/60** | 0.15 |
| `rollout` (default), seed 43 | serial | 93 s | 0.8196 | 2/60 | 0.02 |
| `rollout` (default), seed 43 | batched | 68 s | 0.8245 | **7/60** | 0.04 |
| `replay`, seed 42 | serial | 100 s | 0.8219 | 7/60 | 0.10 |
| `replay`, seed 42 | batched | 76 s | 0.8183 | **1/60** | 0.04 |

Switching the anchor to `replay` takes the batched arm from 10/60 and 7/60 down to **1/60** — and it then clips *less* than serial (7/60). So the batched path is not inherently less stable under the clip term; the excess clipping came entirely from anchoring against a forward it isn't bit-identical to. That is the empirical version of §5 in my previous comment.

Reward stays unaffected in all three configurations: `batched − serial` = −0.0136, +0.0048, −0.0036, i.e. the spread across seeds swamps the spread across arms. Speedup also holds throughout (1.29x / 1.37x / 1.32x on median step time).

Two caveats: the `replay`-anchor group is a single seed, and the serial arm's own clipping rose (1/60 → 7/60) when it switched to a replay anchor. The latter is plausibly the extra `prepare_segment` forward giving two independent chances to diverge rather than one, but I did not isolate it and it is not what this PR changes.

## Pushed: `8d7c8b71` — fail fast on the unsafe pairing

Given the above, the precondition is load-bearing but nothing enforced it, and the default lands on the wrong side of it. The commit adds `_require_replay_anchor_for_batched_replay` in `unirl/algorithms/base.py`, called from `FlowGRPO` and `FlowDPPO` (`BagelFlowUniGRPO` inherits it); `batch_replay_steps=True` with a rollout-sourced anchor now raises at algorithm construction with an actionable message. The three safe combinations are untouched, including the default `batch_replay_steps=False` path — verified across all eight `(flag × anchor × algorithm)` combinations. `cppo`/`dppo` are AR algorithms on `TextSegment` and cannot reach a diffusion stage's flag, so they are deliberately left alone.

It also records the measured per-model gaps in the `BatchedStepReplayMixin` docstring, so the flag can be judged against a recipe's `clip_range` rather than taken as uniformly safe:

| model | Flow | Dance / CFG |
|---|---|---|
| sd3 | 2e-7 | 3e-8 |
| qwen_image | 2e-7 | 1e-6 (CFG) |
| flux2_klein | 1e-5 | 9e-4 (Dance+CFG) |
| z_image | 1e-4 | 6e-3 (Flow+CFG) |

sd3 and qwen_image sit far below any usable `clip_range`; z_image and flux2_klein do not, which is worth knowing before enabling the flag on those two. The configs keep one-line pointers rather than repeating the block, matching the dedup in `1d4753d`.

`ruff check` / `ruff format` clean, lint CI green on the branch.

CjhHa1:
Closing the loop on the clipping observation I raised two comments up. **It is not a numerical defect in the batched path** — I was able to rule that out directly, so the finding should be read as training dynamics under a very tight `clip_range`, not as a problem with this PR.

## What the clipping is not

The obvious worry was that batched replay is less reproducible than serial, so the π_old anchor and the policy forward disagree. Measured directly, on real weights, at the recipes' geometry (B=1, S=3, activation checkpointing on) — the ratio the trainer would compute between an anchor pass under `no_grad` and a policy pass with autograd live:

| model | serial | batched |
|---|---|---|
| sd3 | 0.000e+00 | 0.000e+00 |
| z_image | 0.000e+00 | 0.000e+00 |
| qwen_image | 0.000e+00 | 0.000e+00 |

Bit-identical in every case, so update 1 can never clip and the batched forward is not the source. Three narrower candidates are dead too, each measured on qwen_image and sd3:

- **batch shape** — identical content at `[B]` vs `[S*B]`: `0.000e+00`. The shape change alone is bit-identical.
- **sigma as a scalar vs a per-sample vector** (the one interface difference batched replay always introduces): `0.000e+00`.
- **Qwen's per-call text trim / RoPE slice**, which I had floated as the likely culprit since no other model has it: with uniform-length masks — where the trim cannot vary — the discrepancy is if anything *larger* (3.09e-01 vs 2.83e-01), and sd3, which has no trim at all, shows the same magnitude (2.41e-01 either way). That residual is ordinary bf16 rounding from a different GEMM shape (B=1 vs B=4), equal across models. **The trim hypothesis is wrong.**

## What it is

Clipping comes from the second PPO mini-batch: `num_updates_per_batch: 2` (identical in all four recipes), where the weights have legitimately moved away from the ones π_old was computed with, judged against `clip_range = 1e-4`. Three lines of evidence:

- **It is late-onset.** No clipping in the early rollouts of any arm; qwen's first is rollout 27 (seed 42), sd3's only one is rollout 57, qwen's serial arm at seed 43 clips only at 56 and 59.
- **Within every arm, the rollouts that clip are the high-gradient ones** — 1.26x–1.77x the arm's typical grad norm. This holds in the **serial** arms too, which is the point: serial clips by the same mechanism, just less often.
- **Removing the second update removes the clipping.** Same batched arm, same seed, same node, 40 rollouts each, only `num_updates_per_batch` changed: `=2` → 1 rollout clipped; `=1` → 0.

So `clip_range = 1e-4` in the Qwen-Image and SD3 recipes is tight enough to resolve one optimizer step's worth of weight movement. That is worth knowing independently of this PR.

## What I did not settle

Why the batched arm crosses that threshold more often than serial *for Qwen-Image specifically* (10/60 and 7/60 vs 1/60 and 2/60). That is a trajectory-level difference — the two arms are on different optimization paths, seeded by the ~2e-7 log-prob gap — and I could not isolate it. Caveats on the probe above, in the interest of not overselling it: its control was weaker than the original run (1/40 clipped vs 4/40 in the first 40 rollouts of the original arm), and `num_updates_per_batch=1` also halves the optimizer steps, so the test arm sat at a lower gradient level (7.8e-05 vs 1.1e-04) — some of the "0 clips" is attributable to less total weight movement rather than to removing update 2 alone. Settling it properly needs per-update-index ratio logging at full precision, which is a trainer change rather than an experiment.

Also worth flagging for reproducibility: the `ratio` and `gn` fields are logged to four decimals, so `ratio=1.0000±0.0000` alongside `clip=0.02` is not a contradiction — the spread is simply below display precision. That rounding is why I lean on the clip fraction rather than on `ratio_sd` throughout.

## Net for this PR

Nothing here is a defect in `batch_replay_steps`, and nothing changes the merge picture: `ratio = 1` exact in 24/24 FSDP configurations, no reward effect on any of the four models, 1.02x–1.46x end-to-end. The clipping difference is a downstream symptom of the two arms diverging under an unusually tight clip range, and the flag is default-off regardless.

(Unrelated but it bit me mid-investigation: the pod's env got upgraded to diffusers 0.39 and Qwen-Image stopped running entirely, so these runs are on this branch with #288 cherry-picked. That is the bug #288 fixes.)

PR #158, opened 2026-07-01
feat(veomni-ep): expert parallelism for the VeOmni FSDP2 training backend

CjhHa1:

Current head includes three follow-up fixes from final review:

- `ep_size` now rejects booleans and floating-point values instead of coercing them.
- The example requires an explicit local `QWEN3_MOE_PATH` for both train and rollout.
- Qwen3-MoE initialization registers VeOmni's rewritten attention implementation in the correct order, without hiding `veomni.ops.apply_ops_config`.

Current-head validation supersedes the older counts below:

- The contributor's 12 focused EP cases passed before the PR-added test file was removed; the final PR diff contains no test files.
- Hydra compose succeeds with `QWEN3_MOE_PATH`; omitting it fails with the expected missing-environment error.
- Real CUDA/NCCL checkpoint + Adam round-trip: **4×H20, EP=2 passed**.
- Real random-init Qwen3-30B-A3B training step on the declared Torch 2.11 stack (using eager attention to isolate the EP/FSDP path; default FA2 registration and model construction were verified separately): **8×H20, EP=8 passed** through model build, FSDP2 wrapping, forward, backward, EP-aware gradient clipping, and AdamW step; peak allocated memory was **30.81 GB/GPU**.
- Targeted pre-commit and the seven-PR integration suite passed.

The Torch 2.7 failure encountered during validation was an environment mismatch: VeOmni 0.1.11 uses the Torch 2.8+ FSDP gradient-divide API on GPU. The declared UniRL engine stack pins Torch 2.11, where the same run passes.

## Summary

Adds opt-in training-side expert parallelism (EP) to `VeOmniBackend`, controlled by `backend.fsdp_cfg.ep_size` (`1` disables EP; values greater than `1` create the VeOmni `ep × ep_fsdp` mesh).

- Wires EP model preparation, parameter grouping, gradient clipping, and plan-declared expert sharding into the VeOmni backend.
- Loads both Hugging Face per-expert checkpoints and VeOmni stacked-expert checkpoints without materializing the full expert stack on each rank.
- Preserves the hidden EP dimension across single-file model/Adam checkpoint save and resume; unsafe full DCP checkpoints fail closed.
- Reconstructs full Qwen3-MoE expert tensors for rollout weight sync and converts them back to the Hugging Face per-expert names expected by SGLang.
- Adds the meta-init-only `Qwen3MoeBundle`, a GRPO recipe, and focused EP verification tooling.
- Validates `ep_size` without coercing `0` to `1`, fails on meta-init buffer shape drift, and updates the real-backend harness to exercise the current bundle/selective-import/RoPE recovery contracts.
- Removes `docs/veomni_ep_analysis_report.md` as requested; operational prerequisites remain self-contained in the recipe.

Important constraints:

- `ep_size` must divide `world_size` and the model's expert count.
- Full DCP checkpoints are intentionally rejected with active EP because DCP cannot encode the outer expert split; use the EP-aware single-file checkpoint path.
- The included colocated `TensorWeightSync` recipe requires rollout `tp_size=1`. Serving the 30B model with TP requires a TP-aware sync path (tracked separately).
- LoRA on EP-sharded fused expert parameters is rejected; attention/shared-module LoRA remains supported through the guarded sync paths.

## Test Plan

Current-head validation:

- The contributor's 12-case EP regression suite passed before removal (EP-size validation, plan-declared slicing on dimensions 0/1, checkpoint resume slicing, meta-init shape drift, HF↔fused layout).
- `PYTHONPATH=. python scripts/ep_verify/qwen3_moe_layout_roundtrip.py` — **PASS**, 12/12 tensors bit-exact.
- `PYTHONPATH=. DEVICE=cpu torchrun --standalone --nproc-per-node=4 scripts/ep_verify/unirl_ep_checkpoint_roundtrip.py 2 /tmp/unirl-pr158-ep-checkpoint.pt` — **PASS** with a real `ep=2 × ep_fsdp=2` DTensor mesh; distinct expert blocks and Adam moments survive save/resume.
- `DATA_PATH=/tmp/dummy.jsonl QWEN3_MOE_PATH=/tmp/qwen3-moe python -m unirl.train_ar --config-name=ar/qwen3_moe_grpo_30b_a3b_veomni_ep_sglang --cfg job --resolve` — recipe composes and resolves all targets.
- `python -m compileall -q unirl scripts/ep_verify` — passed on the final source tree.
- `pre-commit run --all-files --show-diff-on-failure` — all hooks passed.

Existing contributor GPU validation (reported before the final hardening commit; core EP math/load/sync paths are unchanged):

- Real `VeOmniBackend` on 1×8 H20: EP1/EP2/EP4 loss trajectories matched; EP4 peak memory decreased from 12.12 GB to 9.90 GB.
- Real GRPO replay/backward/optimizer step: EP4 and EP1 step-0 policy loss matched (`-0.51628`), ratio `1.0`, with lower peak memory.
- EP full-weight sync at EP4 emitted 1536/1536 expert keys; 24/24 sampled tensors matched the source checkpoint bit-exactly.
- A paired 2×8 H20, 40-rollout EP1-vs-EP8 run reported mean train reward `0.0897` vs `0.0916` (paired 95% CI for the difference `[-0.0043, +0.0082]`) and no on-policy invariant regression.

## Compatibility / Risk

The feature remains opt-in (`ep_size=1` by default). The highest-risk paths—initial expert load, model/optimizer checkpoint round-trip, rollout name/layout conversion, LoRA combinations, and unsupported DCP usage—now have explicit conversion or fail-closed guards. The 30B production topology still depends on the separate rollout TP/EP work noted above.


celve:
Do you have the reward curve?

unirl/train/configs.py
```diff
@@ -98,6 +98,18 @@ class FSDPConfig:
     # Must divide the world size and the model's attention head count. Only the
     # VeOmni backend honors it; FSDPBackend ignores it.
     sp_size: int = 1
+    # Expert-parallel degree for MoE models (default 1 = disabled, a true no-op).
```
zzhuoxin1508, line 114: Can you simplify the docstrings?

CjhHa1:
> Do you have the reward curve?

image

unirl/train/configs.py
CjhHa1: Done

zzhuoxin1508:
**EP checkpoint save/resume silently corrupts expert weights and optimizer state**

Under `ep_size > 1`, the ep split is baked into the expert params' DTensor global shape (`[E/ep, …]`), not expressed as a placement, so DCP cannot see it. But checkpoint save/resume still goes through the generic DCP full-state path:

- **Save**: `full_state_dict=True` gathers only over each group's own mesh, so the 8 ep groups produce same-FQN but different-content `[E/ep]` blocks, and only rank 0's (group 0's) block is written — the other groups' experts are never saved.
- **Resume**: `broadcast_from_rank0=True` pushes group 0's block to all ep groups; shapes match exactly, so it overwrites silently.

Net effect: at `ep_size=8`, any save→resume cycle silently corrupts 7/8 of the expert weights and Adam moments, with no error. The PR already routes the initial pretrained load around `set_model_state_dict` for exactly this reason ("torch's rank-0-broadcast loader mis-slices that layout"), but the resume path (`base_backend.py:505`, and `VeOmniBackend._gather_optimizer_state` / `_load_optimizer_state`) still uses that machinery.

zzhuoxin1508:
A few more issues, roughly in severity order:

1. **EP + LoRA weight sync is broken on both branches, with no guard.** With `lora_merged=False`, `_iter_full_tensors_ep` iterates raw `model.state_dict()` names without the PEFT handling `raw_state_dict` performs (prefix strip, `.base_layer.` flatten, `lora_A/B` skip), so SGLang receives unrecognized names. With `lora_merged=True` the EP walk is bypassed and `merged_state_dict` pushes the fused `...experts.gate_up_proj` name holding only this ep group's `[E/ep]` slice. Nothing validates/rejects `ep_size>1 + lora_cfg`, so the combination fails silently.

2. **`qwen3_moe/bundle.py` imports veomni directly, bypassing `_compat`.** `_compat.py`'s contract is that it is the only place importing veomni, and `ensure_installed()` must run before any veomni symbol. The bundle is instantiated before the backend, so `from veomni.models.auto import ...` executes the real `veomni/__init__` (global `_apply_patches()`) and `veomni/models/__init__` (transformers-version-locked model zoo). The other new files in this PR all call `ensure_installed()` first.

3. **The EP-presliced dim in `_load_state_dict_ep_sliced` is inferred by a shape-ratio heuristic.** It applies to every shape-mismatched param (not just experts), takes the first dim where `ckpt == global * ep_size` (a coincidental match silently loads a wrong slice), and when no dim matches it silently falls back to `sl[:]`, producing an opaque downstream shape error instead of a clear mismatch message. `get_parallel_plan()` already declares which params are EP-sharded and on which dim.

4. **Short dtype names are silently coerced to bf16.** The `torch_dtype` whitelist in `from_config` only accepts long names, but `parse_torch_dtype`'s canonical inputs are `bf16/fp16/fp32` — so `model_precision: fp16`/`fp32` builds the model as bfloat16 while `bundle.dtype` says otherwise, with no error.

5. **The bundle diverges from the add-model-bundle conventions.** It hardcodes meta-init instead of honoring `meta_init_transformer`, and hand-rolls RoPE `inv_freq` recovery (swallowing all exceptions — a failure leaves garbage buffers silently) instead of reusing the `meta_init.py` machinery the dense Qwen3 bundle uses; no tests are added.

6. **A second EP expert loader now exists in parallel with `veomni/ep/load.py::load_ep_experts` (HI3 path).** The two implement the same job with divergent strategies (per-rank file reads vs rank-0 broadcast), and the HF↔fused expert layout mapping is now encoded in three places (`_build_expert_block_from_split`, `_iter_full_tensors_ep`'s reverse split, VeOmni's converter) — any drift between them is silent weight corruption.

zzhuoxin1508 (approved):
docs/veomni_ep_analysis_report.md should remove this file

leviking98z-rgb (commented):
Reviewed the current head after the main-branch merge and maintainer fixes. The requested analysis report is removed; ep_size validation now rejects zero/invalid degrees; meta-init restoration fails closed on shape drift; and the real-backend harness follows the current EP preparation, selective-import, and RoPE-recovery contracts. Targeted CPU tests, the 4-process ep×ep_fsdp model/Adam checkpoint round-trip, recipe composition, full pre-commit, and all GitHub checks pass. I found no remaining code-level blocker; the documented rollout TP-aware sync dependency remains an operational constraint rather than a regression in this training-side EP change.

PR #74, opened 2026-06-15
feat(rollout): canonical sglang LLM/VLM rollout engine, retire sglang_llm

celve:
## Summary

Introduces the role-decomposed **`sglang`** LLM/VLM rollout engine as the canonical SGLang autoregressive engine, and retires the legacy HTTP-only **`sglang_llm`**.

`sglang` is a thin core over a runtime seam (`backends` — HTTP SRT-server impl and in-process native `sglang.Engine`), with `adapters` (text base + VLM, keyed by `model_family` / derived from `image_token`), a `WeightSync` component (sync ops + LoRA lifecycle), and self-reserved ports. Migrated in-tree recipes pin `backend: native`; `backend: http` remains available as the separate SRT-server transport.

What changed:
- Add the `sglang` engine (`unirl/rollout/engine/sglang/`): backend seam (HTTP + native), text + VLM adapters, `WeightSync`, engine core, and self-reserved ports.
- Point every in-tree AR, CPPO/DRPO, and composed-PE recipe that used `sglang_llm` at `sglang` with `backend: native`; `model_family` auto-derives from `image_token`.
- Remove the legacy `sglang_llm` package — nothing imports it (the composed engine builds children generically from their `_target_`).
- Carry forward the latest `main`/review fixes: HTTP backend reuses a persistent asyncio loop (#39), VLM image batches are validated before generation, and SGLang server ports are reserved below the runtime's derived gRPC overflow limit (`port + 30000 <= 65535`).
- Update docs/comments so `sglang` = the LLM/VLM engine and `sglang_diffusion` = the diffusion engine (the bare "sglang" shorthand was the diffusion engine's old name).

Mirrors the canonical-engine / retire-v1 pattern of #36 (vllm_omni).

## Related Issue

N/A

## Test Plan

- **Hydra config-load** — all migrated recipes compose and instantiate `SGLangEngineConfig(backend=native)`:
  `python -m unirl.train_ar --config-name=ar/<recipe> --cfg job --resolve` (and `unirl.train_pe` for the composed PE recipes). Verified for the migrated AR recipes + `CPPO`/`DRPO` + the composed PE recipes.
- **Native API smoke** (in-process Engine): generate, LoRA push + rotation, sleep/wake, onload, shutdown — `scripts/lin381_native_smoke.sh` (driver `--backend native`, including `--vlm-model` for the VLM adapter).
[...]

PR #163, opened 2026-07-01
fix(sglang weight-sync): SafeUnpickler allowlist + fused-shard/rename mapping

leviking98z-rgb:
## Summary

Model-agnostic **full-weight-sync fixes** for the `sglang_diffusion` engine, split out of the video-rollout PR (#135) — they are unrelated to video output; any full-FT + sglang run needs them.

- **`patch_safe_unpickler` (new)**: sglang's `SafeUnpickler` (the CVE-2025-10164 mitigation in `srt/utils/common.py`) allowlists `torch.` / `peft.` / `sglang.srt.*` but **not** `unirl.`. Full-weight sync ships serialized CUDA tensors whose reduction references unirl's own `_rebuild_cuda_tensor_modified` (a cross-node rebuild that swaps device index for device UUID, working around pytorch/pytorch#149248), so `update_weights_from_tensor` raises `Blocked unsafe class loading (unirl..._rebuild_cuda_tensor_modified)` and the run dies at the first weight push. This adds the trusted `unirl.` prefix to the allowlist (import-safe, idempotent, no-op on older sglang). The payload is unirl-internal and trusted — this does not weaken the CVE mitigation against untrusted data.
- **`patch_weights_updater`**: `_apply_fused_param_mapping` now also applies a model's **plain string renames** and **empty-string drops** from `param_names_mapping`, not just fused (tuple) entries — rename-based models (WAN) otherwise matched nothing in the in-memory update path → the rollout engine ran **stale base weights** (flat reward / cross-engine divergence). `_write_fused_shard` handles the **trailing-unequal** layout (HunyuanVideo `linear1=[q,k,v,mlp]`) via tail placement.
- **`hijack`**: register `patch_safe_unpickler` in the patch suite.

## Background

**Why it surfaced now.** The sglang `SafeUnpickler` lock landed upstream on 2025-10-24 (CVE-2025-10164) and has always been present in this fork; the trigger is new — unirl PR #74 (2026-06-19) rewrote weight-sync and introduced the custom `_rebuild_cuda_tensor_modified`. So the lock was never the change — the payload that trips it is.

**Why #74's tests missed it.** #74 is an LLM-engine PR whose test plan exercised LoRA push, while the bug lives in the shared full-weight-sync layer that only `sglang_diffusion` full-FT walks; #74 also dropped its CPU unit suite and didn't rerun GPU e2e after final fixes. Notably #74 *did* allowlist `unirl.` on unirl's own unpickler — it just omitted the symmetric sglang-side allowlist, which this PR completes.

**What actually trips it (measured).** The block is caused specifically by unirl's own full-weight-sync path (`full/tensor.py` + `sgl_compat.monkey_patch_torch_reductions`), whose reduction pickles a reference into the non-allowlisted `unirl.` namespace. LoRA and sglang's official `update_weights_from_tensor` path use sglang's own rebuild (`sglang.srt.*`, allowlisted) and do NOT trip it — so this is about *which rebuild is used*, not LLM-vs-diffusion.

## Test Plan

- The `SafeUnpickler` block was **reproduced directly on the deployed sglang (`0.5.12.post1`)**: a pickle whose reduction references `unirl...._rebuild_cuda_tensor_modified` raises the exact `Blocked unsafe class loading` error without the patch; with the `unirl.` prefix added, the allowlist check passes.
- **Path comparison verified on the deployed engine**: after unirl's `monkey_patch_torch_reductions`, `reductions.rebuild_cuda_tensor` points into `unirl.` (blocked); after sglang's official patch it points into `sglang.srt.*` (allowlisted, not blocked) — confirming only the unirl full-weight-sync path is affected.
- `py_compile` + `ruff` (check + format) pass on all three changed files.

### DIFF
--- a/unirl/rollout/engine/sglang_diffusion/_patches/hijack.py
+++ b/unirl/rollout/engine/sglang_diffusion/_patches/hijack.py
@@ -153,6 +153,9 @@ class SglangDiffusionHijack:
         from unirl.rollout.engine.sglang_diffusion._patches.patch_rollout_trajectory import (
             patch_rollout_trajectory,
         )
+        from unirl.rollout.engine.sglang_diffusion._patches.patch_safe_unpickler import (
+            patch_safe_unpickler,
+        )
         from unirl.rollout.engine.sglang_diffusion._patches.patch_sampling_io import (
             patch_sampling_io,
         )
@@ -198,5 +201,6 @@ class SglangDiffusionHijack:
             patch_dance,
             patch_set_timesteps,
             patch_vae_decode_safe,
+            patch_safe_unpickler,
         ):
             _safe_apply(patch)
--- a/unirl/rollout/engine/sglang_diffusion/_patches/patch_safe_unpickler.py
+++ b/unirl/rollout/engine/sglang_diffusion/_patches/patch_safe_unpickler.py
@@ -1,0 +1,33 @@
+"""Allow unirl's trusted tensor-rebuild classes through sglang's SafeUnpickler.
+
+Full-weight sync ships serialized CUDA tensors whose reduction references unirl's
+``_rebuild_cuda_tensor_modified`` (``weight_sync.transfer.sgl_compat``). sglang's
+own ``SafeUnpickler`` (``srt/utils/common.py``, the CVE-2025-10164 mitigation)
+allowlists only ``builtins``/``torch``/``multiprocessing``/… — NOT ``unirl.`` —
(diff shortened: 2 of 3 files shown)


leviking98z-rgb:
Split into three independent PRs — the three fixes are semantically, logically, and code-wise independent (different layers: transport / naming / write-slice; different failure modes; zero cross-reference):
- #164 SafeUnpickler allowlist
- #165 trailing-unequal fused shard placement
- #166 plain-rename/drop param_names_mapping entries

PR #165, opened 2026-07-01
fix(sglang weight-sync): trailing-unequal fused shard placement (HunyuanVideo linear1)

leviking98z-rgb:
## Summary

`_write_fused_shard` writes a trainer's separate-projection tensor into a dim-0 slice of a model's fused param. It split dim 0 into `num_shards` **equal** slices — correct only for all-equal fusions (`q==k==v` in Z-Image `to_qkv`, `w1==w3` in `w13`).

HunyuanVideo's single-block `linear1 = [q, k, v, mlp]` packs three `H`-sized attention shards + one `4H`-sized MLP shard. The equal split carves four `1.75H` chunks and **crashes** writing an `H` tensor into a `1.75H` slot.

Fix: place leading shards at `shard_id * size` and the (possibly larger) trailing shard at the tail (`dim0 - size`). This is exact for **both** layouts — for all-equal fusions the two formulas coincide (`(n-1)*size == dim0 - size`) — needs no sibling shards (robust to the sender's bucketing, since a block's q/k/v/mlp may arrive in different buckets), and adds a fits-in-param bounds check. The param's own `weight_loader` is still tried first when it accepts a shard id (TP-correct). It deliberately does not support an unequal *middle* shard (no SGLang fused param has one).

## Test Plan

- `py_compile` + `ruff` (check + format) pass.
- Placement math verified for the two known layouts (all-equal q/k/v; trailing-unequal `linear1` with the MLP shard landing at the tail), guarded by the fits-in-param bounds check.


zzhuoxin1508 (approved):
LGTM

PR #166, opened 2026-07-01
fix(sglang weight-sync): apply plain-rename/drop param_names_mapping entries (WAN stale weights)

leviking98z-rgb:
## Summary

`_apply_fused_param_mapping` aligns the trainer's incoming named tensors to the model's params via `param_names_mapping`. It only handled **fused-projection** entries (`{regex: (replacement, shard_id, num_shards)}`) and skipped everything else. But a model's `param_names_mapping` (the same dict its checkpoint loader applies) also has:

- **plain renames** — `{regex: replacement_str}` — the model simply renamed a param vs the checkpoint. WAN's mapping is entirely this kind (`blocks.N.attn1.to_q.* -> blocks.N.to_q.*`, `ffn.net.0.proj -> ffn.fc_in`, …).
- **drops** — empty replacement — the model dropped that param (e.g. WAN `attn2.norm_added_q`).

So rename-based models matched **nothing** in the in-memory update path → 112/113 transformer tensors silently skipped → the rollout engine ran **stale base weights** → flat reward curve (cross-engine divergence), exactly the fused-model bug one step removed.

Fix: handle both string kinds — `re.sub` rename into the leftover exact-match list, and empty-string drop — alongside the existing fused (tuple) path. Log `fused`/`renamed`/`dropped` counts.

## Safety warning

A leftover tensor that is still not a real model param after every mapping branch (an unmatched pattern, or a rename whose target does not exist) would **silently no-op** in the exact-match loader — the same stale-weight failure this mapping prevents. So the function now **warns** at the end for any such unmatched tensor instead of dropping it silently. (Log only; no behavior change on the normal path.)

## Test Plan

- `py_compile` + `ruff` (check + format) pass.
- Independent of the fused-shard placement change: rename-based models (WAN) do not exercise the fused (tuple) branch at all.


zzhuoxin1508 (approved):
LGTM

PR #128, opened 2026-06-26
feat(qwen-image-edit-plus): DiffusionNFT + FlowGRPO on sglang_diffusion

NancyFyong:
## Summary

Add **Qwen-Image-Edit-2511** (Qwen-Image-Edit-Plus) as a first-class image-editing model in UniRL, with **DiffusionNFT, FlowGRPO, and FlowDPPO training verified end-to-end on the `sglang_diffusion` colocate backend**. This is the first image-editing pipeline in the tree — source-image latents are fed to the transformer via token concatenation, exercising the conditions plumbing in ways existing T2I models do not.

- **Model bundle** (`unirl/models/qwen_image_edit_plus/`): bundle, config, conditions, VAE (per-channel norm + 5D encode), diffusion step with source-latent token concat, and `QwenImageEditPlusPipeline`.
- **sglang_diffusion adapter** (`unirl/rollout/engine/sglang_diffusion/adapters/qwen_image_edit_plus.py` + `_patches/`): source-latent token concat during sampling, conditions routing, `LocalLoraWeightSync` for NFT's dual-adapter EMA.
- **Trainsite pipeline** (`unirl/models/qwen_image_edit_plus/pipeline.py`): replay-side mirror of the rollout `predict_noise` for NFT's forward-process K-loop.
- **Recipes** (`examples/diffusion/qwen_image_edit_plus/`): NFT sglang (primary, validated), FlowGRPO sglang (validated), FlowDPPO sglang (validated), NFT trainsite (validated).
- **Build**: `pyproject.toml` raises sglang's `diffusers` floor 0.37 → 0.38 (required by Edit-Plus pipeline).
- **vllm_omni adapter** split into a follow-up PR (feat/qwen-image-edit-vllm-omni).

## Related Issue

N/A

## Test Plan

**Validation matrix (Qwen-Image-Edit-2511):**

| Algorithm | Backend | Status |
|---|---|---|
| DiffusionNFT | sglang_diffusion | ✅ Verified (390-rollout long run) |
| FlowGRPO | sglang_diffusion | ✅ Verified |
| FlowDPPO | sglang_diffusion | ✅ Verified |
| DiffusionNFT | trainsite | ✅ Verified |

**DiffusionNFT long-run (sglang_diffusion, primary):**
- 8× H20, Qwen-Image-Edit-Plus 20.43B, PickScore reward, `qwen_image_edit_plus_nft_sglang.yaml`
[...]

PR #172, opened 2026-07-03
fix(qwen-image-edit-plus): harden sglang edit-plus rollout + doc follow-ups

xshrz:
## Summary
Follow-up to #128 (merged) addressing review findings — three low-risk
robustness/correctness fixes plus doc corrections in the Qwen-Image-Edit-Plus
sglang path:
- **Deterministic SDE fallback seed** (`_patches/patch_denoising.py`): the
  single-`torch.Generator` fallback (hit by Qwen-Image / Edit-Plus FlowGRPO /
  FlowDPPO, whose denoising stage overrides `_run_denoising_step`) seeded z_t
  from `os.urandom`, which is non-reproducible. It now derives the per-request
  seed from `(base_seed, denoise_seeds[0])` — the per-sample-unique `sample_id`
  that `patch_latent_prep` slices onto each B=1 output — via the same blake2b
  derivation as `_make_step_generators`. This keeps the per-sample independence
  that fixes the reward regression while restoring fixed-seed reproducibility;
  `os.urandom` remains only as a last resort when no stable key is available.
- **Encode-dedup value check** (`_patches/patch_conditions.py`):
  `_coalesce_duplicate_single_sample_encodes` now verifies entries are
  byte-identical (`torch.equal`) before collapsing, so a future same-shape
  multi-encoder can't have an encoder silently dropped (the shape-only heuristic
  documented in `examples/.../README.md`). Behavior-preserving for all current
  models.
- **Group-consistent source-image collapse**
  (`adapters/qwen_image_edit_plus.py`): collapse one source image per group in
  first-seen group order (mirroring `deexpand_prompts_from_groups`) instead of
  `pil_images[::k]`, which assumed a contiguous group-major layout and would
  misalign images vs prompts for interleaved `group_ids`. Equivalent to the old
  behavior for the validated contiguous layout; adds a count-consistency guard.
- **Docstring corrections** (`diffusion.py`, `conditions.py`): the module
  docstrings described a T2I fall-through that no longer exists — Edit-Plus is
  edit-only and `predict_noise` raises when `image_latent` is `None`.
- **Comment cleanup** (`nft.yaml`, `nft_sglang.yaml`): drop the internal
  `-merged_v9.7` checkpoint tag (default is the public `Qwen/Qwen-Image-Edit-2511`).
## Related Issue
N/A
## Test Plan
- `SKIP=no-commit-to-branch pre-commit run --files unirl/rollout/engine/sglang_diffusion/_patches/patch_denoising.py unirl/rollout/engine/sglang_diffusion/_patches/patch_conditions.py unirl/rollout/engine/sglang_diffusion/adapters/qwen_image_edit_plus.py unirl/models/qwen_image_edit_plus/diffusion.py unirl/models/qwen_image_edit_plus/conditions.py examples/diffusion/qwen_image_edit_plus/qwen_image_edit_plus_nft.yaml examples/diffusion/qwen_image_edit_plus/qwen_image_edit_plus_nft_sglang.yaml --show-diff-on-failure` → all hooks Passed (ruff-check, ruff-format, check-yaml, recipe `_target_` resolver).
- `python3.12 -m py_compile` on the changed modules → OK.
- GPU E2E: **Not run; reason:** no GPU available in this environment. The seed change preserves the per-sample independence property of the merged `os.urandom` fix (distinct `sample_id`s → distinct seeds), so GRPO exploration is equivalent while runs become reproducible; a FlowGRPO/FlowDPPO sglang short run is recommended to confirm the reward curve is unaffected. The `torch.equal` and group-collapse changes only tighten guards on paths the validated configs don't exercise (N=1 per generate; contiguous group layout).
## Compatibility / Risk
- The fallback seed change alters the exact z_t values drawn on the sglang
  single-generator path (Qwen-Image / Edit-Plus FlowGRPO / FlowDPPO) but keeps
  per-sample independence and makes runs reproducible. DiffusionNFT (eta=0,
  num_sde_steps=0) never draws z_t, so it is unaffected.
- The dedup and image-collapse changes are stricter guards, behavior-preserving
  for current models/recipes.
- No config keys, checkpoints, data formats, or public APIs change.
## Reviewer Notes
N/A
## Checklist

### DIFF
--- a/examples/diffusion/qwen_image_edit_plus/qwen_image_edit_plus_nft.yaml
+++ b/examples/diffusion/qwen_image_edit_plus/qwen_image_edit_plus_nft.yaml
@@ -105,7 +105,7 @@ backend:
   # the v1 training.ema_lora block. target_modules keep the v1 DiffusionNFT set: bare
   # (un-prefixed) names that also cover the img/txt MLP projections — broader
   # than the GRPO recipe's attn-only `attn.*` set. Verified verbatim against
-  # transformer/model.safetensors.index.json (Qwen-Image-Edit-2511-merged_v9.7).
+  # transformer/model.safetensors.index.json (Qwen-Image-Edit-2511).
   ema_lora_cfg:
     _target_: unirl.train.configs.EmaLoraConfig
     rank: 64
--- a/examples/diffusion/qwen_image_edit_plus/qwen_image_edit_plus_nft_sglang.yaml
+++ b/examples/diffusion/qwen_image_edit_plus/qwen_image_edit_plus_nft_sglang.yaml
@@ -131,7 +131,7 @@ backend:
   # the v1 training.ema_lora block. target_modules keep the v1 DiffusionNFT set: bare
   # (un-prefixed) names that also cover the img/txt MLP projections — broader
   # than the GRPO recipe's attn-only `attn.*` set. Verified verbatim against
-  # transformer/model.safetensors.index.json (Qwen-Image-Edit-2511-merged_v9.7).
+  # transformer/model.safetensors.index.json (Qwen-Image-Edit-2511).
   ema_lora_cfg:
     _target_: unirl.train.configs.EmaLoraConfig
     rank: 64
--- a/unirl/models/qwen_image_edit_plus/conditions.py
+++ b/unirl/models/qwen_image_edit_plus/conditions.py
@@ -8,9 +8,10 @@ dimension before the transformer call, then slices the prediction back to the
 noise segment — mirrors ``vde_editplus.py:232,246`` and the FLUX.2-Klein
 image-edit pattern (``flux2_klein/diffusion.py:160-183``).
 
-``image_latent`` is optional so the conditions container round-trips cleanly
-for the T2I degenerate path (no source image); the Edit-Plus pipeline itself
-requires a source image and raises at ``generate(req)`` time if absent
+``image_latent`` is declared ``Optional`` only so ``from_dict`` / ``to_dict``
+round-trip cleanly when the key is absent from a raw dict; Edit-Plus is
+edit-only, so both the pipeline (``generate(req)``) and the diffusion step
+(``predict_noise``) require a source image and raise if it is missing
 (fail-fast, constraint #27).
 """
 
--- a/unirl/models/qwen_image_edit_plus/diffusion.py
+++ b/unirl/models/qwen_image_edit_plus/diffusion.py
@@ -18,10 +18,10 @@ overridden Edit-Plus step). Verified: ``step_with_logp`` → ``step`` →
 ``self.predict_noise`` (``qwen_image/diffusion.py:326→349→304``), so the
 override propagates to ``replay`` automatically.
 
-When ``conditions.image_latent is None`` the step falls through to the
-base T2I behavior — a genuine degenerate path, not impossible-scenario
-handling (the base Qwen-Image T2I is a valid input mode for this
-transformer).
+Edit-Plus is edit-only: :meth:`predict_noise` requires
+``conditions.image_latent`` and raises ``ValueError`` when it is ``None``
+(fail-fast, constraint #27). There is no T2I fall-through — a source image
+is always required.
 """
 
 from __future__ import annotations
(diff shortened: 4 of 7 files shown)


leviking98z-rgb:
## Review

Reviewed all five changes against the current `main`. The correctness reasoning holds throughout; the changes are low-risk and well-scoped as described. One merge conflict to resolve (mechanical) and a couple of minor notes.

### ✅ Verified correct

**1. Deterministic SDE fallback seed (`patch_denoising.py`).** `_resolve_fallback_seed` uses the *same* blake2b derivation as `_make_step_generators` (`digest_size=8`, `% _MAX_TORCH_SEED`, big-endian) with a distinct `::fallback::sample::` namespace vs the per-step `::step::N::sample::` — so a fallback seed can't collide with a per-step generator seed. Falls back to `os.urandom` only when neither `base_seed` nor `denoise_seeds[0]` is available. Per-sample independence (distinct `sample_id` → distinct seed) is preserved, and runs become reproducible. The single generator is stashed on `batch` and advances across SDE steps within a request — consistent with the original `os.urandom`-once design. DiffusionNFT (eta=0) never hits this path, as noted.

**2. Group-consistent source-image collapse (`qwen_image_edit_plus.py`).** `_first_per_group` (first item per gid, first-seen order) exactly mirrors `deexpand_prompts_from_groups`, which builds `unique_prompts` by iterating `order` and taking `prompts[idxs[0]]`. Both `_deexpand_prompts` and `_first_per_group` operate on the same `list(req.group_ids)`, so images and prompts now collapse in lockstep under interleaved group_ids — the old `pil_images[::k]` genuinely assumed contiguous group-major layout. The `len(unique_pils) != len(unique_prompts)` guard is a correct (belt-and-suspenders) safety net. Equivalent to the old behavior for the validated contiguous layout.

**3. Encode-dedup value check (`patch_conditions.py`).** Adding `torch.equal` before collapsing is the right hardening: shape-only matching could silently drop a distinct-but-same-shape encoder output. Behavior-preserving for current models (the duplicates are genuinely byte-identical shared-list encodes).

**4. Docstring corrections (`diffusion.py`, `conditions.py`).** Accurate — `main`'s `predict_noise` raises `ValueError` when `image_latent is None` (diffusion.py:95-97), so the old "T2I fall-through" docstring was stale. The rewrite matches the code.

**5. Comment cleanup (yamls).** Dropping the internal `-merged_v9.7` tag is correct; the default is the public `Qwen/Qwen-Image-Edit-2511`.

### ⚠️ Needs resolution: merge conflict in `patch_conditions.py`

`main` has since **rewritten** `_coalesce_duplicate_single_sample_encodes` (compressed docstring; refactored to `first = value[0]` + shape-only check + `return [first]`, dropping the explicit `None`-holes branch in favor of `all(torch.is_tensor(t))`). The PR's `torch.equal` check was inserted into the *old* structure, so it conflicts. The fix is mechanical — drop the guard in just before `main`'s `return [first]`:

```python
    if any(tuple(t.shape) != first_shape for t in value[1:]):
        return value
    if not all(torch.equal(first, t) for t in value[1:]):   # <-- PR's guard, adapted
        return value
    return [first]
```

The intent is fully compatible with `main`'s rewrite; please rebase and re-apply the one-liner there. The other four files auto-merge cleanly.

### Minor notes (non-blocking)

- The count-consistency guard in `build_prompts` can't actually trip given `_deexpand_prompts` and `_first_per_group` share `req.group_ids` and identical semantics — it's defensive, which is fine, but worth knowing it's unreachable under the current call path.
- Test plan is reasonable for the risk level (pre-commit + py_compile; no GPU). A FlowGRPO/FlowDPPO sglang short run to confirm the reward curve is unaffected by the seed change would be the one thing worth doing before relying on it in a real run, as the PR itself suggests.

Overall: correct and safe to merge once rebased past the `patch_conditions.py` rewrite.

leviking98z-rgb:
Rebased onto latest `main` to clear the `patch_conditions.py` conflict (main had rewritten `_coalesce_duplicate_single_sample_encodes` to the `value`/`first` form). Re-applied the `torch.equal` value check on top of main's structure:

```python
    if any(tuple(t.shape) != first_shape for t in value[1:]):
        return value
    if not all(torch.equal(first, t) for t in value[1:]):  # value check, adapted to main's rename
        return value
    return [first]
```

Force-pushed (author preserved as @xshrz). The other 6 files auto-merged; no behavior change. Now MERGEABLE.

leviking98z-rgb (approved):

PR #175, opened 2026-07-03
fix(bagel): stabilize shared VAE dtype across decode and encode

CjhHa1:
## Summary

- Keep the shared BAGEL VAE in fp32 after its first decode. This makes the module state stable across `activation_checkpoint` forward/backward recomputation instead of restoring bf16 before backward and being re-upcast during recompute.
- Make image-edit prefill dtype-safe at the vendor encode boundary: cast encode inputs to the VAE's live parameter dtype, and cast output latents to the live `vae2llm` consumer dtype. This avoids assuming `bundle.vae_dtype` and the consumer precision are identical.

## Verification

- Full pre-commit suite passed (AST, Ruff, formatting, recipe target resolution, and repository guards).
- Real BAGEL-7B-MoT checkpoint on one NVIDIA H20 (`torch 2.11.0+cu130`): two consecutive 512x512, 14-step it2i rollouts passed with finite pixels, trajectory shape `[1, 5, 1024, 64]`, and SDE log-prob shape `[1, 3]`.
- Checkpointed VAE decode/backward from an initial bf16 VAE produced finite latent gradients and left the VAE fp32 after both forward and recomputation.

## Ablation

Controlled on the same H20 and loaded model, 512x512 / 14 steps / three paired seeds:

- Old restore behavior vs sticky-fp32 + encode-boundary cast was numerically identical: max absolute delta was `0` for decoded pixels, stored latents, and SDE log-probs in every pair.
- End-to-end time was statistically unchanged at this sample size: restore `2.676 s` mean vs sticky `2.689 s` mean (`-0.49%`, noise-level). This PR therefore makes no end-to-end throughput claim.
- Isolated VAE dtype round-trip cost `5.02 ms` mean; the steady fp32 no-op cost `0.87 ms`, removing about `4.15 ms` per decode.
- Sticky fp32 added `158.15 MiB` steady allocated memory. Peak allocation was effectively unchanged (`~29.67 GiB`) because the restore path also upcasts during decode.
- Without the encode-boundary cast, a warm fp32 VAE emitted fp32 latents and the bf16 `vae2llm` projection failed with `mat1 and mat2 must have the same dtype`; with the cast, latents were bf16 and the projection was finite.

The encode boundary itself works without autocast. A complete CUDA no-autocast vendor cache update remains outside this PR's scope: the existing vendored `TimestepEmbedder` constructs fp32 sinusoidal features for a bf16 MLP. The production BAGEL path uses bf16 autocast, and vendor code remains untouched.

## Test Plan

- [x] Run cold-then-warm BAGEL image-edit rollouts on a persistent real-GPU worker.
- [x] Compare pre-change restore behavior against sticky-fp32 with identical seeds.
- [x] Verify output, latent, and SDE log-prob parity.
- [x] Verify encode boundary dtype with and without autocast.
- [x] Verify activation-checkpoint backward and terminal VAE dtype.
- [x] Measure isolated cast cost, steady memory, and peak memory.


heguangxin (approved):
Non-blocking documentation note: `vendor/inferencer.py`, `vendor/__init__.py`, and `vendor/VENDOR_COMMIT.txt` still describe the reference inferencer's temporary fp32 decode followed by dtype restoration as mirroring `BagelVAEDecodeStage`.

It would be good to update that wording in this PR or a follow-up. This does not block the approval.

PR #183, opened 2026-07-06
feat(rollout): WAN 2.1 T2V sglang_diffusion rollout (full+LoRA, curve-aligned)

leviking98z-rgb:
## Summary

Adds **WAN 2.1 T2V** to the `sglang_diffusion` rollout engine (full-FT + LoRA), curve-aligned with the trainside sampler.

- `Wan21T2VAdapter` — rides the `VideoAdapter` base (single UMT5 text encoder; no CFG negative branch when `guidance_scale <= 1`); the sglang server resolves the WAN pipeline from `model_path`.
- WAN `param_names_mapping` + **sglang-leaf-name LoRA targets** (`to_q/to_k/to_v/to_out`) so all 240 attention projections wrap (vs 90 with diffusers paths) — otherwise sglang ran ~base weights and the curve went flat.
- Configs: `wan21_t2v_sglang{,_loramerge,_full,_full_nccl}.yaml`.

## Related Issue

Engine support matrix — #127. Depends on the foundation #135.

## Test Plan

- `py_compile` + import smoke (registers as `wan21`; video module imports) — passes.
- GRPO rollout<->replay reward-curve alignment vs trainside (`old_logp_source=replay`, cross-engine safe): tracks the trainside curve with minimal deviation.

## Compatibility / Risk

Additive — new adapter + configs; depends only on the foundation. No change to existing engines.

## Reviewer Notes

Independent of the other model PRs (#137–#139). Diff vs `main` includes the foundation commit (#135) until that merges.

## Checklist

## EXP
WAN21 TRAINSIDE LORA
截屏2026-07-01 11 54 22

WAN21 SGLANG LORA
截屏2026-07-01 11 54 44

WAN21 SGLANG FULL
截屏2026-07-01 11 56 16

WAN21 TRAINSIDE FULL
截屏2026-07-01 11 55 30

### DIFF
--- a/examples/diffusion/wan21/wan21_t2v_full.yaml
+++ b/examples/diffusion/wan21/wan21_t2v_full.yaml
@@ -1,0 +1,136 @@
+# @package _global_
+# GRPO WAN 2.1 T2V trainside, FULL fine-tuning (no LoRA). Curve-alignment twin
+# of wan21_t2v_sglang_full.yaml — IDENTICAL config except the rollout engine
+# (trainside FSDP sampler here vs SGLang there). Trains the whole transformer
+# (no lora_cfg); trainside uses the live model so there is no weight sync.
+
+num_devices: 16
+batch_size: 24
+adv_use_global_std: true
+num_rollouts: 60
+
+model_config:
+  _target_: unirl.models.wan21.config.WAN21PipelineConfig
+  pretrained_model_ckpt_path: ${oc.env:PRETRAINED_MODEL,Wan-AI/Wan2.1-T2V-1.3B-Diffusers}
+  model_precision: bf16
+  autocast_precision: bf16
+  trajectory_precision: fp16
+  logprob_precision: fp32
+  shift: 5.0
+  max_sequence_length: 512
+  use_lora: false
+
+strategy:
+  _target_: unirl.sde.kernels.FlowSDEStrategy
+
+bundle:
+  _target_: unirl.models.wan21.bundle.WAN21Bundle.from_config
+  config: ${model_config}
+
+pipeline:
+  _target_: unirl.models.wan21.pipeline.WAN21Pipeline
+  shift: 5.0
+  autocast_precision: bf16
+  trajectory_precision: fp16
+  logprob_precision: fp32
+  max_sequence_length: 512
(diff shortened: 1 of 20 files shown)


examples/diffusion/wan21/wan21_t2v_full.yaml
```diff
@@ -0,0 +1,136 @@
...
+
+num_devices: 16
+batch_size: 24
+adv_use_global_std: true
+num_rollouts: 60
+
+model_config:
+  _target_: unirl.models.wan21.config.WAN21PipelineConfig
+  pretrained_model_ckpt_path: ${oc.env:PRETRAINED_MODEL,Wan-AI/Wan2.1-T2V-1.3B-Diffusers}
+  model_precision: bf16
```
zzhuoxin1508, line 15: These three knobs (autocast_precision / trajectory_precision / logprob_precision) are duplicated across model_config, pipeline, and sampling, but only the pipeline: copy is actually consumed at runtime. The model_config/sampling copies don't seem to have a consumer here. You might want to confirm this — if they're redundant, dropping them would avoid the impression that tweaking them has an effect.

zzhuoxin1508 (commented):
These three knobs (autocast_precision / trajectory_precision / logprob_precision) are duplicated across model_config, pipeline, and sampling, but only the pipeline: copy is actually consumed at runtime. The model_config/sampling copies don't seem to have a consumer here. You might want to confirm this — if they're redundant, dropping them would avoid the impression that tweaking them has an effect.

examples/diffusion/wan21/wan21_t2v_full.yaml
leviking98z-rgb: Confirmed — only the `pipeline:` copy is consumed at runtime. The pipeline is instantiated via its constructor (so `model_config`'s `from_config`-path copy is never read), and no runtime path reads the `sampling` copy under `init_same_noise=false` (the sole reader, sglang engine's `init_same_noise` branch, is gated off; the class default `bf16` matches anyway). Dropped both redundant copies across all five wan21 configs in 6c2ed67, keeping only `pipeline:`.

zzhuoxin1508 (approved):
LGTM

PR #186, opened 2026-07-07
feat(hi3): trainside it2i (image-edit) GRPO

### LINKED ISSUE #150: Support HunyuanImage3 it2i RL

heguangxin:

Final review found and fixed one routing omission: `hi3_it2i.yaml` declared `stage_config.task=it2i`, but `UnifiedModelTrainer` did not carry that metadata into `RolloutReq`, allowing the trainside path to fall back to t2i. The current head now wires the recipe value into the trainer and preserves it across root, sharded, AR, DiT, and reward requests.

This supersedes the older statements below that the trainer was untouched.

Current-head validation:

- The contributor's 10 runtime regression cases passed before the PR-added test file was removed; the final PR diff contains no test files.
- `unified_model/hi3_it2i` Hydra compose: passed.
- Targeted pre-commit and `git diff --check`: passed.
- Clean synthetic merge with #255; the compatibility alias there preserves this PR's standalone `stage_config` API while mapping it to canonical `task_config`.

The stale `CHANGES_REQUESTED` review state predates the already-resolved review threads and this final routing fix; it still needs a reviewer to clear it.

## Summary

Single-engine **trainside** GRPO for HunyuanImage-3 image editing (it2i): text + source
image → one `gen_image` DiT forward; the FSDP-wrapped train model IS the sampler (no
separate rollout engine, no weight sync), scored by a remote EditReward service. Changes
are concentrated in HI3; the only shared-code change threads an optional RNG generator through
the SDE strategy while preserving the historical `generator=None` behavior. The trainer and
algorithm core are untouched.

- **rope_cache → first-class stacked `[B,2,L,D]` tensor** (was a `(cos,sin)` tuple that
  could not survive DP transport): now rides CONCAT transport/merge/scatter like
  `input_ids`, and the lazy-`TensorRef` path never has to recurse into a Python tuple.
  `predict_noise` seeds the model's `CachedRoPE` from it; consumers unbind to `(cos,sin)`
  at the model boundary. The hi3 `concat` override materializes lazy refs and pads the L
  dim locally, so `distributed/tensor/batch.py` is left untouched. On the two-engine
  vllm-omni path the engine-layout rope is deliberately NOT captured; replay rebuilds
  rope natively from `gen_image_mask` (re-verified on GPU — see the 2026-07-14 comments).
- **Cond-image timestep embedding injected** (review fix): `predict_noise` now passes
  `cond_timesteps_index` (`fused.cond_timestep_scatter_index`) like native
  (modeling:2842), so `instantiate_continuous_tokens` runs for the cond `<timestep>`
  slot (t≈0 = clean source). Previously it was pinned to `None`, which silently skipped
  the injection (`_check_inputs` is bypassed) — consistent between sampling and replay,
  so invisible to the ratio probes, but off-native conditioning and a train-vs-deploy
  input gap.
- **Per-rank VAE decode**: mask `torch.distributed.is_initialized()` around the 3D-VAE
  `decode` so every DP rank decodes its own samples. The stock rank-0-only tile-gather is
  correct for one big image but wrong for per-rank RL rollout (otherwise the merged
  track's `decoded` has 1/dp_size the rows and the reward scatter slices out of bounds).
- **it2i device placement**: move the source-image cond payloads (VAE latents / ViT
  features / cond timestep) to `sample.device` for the gen_image forward, avoiding a
  cpu-vs-cuda mismatch in the token-instantiate Linears.
- **`diffuse_kv_cache` config knob** (replaces the earlier `HI3_DIFFUSE_KV_CACHE` env
  gate): `false` makes sampling full-prefill (bit-matching replay) for the exact native
  on-policy ratio under `old_logp_source=rollout`; the recipe sets it explicitly, the
  default `true` keeps the KV-cached fast sampling.
- **Driver-authored deterministic x_T** (review fix): `HunyuanImage3Pipeline.latent_shape`
  opens `DiffusionTrainer`'s x_T-recipe gate, so `init_noise_group_ids` are authored and
  the it2i bridge attaches them — x_T is now deterministic per
  `(sampling.seed, rollout_id, sample_id)`: reproducible runs, cross-engine-aligned
  curves, `init_same_noise=true` usable trainside. NB: the earlier "seed wiring" commit
  (9f7a8f46) alone was inert — hi3 shipped no `latent_shape`, so the ids were always
  empty and x_T stayed unseeded `torch.randn`; pre-fix runs drew x_T from engine RNG
  regardless of `seed` (their conclusions are unaffected — replay never consumes x_T).
- **Recipe** `examples/unified_model/hi3_it2i.yaml`: bit-exact `guidance_scale=1.0`
  default (train unguided, eval/deploy guided — the standard train-cfg=1 paradigm).
- **Optional guided training (`guidance_scale>1`)**: the rollout samples the CFG-combined
  velocity from a `[cond; uncond]` (N=2B) batch; a naive replay recomputes unguided and
  the ratio drifts (~0.998). To stay on-policy, `modes/it2i.py` stores the fused cond-only
  plus its uncond branch (`fused_uncond`, both B-batched so they survive the B-sample
  track transport) and the stage re-stacks `[cond; uncond]` for BOTH sampling and replay,
  so `predict_noise` runs cfg=True on both sides. This recovers the ratio to ~0.9999
  (clip_frac=0); a ~1e-4 residual remains — the N=2 guided-forward's own bf16
  non-reproducibility amplified ~guidance_scale× by the cfg combine (non-clipping).
  cfg=1 stays bit-exact.

## Results

image

image2

- **On-policy importance ratio**: cfg=1 → **1.0000 ± 0.0000** (bit-exact);
  cfg=2.5 → **0.9999 ± 0.0001**, clip_frac = 0.
- **Reward** (EditReward logit; GRPO uses the per-group *relative* advantage, so judge the
  trend): the validated setting (native, B=8 / G=2, CFG=2.5, LoRA r=8/α=16, 3 SDE steps)
  rises from ~−0.88 to ~−0.29 over ~196 rollouts and holds — full curves in the
  2026-07-09 and 2026-07-16 comments. The r=64/α=128 divergence reported in-thread is a
  tuning effect (ratio ≈ 1, clip_frac = 0 throughout).

## Test Plan

- [x] CPU Hydra compose: `python -m unirl.train_diffusion --config-name unified_model/hi3_it2i --cfg job --resolve` resolves.
- [x] Multi-node (2×8) smoke: ratio 1.0000 (cfg=1) / 0.9999 (cfg=2.5), clip_frac=0, no crashes; VAE decode + reward path clean.
- [x] Short training runs: on-policy ratio holds; reward trends upward.
- [x] Longer reward curves to a plateau — attached in the 2026-07-09 / 2026-07-16 comments.
- [x] vllm-omni two-engine regression after the rope migration: stock `hi3_vllmomni` smoke full pass, image ratio=1.0000±0.0000 — see the 2026-07-14 verification comment.
- [x] Review-fix CPU checks: `latent_shape` geometry + trainer-gate resolution (hydra `get_object`); `NoiseRecipe` byte-determinism / seed-sensitivity / per-sample distinctness; ruff, recipe-`_target_` hook, Hydra compose.
- [x] Post-review GPU re-smoke after the cond-timestep/RNG fixes (`ab3ca917`): same-seed x2 matched all 16 x_T, VAE-latent, clean-latent, float-image, and uint8-image hashes; cfg=1 ratio remained 1.0000. The vllm-omni rollout/reward/replay/optimizer smoke also completed.
- [x] The 10 CPU regressions passed before removal, covering per-request x_T recipe slicing, `[B]`/`[2B]` CFG sigma handling, per-sample VAE RNG, stacked RoPE concat, wrapper mode, and per-rank VAE decode.
- [x] Current-head Hydra compose, full changed-file pre-commit suite, and `git diff --check origin/main...HEAD`.

## Compatibility / Risk

- Model behavior changes are HI3-local (`unirl/models/hunyuan_image3/*`) plus one recipe and HI3 vllm-omni adapters. The shared `unirl/sde/kernels.py` change only forwards an explicitly supplied generator; all existing `generator=None` callers retain their prior behavior. Shared batch/transport primitives and the trainer are unchanged.
- The `rope_cache` field type change is local to the hi3 `FusedMultimodalCondition`
  subclass; the vllm-omni producer was migrated with it and the two-engine path
  re-verified on GPU (2026-07-14 comments).
- The VAE `is_initialized` mask is scoped (try/finally) to the frozen decode forward only.
- x_T now varies with `sampling.seed` (was: unseeded engine RNG). Training behavior is
  statistically unchanged; set `seed: null` to keep engine-RNG x_T.
- Known follow-up (not in this PR): `samples_per_prompt>2` trips a `cond_vit` list-merge —
  now fail-fast in `modes/it2i.py` and pinned in the recipe. Making those per-sample
  tensors would unlock larger GRPO groups and remove the SHARED `spatial_shapes`
  replica-0 caveat noted in the recipe (source images must share one resolution).

## Checklist

- [x] Longer reward evidence attached (2026-07-09 / 2026-07-16 comments).
- [x] Review fixes (`b495c3af`…`ecb3272`) reviewed and covered by targeted regressions.

## Related issue

Closes #150


Vampire689:
Hi @heguangxin, thanks for the detailed PR description! I When do you expect to have the longer reward-improvement curves and final results ready?

heguangxin:
> Hi @heguangxin, thanks for the detailed PR description! I When do you expect to have the longer reward-improvement curves and final results ready?

Hi @Vampire689 — update with longer runs (same code, cfg=2.5, `HI3_DIFFUSE_KV_CACHE=0`;
batch_size=8 × samples_per_prompt=2, num_updates_per_batch=2, clip_range=5e-3, LoRA r=8/α=16):

**Reward** (EditReward logit; climbs from ~−1.0 — see plot below):
- **replay** (`old_logp_source=replay`), seed 42 → steady climb to ~−0.40 (~115 rollouts).
- **native** (`old_logp_source=rollout`), seed 123 → climbs to ~−0.22 and plateaus (256 rollouts).


reward_ab3

**On-policy ratio** (see plot below): mean |1−ratio| ≈ 3–6e-5 in both modes, clip_frac = 0
throughout (clip_range = 5e-3).

ratio_ab3

Native shows a bit more seed-to-seed variance than replay at the current small group size
(samples_per_prompt=2) — the two native-seed curves bracket the replay run. **Next:** I'll run
a larger batch / group-size setting to see whether native training gets more consistently
stable, and post those curves + a final summary.

heguangxin:
Latest update: commit `9f7a8f4` fixes the HI3 trainside seed wiring so that `sampling.seed` is correctly consumed when reconstructing the initial `x_T`. After this fix, the native seed-42 run also shows a sustained improvement in reward.

haonan3:
@heguangxin **Review finding (P0): this branch breaks the shipped vllm-omni hi3 path. Reproduced on GPU with the stock recipe — details + fixes below.**

## What breaks

This PR changes `rope_cache` on the hi3 fused condition from the base's `(cos, sin)` tuple (`shared_field`) to a stacked `[B, 2, L, D]` `CONCAT` tensor — the right call for trainside DP transport. But the vllm-omni producer side was not migrated: the capture still stores a tuple (`pipelines/hi3/pipeline.py:158`) and the adapter still assembles `fused_dict["rope_cache"] = (cat_cos, cat_sin)` (`adapters/hi3.py`). Every framework op that now treats the field as a per-sample CONCAT tensor chokes on that tuple.

Ironically, `unirl/trainer/unified_model.py:437-448` predicted exactly this: *"If a future change makes replay consume `fused.rope_cache` … make rope_cache a tuple-aware CONCAT field before relying on it."* The field was migrated; the producer wasn't.

## Reproduction (real run, unmodified PR head 2b2fae89)

8×H20, stock `examples/unified_model/hi3_vllmomni.yaml`, only `num_rollouts=1 logging.report_to_wandb=false`. AR + DiT rollout complete, PickScore scores fine, then the **first reward DP-merge** dies — the run never reaches training:

```
File "unirl/trainer/unified_model.py", line 637, in train_step
    scored = self.reward.score_and_attach(req=reward_req, track=img_track)
  ...
File "unirl/models/hunyuan_image3/conditions.py", line 165, in concat
File "unirl/distributed/tensor/batch.py", line 303, in _concat_value
ValueError: Cannot concat values: types=['tuple', 'tuple', 'tuple', 'tuple',
    'tuple', 'tuple', 'tuple', 'tuple'], batch_sizes=[1, 1, 1, 1, 1, 1, 1, 1]
```

Two more sites crash on the same root cause (verified in an A/B repro that drives the real `hi3_fused_conditions` adapter output through the real code; both PASS on main `451c746d`, both CRASH on this branch):

- ragged-L merge path: `_pad_seq(tuple)` → `AttributeError: 'tuple' object has no attribute 'shape'` (`conditions.py:182`)
- training replay: `fused.rope_cache[:, 0]` → `TypeError: tuple indices must be integers or slices, not tuple` (`diffusion.py:226`)

## Fixes being pushed to this branch (one commit each)

1. **fix(hi3): migrate the vllm-omni rope_cache producer to the stacked layout** — stack at capture, plain pad/cat in the adapter, plus a fail-fast type check in `from_dict` so a future producer regression errors at the boundary instead of deep inside a merge.
2. **fix(hi3): `predict_noise_at_step` applies `_expand_cfg_for_forward`** — otherwise the split cond/uncond layout silently runs unguided at `guidance_scale>1` for forward-process algorithms, diverging from `diffuse()`/`replay()`.
3. **feat(hi3): promote `HI3_DIFFUSE_KV_CACHE` to `HunyuanImage3PipelineConfig.diffuse_kv_cache`** — with `old_logp_source=rollout` this knob decides on-policy exactness; forgetting the env silently drifts the ratio to ~0.997 with no error. A correctness knob belongs in the composed config; the it2i recipe now sets it explicitly.
4. **docs(hi3): refresh stale comments** — `text_embed.py` rope docstrings (still say tuple), `unified_model.py` deep_hydrate/CAVEAT blocks (describe the pre-migration world), recipe eval note (the `cfg_text_scale` TypeError it cites was fixed by #202) and the leftover `VERIFY at smoke` marker.

I'll post the post-fix GPU verification (same stock vllm-omni smoke, plus the CPU A/B flipping green) as a follow-up comment.

## Still open — on you before merge

- Longer reward curves to plateau + final summary, **re-run on top of 9f7a8f46** (the seed fix invalidates the earlier native-mode curves), as the PR body promises.
- Drop the "Status: DRAFT — evidence updating" header once that lands.
- After pulling these commits, please re-run your trainside it2i validation once (`diffuse_kv_cache: false` in the recipe replaces `export HI3_DIFFUSE_KV_CACHE=0`).

haonan3:
**Post-fix verification: the vllm-omni path is restored — full pass on GPU.** Evidence chain below; all runs are 8×H20, stock `examples/unified_model/hi3_vllmomni.yaml` with `num_rollouts=1 logging.report_to_wandb=false` (later runs add `batch_size=4`, see footnote 1).

| head | result |
|---|---|
| `2b2fae89` (pre-fix) | ❌ dies at the **first reward DP-merge**, never reaches training: `ValueError: Cannot concat values: types=['tuple' ×8]` |
| `89c638ea` (layout fix) | past the merge ✅, then ❌ in replay `apply_rotary_pos_emb`: `The size of tensor a (128) must match the size of tensor b (64)` — the captured **engine** rope uses vllm-omni's own half-dim table convention, incompatible with the HF-side replay forward |
| `a38d02c1` (rope-source fix) | ✅ **full pass, exit 0** — `rollout 1/1 reward=0.8402 ar[ratio=0.9936±0.0646 clip=0.39] image[loss=0.0002 ratio=1.0000±0.0000 clip=0.00]` |

The image-track step-0 probe at **`ratio=1.0000±0.0000, clip=0.00`** confirms the replay is numerically correct, not merely non-crashing.

## What is on the branch now (5 commits)

- `b80c3633` fix: rope_cache rides as a stacked `[B,2,L,D]` tensor end-to-end; `from_dict` fails fast on legacy tuples
- `d43a8543` fix: `predict_noise_at_step` applies the same CFG expansion as `diffuse()`/`replay()`
- `63e62c9a` feat: `HI3_DIFFUSE_KV_CACHE` env → `HunyuanImage3PipelineConfig.diffuse_kv_cache` (the it2i recipe sets `false` explicitly; env deleted)
- `89c638ea` docs: refresh comments the rope/eval changes had left stale
- `a38d02c1` fix: the two-engine path ships **no** engine rope; replay rebuilds `rope_image_info` natively from `gen_image_mask` (the pre-branch reconstruction this path was validated with). **Trainside keeps your exact-seed mechanism untouched** — `rope_cache` is always present there (built by `text_embed`, same modeling code both sides), so the it2i ratio=1 story is unaffected.

Design note now baked into the code comments: captured-rope seeding is only valid when rollout and replay run the *same modeling code* (trainside). Engine-internal representations must not cross into the HF-side replay.

## Footnotes / still open

1. *(pre-existing, not touched here)* stock `hi3_vllmomni.yaml` defaults `batch_size=2` × `ar.samples_per_prompt=2` = 4 AR samples, which is **not divisible by the 8-way train DP** — training dispatch dies with `ValueError: batch_size=4 not divisible by dp_size=8`. Verification runs used `batch_size=4`. The recipe default deserves a follow-up fix.
2. Still on you before merge: longer reward curves to plateau + final summary **re-run on top of 9f7a8f46**, drop the "Status: DRAFT" header, and please re-run your trainside it2i validation once on this head (`diffuse_kv_cache: false` in the recipe replaces the old env export; trainside behavior is otherwise unchanged).

Full logs archived for the team on the shared ceph: `gpu_registry/runs/20260714/pr186_p0_verify/`.

Vampire689:
Based on `hi3_it2i.yaml`, I make the following change (cfg 2.5, LoRA r=64/α=128, timestep_fraction: [0.3, 0.7]), but the reward continuously declining. Which configs require updates?
image

heguangxin:
@Vampire689 

Re-ran on a38d02c1 (HI3 it2i code unchanged through the current head ee4ab82).

**Validated setting** — native, B=8 / G=2, CFG=2.5, LoRA r=8 / α=16, seed=42, 3 SDE steps `timestep_fraction=[0, 0.5]`, `diffuse_kv_cache=false` (all in the YAML). Over ~196 rollouts the EditReward logit rises from ～−0.879 (first-20 mean) to ～−0.29 and holds.

figA_validated_setting

**On the r=64 / α=128 config** — I ran it in both the original `[0, 0.5]` and your `[0.3, 0.7]` SDE window. Neither sustains an upward trend: `[0, 0.5]` climbs early then declines, `[0.3, 0.7]` stays roughly flat then drifts down. The on-policy ratio stays ≈ 1 with clip_frac = 0 in every run, so the decline is more likely a hyperparameter-choice / tuning effect than an implementation issue. The setting this PR validates is r=8 / α=16.

The numbers here are the raw EditReward logit (higher = better); the earlier plot uses a different scale and axis orientation.

If there is a further issue, feel free to open a separate issue.

haonan3:
**Review pass 2 — two correctness fixes + guards pushed to this branch (`b495c3af`…`f813da31`), and one record correction.**

## What was pushed

**`b495c3af` — fix: inject the cond-image timestep embedding.** `predict_noise` pinned `cond_timesteps_index` to `None`, so upstream `instantiate_continuous_tokens` never ran for the cond `<timestep>` slot (modeling:2201 gates on the index) and the slot kept the plain vocab embedding instead of `timestep_emb(t≈0)` ("this source image is clean"). It was silent because our hand-built `model_inputs` bypasses `_check_inputs` (which demands the index alongside `cond_vae_images`, modeling:2142-2145); native passes `output.cond_timestep_scatter_index` (modeling:2842), and our own AR path already did too (`ar.py:231-232`) — only the DiT kernel dropped it. Sampling and replay were consistently wrong *together*, so every ratio/clip probe stayed green; the real costs were off-native it2i conditioning and a train-vs-deploy input gap (deploy runs native, WITH the injection). Expect ratio unchanged after the fix (both sides change identically) and conditioning fidelity same-or-better. Credit: surfaced by a Codex review pass.

**`3bc52795` — feat: `latent_shape` opens the x_T recipe gate (makes the seed wiring real).** Record correction first: the 07-13 note "`9f7a8f4` fixes the HI3 trainside seed wiring" was **inert as shipped** — `DiffusionTrainer` only authors `init_noise_group_ids` when the pipeline implements `LatentShapeProvider.latent_shape` (`trainer/diffusion.py:299-333`, `:378-379`), which hi3 didn't. So `recipe_ids` was always `[]`, the it2i bridge never fired, and x_T stayed unseeded `torch.randn` (`sde/noise.py:147-152`) — `sampling.seed` had no effect on x_T. That also means my 07-14 remark "the seed fix invalidates the earlier native-mode curves" was wrong: pre-fix runs (seed 42 and 123 alike) drew x_T from the same engine RNG, and nothing was invalidated. Training conclusions are unaffected throughout — replay starts from stored latents and never consumes x_T. @Vampire689 heads-up: don't read seed-sensitivity into pre-fix runs.

With `latent_shape` implemented (32ch / 16x, the stage defaults; non-16-multiple sizes opt out via `NotImplementedError`), the gate opens and the existing bridge + `diffuse()` recipe path go live: x_T is deterministic per `(seed, rollout_id, sample_id)` — reproducible runs, cross-engine-aligned curves, and `init_same_noise=true` becomes usable trainside. HI3 snaps non-preset sizes inside `diffuse()` and regenerates x_T at the snapped grid via `NoiseRecipe.for_batch`, so the returned shape only gates/labels the recipe.

**`f813da31` — fail-fast guards + stale comments.**
- Rope-cache seeding now raises if the HF **wrapper** is in eval mode: eval computes `seqlen=max_position_embeddings` (modeling:2155-2159), which silently MISSES the seeded `CachedRoPE` and rebuilds a wrong 1-D rope. It works today only because `TrainsideRolloutEngine` eval()s just `trainable_module()` (the inner decoder) — now a loud failure instead of silent degradation if that ever changes.
- it2i rejects `samples_per_prompt>2` up front (was a bare `Cannot concat values` deep in the dp>1 merge).
- VAE decode comment corrected: `AutoencoderKLConv3D.decode` returns `DecoderOutput` on the normal path (autoencoder_kl_3d.py:825), a bare tensor only on the rank!=0 discard / `_Dist` paths.
- Recipe notes the source-image homogeneity requirement: `cond_vit.spatial_shapes` is a SHARED field, so a dp>1 merge keeps replica-0's per-sample list — heterogeneous ViT grids would silently cross-feed at replay. (Today's edit sets are homogeneous, which is why ratio=1.0000 holds; the per-sample-tensors follow-up removes the caveat and unlocks larger groups.)

## Verified (CPU) / on you before merge (GPU)

CPU, all green: `latent_shape` geometry + the trainer gate resolving via hydra `get_object`; `NoiseRecipe` byte-determinism / seed-sensitivity / per-sample distinctness; ruff, the recipe-`_target_` hook (1843 paths), and the Hydra compose check.

@heguangxin please re-run the it2i smoke once on this head: expect **ratio still 1.0000** at cfg=1 (the cond-timestep injection changes sampling and replay identically) and reward same-or-better; plus a same-seed ×2 rollout-0 x_T/image-hash equality check to confirm recipe determinism end-to-end. PR body refreshed accordingly (env-gate bullet → `diffuse_kv_cache`, draft markers dropped, the new fixes and the remaining GPU item listed in Test Plan).

Follow-up to file separately: unify the x_T recipe conventions — de-dup the two trainer-side producers, one consumer helper across sd3/bagel/hi3 variants, and decouple id-authoring from the `latent_shape` gate (at which point hi3's `latent_shape` can be dropped again).

*(Review + fixes assisted by Claude Code; human-reviewed before push.)*

heguangxin:
**GPU follow-up for review pass 2: strict same-seed reproducibility now passes end to end (`ab3ca917`).**

On `f813da31`, the driver-authored initial `x_T` was already deterministic, but same-seed runs still diverged afterward because the source-image VAE posterior and per-step SDE transition noise consumed engine-local RNG state.

`ab3ca917` derives independent request-local generators from `sampling.seed` and stable request/sample/step identifiers, then threads them through those two HI3 it2i paths. It does not reset global RNG state.

The change is intentionally opt-in. UniRL's existing trainside contract remains unchanged: `NoiseRecipe` makes the initial `x_T` reproducible, while per-step SDE noise uses engine-local RNG and replay evaluates the stored transition rather than re-drawing it. Callers that leave `generator=None` retain this behavior; HI3 it2i explicitly supplies request-local generators for strict end-to-end reproducibility.

## GPU verification

- **Trainside same-seed x2:** two independent one-rollout runs with the stock cfg=1 setting (`B=8`, `G=2`, LoRA r=8/alpha=16, seed 42, three SDE steps in `[0, 0.5]`, `diffuse_kv_cache=false`) produced identical request IDs and matching 16/16 hashes for:
  - initial `x_T`;
  - source-image VAE latents;
  - clean latents;
  - final float and uint8 images.
- Both runs logged `ratio_mean=1.0000`, `ratio_std=0.0000`, `clip_fraction=0.00`, and reward `-1.6162`. The reward remains within the pre-fix same-setting A/B range (`-1.6229` to `-1.5928`), so no one-rollout reward regression was observed.
- **vLLM-Omni regression check:** a one-rollout GPU smoke (`batch_size=4` for 8-way DP) completed AR + DiT rollout, 16/16 image generation, reward DP merge, train-side replay, and the optimizer step with exit code 0. The image stage retained the FlowGRPO default `old_logp_source=rollout` and logged `ratio_mean=1.0000`, `ratio_std=0.0000` at the logger's four-decimal precision.
- **Compatibility:** the pre-fix and patched shared `generator=None` paths produced identical output and log-probability hashes. Full pre-commit and all PR checks pass.

This addresses the remaining GPU item from review pass 2 without changing the default SDE RNG semantics for callers that do not opt into explicit generators.

haonan3 (approved):
LGTM

unirl/models/hunyuan_image3/pipeline.py
```diff
@@ -119,6 +119,29 @@ def from_config(
...
+        the recipe; it is NOT the materialization shape for HI3.
+        """
+        height = int(sampling_spec.height)
+        width = int(sampling_spec.width)
+        if height <= 0 or width <= 0 or height % 16 or width % 16:
+            raise NotImplementedError(
+                f"HunyuanImage3Pipeline.latent_shape: {height}x{width} is not a multiple of the "
+                "16x VAE factor; opting out of the driver x_T recipe (engine RNG fallback)."
+            )
+        return (32, height // 16, width // 16)
```
CjhHa1, line 142: [P1] Slice HI3 x_T recipes per vLLM request

Adding `latent_shape()` activates driver-authored `init_noise_group_ids` for every HI3 `DiffusionTrainer` request, not only trainside it2i. `Hi3InputAdapter` currently puts the entire request's ID list into one diffusion sampling object shared by all prompts. vLLM-Omni runs each prompt as a `batch_size=1` DiT request, while `RLHunyuanImage3Pipeline` copies the full list and rejects `len(gids) != batch_size`; therefore registered `hi3_t2i`/`hi3_it2i` requests with more than one sample now fail in `prepare_latents`. The `dit_recaption` adapter already avoids this by sending one gid per prompt. Please apply equivalent request-ID slicing here, or scope the recipe gate to paths that consume it correctly.

unirl/models/hunyuan_image3/diffusion.py
```diff
@@ -179,42 +213,80 @@ def predict_noise(
...
+        # is a CONCAT field, so under DP_SCATTER each rank already holds ONLY its
+        # own rows — no replica-0 cross-feed.
+        if fused.rope_cache is not None:
+            # Hit-precondition: the HF wrapper computes seqlen as
+            # ``input_ids.size(1)`` ONLY in train mode (modeling:2155-2159; eval
+            # uses max_position_embeddings) — in eval the seed silently MISSES
+            # and build_2d_rope rebuilds a wrong 1-D rope. TrainsideRolloutEngine
+            # eval()s only ``trainable_module()`` (the inner decoder), never this
+            # wrapper; fail loudly if that ever changes.
+            if not transformer.training:
```
CjhHa1: [P1] Keep eager bundles compatible with this RoPE guard

`AutoModelForCausalLM.from_pretrained()`, used by `HunyuanImage3Bundle.from_config`, returns the outer wrapper in eval mode. The FSDP backend and `TrainsideRolloutEngine` toggle only `bundle.trainable_module()` (the inner decoder), so `transformer.training` remains false and every eager trainside forward carrying `rope_cache` reaches this `RuntimeError`. The meta-init recipe does not expose this because `from_config()` is not used there. Please make cache seeding independent of the wrapper's mode, or explicitly establish the required outer-wrapper mode without recursively putting the frozen VAE/ViT into training mode.

unirl/models/hunyuan_image3/diffusion.py
```diff
@@ -838,6 +1006,9 @@ def predict_noise_at_step(
         cache is only meaningful inside an SDE trajectory, which DiffusionNFT-style
         forward-process algorithms don't traverse.
         """
+        # Re-stack the guided [cond; uncond] batch exactly as diffuse()/replay()
+        # do, so guidance behavior matches them (no-op when fused_uncond is None).
+        conditions = _expand_cfg_for_forward(conditions)
```
CjhHa1, line 1017: [P2] Duplicate batched sigma when rebuilding CFG

Batched `DiffusionNFT`/SFT callers pass `sigma` as `[B]`. After this line expands the conditions to `2B`, `predict_noise()` duplicates `sample` to `2B` but executes `t_scalar.expand(2B)`. A non-singleton `[B]` tensor cannot be expanded to `[2B]`, so guided batches with `B > 1` fail before the model forward. Please repeat sigma blockwise (for example, `torch.cat([sigma, sigma])`) or make `predict_noise()` explicitly accept and validate both `[B]` and `[2B]`.

unirl/models/hunyuan_image3/modes/it2i.py
```diff
@@ -56,39 +91,40 @@ def generate(pipeline: "HunyuanImage3Pipeline", req: RolloutReq) -> RolloutResp:
...
+    #    cfg uncond branch keeps the SAME source image, so cfg doubling of these
+    #    payloads is a pure block duplication (_encode_cond_image / vit_kwargs do
+    #    ``.repeat`` / ``list*cfg``); the diffusion stage re-applies it when it
+    #    expands CFG. Keeping them B-batched (not doubled) means they survive the
+    #    B-sample track transport that a 2B batch would not.
     cond_vae_images, cond_timestep, cond_vit_images = pipeline.bundle.transformer._encode_cond_image(
-        vit["joint_image_info"], cfg_factor=cfg_factor
+        vit["joint_image_info"],
+        cfg_factor=1,
+        generator=condition_vae_generators,
```
CjhHa1: [P2] Select the matching VAE generator per sample

HunyuanImage3's `_encode_cond_image()` loops over samples and calls `vae_encode()` on a one-row tensor, forwarding this entire generator list each time. Diffusers' `randn_tensor()` uses only `generator[0]` when `shape[0] == 1`, so generators `1..B-1` are never consumed. `forward_batch_size=1` hides this in the checked-in recipe, but any batched pipeline call makes the condition-VAE posterior noise depend on batch order and DP/chunking. Please pass the matching generator for each outer sample (or batch the VAE encode) rather than forwarding the full list into every single-image call.

examples/unified_model/hi3_it2i.yaml
```diff
@@ -0,0 +1,229 @@
...
+#   flow_grpo #198. Use guidance_scale>1 only if guided-sample training is worth the
+#   small non-exact (but non-clipping) ratio.
+#
+# Deploy (EditReward OFF the training GPUs — see PODS_AND_ENV 8+1 split):
+#   reward node: standalone MiMo EditReward HTTP server on :8080.
+#   train node:  export REWARD_SERVICE_URL=http://<reward_node_ip>:8080
+#                export PRETRAINED_MODEL=... EDIT_DATA_PATH=... EDIT_EVAL_DATA_PATH=...
+#                PYTHONPATH=$PWD python -m unirl.train_diffusion --config-name unified_model/hi3_it2i
+#
+# Smoke: override batch_size=1 num_rollouts=2 logging.report_to_wandb=false.
```
CjhHa1: [P3] Keep the documented smoke override divisible by DP

With `num_devices=8` and `samples_per_prompt=2`, `batch_size=1` creates only two rollout samples. `DiffusionTrainer` preflight / `DP_SCATTER` rejects that because it is not divisible by `dp_size=8`. Please use `batch_size=4` for the documented 8-GPU smoke, or include a matching `num_devices=2` override.

CjhHa1 (commented):
I found several runtime correctness issues. Details are attached to the relevant lines.

CjhHa1 (changes requested):
Need change

heguangxin:
> I found several runtime correctness issues. Details are attached to the relevant lines.



> Need change

Thanks for the detailed review. I am working on them.

unirl/models/hunyuan_image3/pipeline.py
leviking98z-rgb: Addressed in add05db: multi-prompt HI3 AR+DiT requests are split into single-prompt GenerateCalls and each DiT sampling object receives only its matching x_T recipe gid. Current-head regression test_hi3_input_adapter_slices_noise_recipe_per_worker_request also checks ordering, seed propagation, and mismatch rejection.

unirl/models/hunyuan_image3/diffusion.py
leviking98z-rgb: Addressed in add05db: rope-cache seeding sets only the outer wrapper training flag, without recursively changing the frozen VAE or ViT mode. Current-head guided-forward regression verifies the wrapper flips while both frozen children stay in eval mode.

unirl/models/hunyuan_image3/diffusion.py
leviking98z-rgb: Addressed in add05db and hardened in ecb3272: guided [B] sigma is duplicated blockwise to [cond B; uncond B], [2B] is accepted directly, scalar and legacy [1] forms broadcast, and misaligned shapes fail fast. The current-head regression covers all four cases.

unirl/models/hunyuan_image3/modes/it2i.py
leviking98z-rgb: Addressed in add05db: source-image VAE encoding now invokes _encode_cond_image once per outer sample with that sample own one-element generator list. The current-head regression checks generator identity and reconstructed VAE/timestep/ViT ordering.

examples/unified_model/hi3_it2i.yaml
leviking98z-rgb: Addressed in add05db: the documented 8-GPU smoke override now uses batch_size=4, so 4 prompts x 2 samples is divisible by DP=8. The current-head Hydra compose check also passes.

leviking98z-rgb (commented):
Reviewed the current head after the batched-runtime fixes. The per-request HI3 x_T recipe slicing and CFG [B]/[2B] sigma paths are correct and now covered by targeted regressions; I also verified per-sample VAE generators, stacked RoPE concat, wrapper-mode handling, and per-rank VAE decode against the upstream HI3 contracts. Ten targeted tests, Hydra compose, all changed-file pre-commit hooks, diff validation, lint, PR metadata checks, and CodeCC pass. All five review threads have evidence-backed replies and are resolved. Code is ready for the original changes-requesting reviewer to clear the stale review state.

PR #187, opened 2026-07-07
feat(metrics): add rollout replay K3 metric

cboss6:
## Summary
Add per-token K3 KL estimator between rollout and replay log-probs.

## Test Plan
Not run; reason: not necessary

## Checklist


leviking98z-rgb (approved):

zzhuoxin1508:
rollout_replay_k3 is defined and exported here but not called anywhere in the repo yet. That's fine if the call site is coming later, but as-is a reader has no way to know how it's meant to be used. Could you document the intended usage — either in the docstring，or in the PR description 。

cboss6:
@zzhuoxin1508 Thanks for flagging this. You're right that rollout_replay_k3 had no call site in this PR's original scope — this PR is really about defining and exporting the metric helper. The actual call site lives in a separate, upcoming PR of mine. To address your point and keep this PR self-contained rather than shipping a dead export, I've temporarily cherry-picked that call site over here.

zzhuoxin1508:
LGTM. Next time please write a more detailed PR description

PR #189, opened 2026-07-07
fix(trainer/ar): evaluate over the full validation set, not one batch

cboss6:
## Summary
1) Add iter_eval_batches(batch_size) that yields the full dataset deterministically; get_eval_samples now returns next(iter, empty) so callers that only want one batch still work.  
    Old evaluate() pulled the first eval_num_prompts prompts from val.jsonl and reported their avg@k as "eval accuracy". Rewrite to iterate the whole eval_dataset in batches of eval_num_prompts(now: eval *batch size*, not eval size), accumulate reward_sum / reward_n across batches, and emit ONE aggregate scalar per eval — wandb still gets one eval/acc point per rollout, mathematically identical to a single mean over the concatenated batch.

2) Also move rollout.sleep() into a try/finally so a mid-eval crash still the rollout engine back to sleep (previously an OOM during eval left the workers awake).
 

## Test Plan
No test, reason: no need

## Checklist


NancyFyong (commented):
@cboss6 Thanks for fixing both issues — the core approach and the `try/finally` handling look good to me. A few things I'd like to discuss before merge:

1. **On the `eval_num_prompts` semantics change.** As I understand it, this config used to mean "how many prompts to evaluate," and after this PR it becomes "the batch size for iterating the full eval set." If someone has `166` configured against a 1000-row `val.jsonl`, each eval now takes ~6× longer with no change in the config name — easy to miss. Would you be open to introducing a separate `eval_batch_size`, or keeping `eval_num_prompts` as a cap? If the change is intentional, noting it in the PR body works too. Totally open to whatever you think fits best.

2. **`iter_eval_batches(0)` seems to sweep the whole dataset.** I tried it quickly: `max(1, int(0))` turns `0` into `1`, which then iterates the full val set prompt-by-prompt, while `get_eval_samples(0)` returns an empty batch — the two disagree on the `0` case. Would you mind considering a short `if batch_size <= 0: return` guard at the top? If the behavior is intentional, please let me know — I may be missing something.

The pooled mean and the `DefaultDataSource` return-type alignment both look good to me. These are just suggestions — happy to discuss, and please push back if any of them don't make sense to you.

cboss6:
@NancyFyong
thanks for your response! I pushed a follow-up commit:

1. eval_num_prompts semantics

Agreed that silently repurposing this field would be confusing. I looked at how verl handles the related behavior: verl has data.val_batch_size, where val_batch_size=None means evaluating the full validation set in one batch.

I updated the PR to split the two concepts explicitly:
```python
eval_num_prompts: int = -1 
# Cap on prompts scored per eval: 
#   -1 = full eval set
#    0 = evaluate nothing / explicit skip
#    N = evaluate the first min(N, len(dataset)) prompts  

eval_batch_size: int = 8 
# Batch size used when iterating over the selected eval prompts 
```

I used -1 as the “full eval set” sentinel instead of 0, so that 0 can keep its literal meaning: evaluate zero prompts. Otherwise, there would be no clean way to express that intent via config.

Does this look reasonable to you? I’m also happy to discuss further if you think there’s a clearer config design.

NancyFyong:
LGTM

NancyFyong (approved):

leviking98z-rgb (approved):

PR #193, opened 2026-07-09
feat(qwen3-omni): add qwen3-omni thinker AR with train-side colocate mode as rollout backend (text + video).

cboss6:
## Summary

This PR adds a **train-side colocate** rollout path for Qwen3-Omni Thinker RL, as a dependency-free counterpart to the vLLM-Omni backend. In this mode the training pipeline *is* the sampler: `TrainsideRolloutEngine`
autoregresses with the same FSDP-wrapped `Qwen3OmniMoeThinkerForConditionalGeneration` that is later optimized. There are no rollout actors, no engine process, and no LoRA weight synchronization — the sampler and the learner are the same module, so rollout and replay are consistent by construction rather than by a sync contract.
That makes it useful for three things:
* bringing up Qwen3-Omni RL in environments where vLLM / vLLM-Omni is not installed or not yet ported to the target stack;
* debugging the multimodal replay contract in isolation — video embedding fusion, DeepStack features, and TMRoPE position rebuilding are exercised without an inference engine in the loop;
* serving as a reference oracle for rollout↔replay alignment, since `k3` and `|Δlogp|` here measure only replay numerics, with no engine-side divergence mixed in.
It trains the same parameter set, over the same prompts, with the same video preprocessing as the merged vLLM-Omni recipes.

## Related PR
* **#235**
This is the train-side PR that `#235` referred to as *"supports trainside as the rollout backend, and needs to be refactored after this PR merged."* That refactor is done here: the branch is rebased onto `main` at/after `#235`'s merge commit (`e1a814f`), and **all shared `unirl/models/qwen3_omni/*` files are taken from `main`**. This PR therefore adds a recipe and a dataset script, and touches exactly one shared source file (see *Transformers compatibility* below). No vLLM-Omni recipe or stage config is modified.

## Recommendation: prefer the vLLM-Omni recipe for real training
**Train-side colocate is significantly slower than the engine-backed path, and is not the recommended way to actually train Qwen3-Omni.** For any real run, use the recipes from #235:
* `examples/ar/qwen3_omni_video_r1_gspo_lora_vllm_omni_1x8.yaml`
* `examples/ar/qwen3_omni_video_r1_gspo_lora_vllm_omni_1x4.yaml`
The gap is structural, not a tuning artifact. `TrainsideRolloutEngine` decodes through the eager HF forward at `forward_batch_size: 1`, so per rank it performs `samples_per_prompt × max_new_tokens` **sequential** forwards, with no paged KV cache and no CUDA graphs, while FSDP re-all-gathers the sharded parameters on every decode step. Decode therefore dominates step time and scales linearly in generation length. `old_logp_source: replay` adds a second replay pass per step on top of that.
Measured on 4×96GB with `Qwen3-Omni-30B-A3B-Instruct`, `num_devices=4`,
`batch_size=4`, `samples_per_prompt=4` (16 responses/step):
| generation length | wall-clock per step |
| --- | --- |
| `max_new_tokens=256` | ~4.5–8 min |
| `max_new_tokens=512` (recipe default) | a single rollout had not finished after ~11 min of decode |
By contrast the vLLM-Omni stage in #235 runs TP=4 with chunked prefill, paged attention, and up to 64 concurrent sequences / 32,768 batched tokens, and its 1x8 recipe trains **8,192-token** responses at 64 responses per update — a generation length 16× beyond what is practical here. Train-side should be read as a correctness and portability tool, not a throughput path.

## Test Plan
No test; Reason: no need.

## Checklist


CjhHa1:
closed as this PR is useless now

PR #199, opened 2026-07-10
feat(eval): add unified standalone evaluation entrypoint (#182)

### LINKED ISSUE #182: How to evaluate a given ckpt?
For example, https://github.com/Tencent-Hunyuan/UniRL/tree/main/FlowDPPO, how could one eval the huggingface ckpt on both in-domain and out-of-domain benchmarks?

CjhHa1:
## Summary

Adds one Hydra-native entrypoint for evaluating AR and diffusion checkpoints without entering the training loop or using the `num_rollouts=0` workaround.

- `python -m unirl.eval --domain ar|diffusion` selects the trainer while forwarding `--config-name` and all remaining overrides to Hydra unchanged.
- Only the selected domain's trainer is imported, avoiding cross-domain initialization and optional dependency coupling.
- `BaseTrainer.run_eval(load_dir=...)` optionally restores a checkpoint, initializes logging, calls the trainer's existing `evaluate()` implementation once, and closes the wandb session in a `finally` block.
- Documentation covers full-weight and LoRA checkpoint evaluation, eval datasets, and Hydra's required `+eval_*` / `+load_dir` syntax for keys not declared in recipe YAMLs.

## Related Issue

Closes #182.

## Test Plan

- Unified Hydra dispatch and config composition:
  - `python -m unirl.eval --domain ar --config-name=ar/qwen_vl_grpo_geo3k_mc_4x8 --cfg job`
  - `python -m unirl.eval --domain diffusion --config-name=diffusion/sd3/sd3_flowdppo --cfg job`
- Repository checks on the changed files: `SKIP=no-commit-to-branch pre-commit run --files unirl/eval.py FlowDPPO/README.md examples/README.md unirl/README.md unirl/trainer/README.md --show-diff-on-failure`.
- AR runtime validation on 8×H20:
  - Geo3K train-side evaluation: `eval_acc(avg@2 over 8)=0.25`.
  - Qwen3-4B SGLang smoke test: the standalone evaluation completed successfully.
  - Full AIME evaluation: `eval_acc(avg@16 over 60)=0.0250`.
- No new unit test is included; the entrypoint is a thin dispatcher over the existing trainer evaluation paths and was validated through Hydra composition and GPU evaluation runs.

## Compatibility / Risk

- Additive CLI and `BaseTrainer` API changes; existing training and periodic evaluation behavior is unchanged.
- `--domain ar|diffusion` is required. Callers should pass `--config-name` explicitly when not using that domain's default eval recipe.
- No checkpoint or data-format changes. `+load_dir` uses the existing backend checkpoint restore path and rollout weight synchronization.
- Eval-only overrides absent from recipe YAMLs must use Hydra's `+` prefix.
- Runtime validation covered AR; diffusion was validated at Hydra composition level only.

## Reviewer Notes

- Scope is limited to the synchronous AR and diffusion trainers; async AR, PE, and unified-model evaluation are not added here.
- Please focus review on argument forwarding, lazy domain dispatch, and the checkpoint restore → wandb lifecycle → single `evaluate()` call sequence.
- AI-assisted; the author reviewed the final changes.

## Checklist


haonan3:
Hi @CjhHa1 — thanks for putting this together, and especially for validating it on real 8×H20 runs. The Hydra forwarding and the docs on LoRA-vs-full-weight loading are genuinely useful. That said, after thinking through what #182 actually needs, I've decided to go a different route and close this PR. Reasoning below, since the code itself is fine:

**1. What #182 asks for is benchmark reproduction, not another path into `evaluate()`.** The follow-up in the issue thread makes this concrete: trainer eval on SD3.5-medium gave `8.68` while the paper table says `12.4`. That gap is structural — `evaluate()` reports a *training-time* signal (our eval prompt list, our sample counts, the reward-service scorer as configured), which will generally not match an official benchmark protocol (official prompt sets, seeds, sample counts, scorer version, aggregation). A standalone entrypoint over the same `evaluate()` makes that number easier to produce, but it's still a number users can't map to paper tables — which is exactly the confusion in the issue.

**2. Core-surface cost.** `unirl/eval.py` mirrors each trainer's constructor wiring, so every trainer-signature change now has a third place to update, and `BaseTrainer.run_eval()` becomes public API for what `num_rollouts=0` already covers today. We're trying to keep the core package surface minimal, so this trade only makes sense if it fully solves the user-side problem — per point 1, it doesn't.

**Instead:** I'm drafting a top-level `benchmarks/` suite (no core changes) that treats checkpoints as artifacts (HF full weights, or LoRA adapters via `unirl.tools.export_adapter`) and evaluates them under *official* protocols, per modality:

- **image**: GenEval / GenEval2, DPG-Bench, UniGenBench++, HPSv3 / PickScore — with reference numbers for the released `Tencent-Hunyuan-Multimodal-RL` adapters;
- **video**: VBench / VBench-2.0 for the wan / hunyuan_video / ltx2 families;
- **text / VL**: AIME, MATH500, GPQA for qwen3; Geo3K test, MathVista for qwen_vl;
- **speed**: a subfolder for same-setting throughput comparison against verl-omni (Qwen-Image FlowGRPO, Wan2.2 DanceGRPO — same GPUs, reward, rollout backend where possible).

Your docs work here (checkpoint-loading modes, the `+eval_*` Hydra syntax) is exactly the usage guidance those benchmark READMEs need, and your GPU validation runs would carry over directly — I'd be glad to have your review (or hands) on that PR. Keeping #182 open until the suite lands. Thanks again 🙏

PR #195, opened 2026-07-09
feat(sft): generic SFT domain + qwen3(AR/CE) & sd3(diffusion/MSE) tasks

leviking98z-rgb:
## Summary

Adds a **generic supervised-finetuning (SFT) training domain** to UniRL — the first non-RL domain — and proves it is **model-agnostic** by shipping two validation task adapters over **already-in-tree** models: **Qwen3 (autoregressive, cross-entropy)** and **SD3.5 (diffusion, flow-matching MSE)**.

Where GRPO/ReFL do `rollout → reward → advantage → update`, SFT does the plain supervised loop `read data → compute loss → backward → update` — no rollout engine, reward, advantages, or weight sync. It mirrors the ReFL domain's shape (`BaseTrainer` + FSDP-wrapped policy), the codified template for non-GRPO domains.

**Design: the skeleton knows no model.** `SFTPolicy` resolves a **task adapter** from config (`task_target` → `get_class(...).from_config(model_config)`) and drives it; everything family-specific — which model, how to read a record, and *what the loss is* — lives in the task. That is why one skeleton serves both AR and diffusion:

| Task | Family | Loss |
|------|--------|------|
| `Qwen3SFTTask` | autoregressive | masked next-token **cross-entropy** |
| `SD3SFTTask` | diffusion | flow-matching **velocity MSE** (target `noise - x0`) |

The contract is codified in `unirl/train/sft/task.py` (`SFTTask` Protocol + `SFTTaskBase` ABC).

### What's here
- **Skeleton** (extracted from #181, cosmos-free): `unirl/trainer/sft.py`, `unirl/train/sft/{__init__,policy,data,task}.py`, `unirl/train_sft.py`.
- **Qwen3 task** (`unirl/models/qwen3/sft_task.py`): reuses the AR RL path's `_replay_aware_forward` (chunked, FSDP2-safe per-token logp), so CE = `-mean(logp)` over the response with prompt tokens masked.
- **SD3 task** (`unirl/models/sd3/sft_task.py`): reuses `SD3Bundle` / `SD3TextEmbedStage` / `SD3DiffusionStep.predict_noise` / `get_sigma_schedule`; adds the VAE image→latent encode. Target `noise - x0` matches `FlowSDEStrategy.step`'s velocity convention.
- **Recipes** `examples/sft/{qwen3,sd3}_sft_smoke.yaml` + `examples/sft/README.md`.

## Relationship to #181

[...]

PR #200, opened 2026-07-10
feat(cosmos3): Cosmos3 omnimodal SFT task adapters (videopred + action BC)

leviking98z-rgb:
## Summary

Adds Cosmos3-Nano (16B omnimodal MoT world model) supervised training on UniRL's shared SFT stack — video prediction and policy-mode action behavior cloning:

`SupervisedDataSource → Cosmos3SupervisedTrackBuilder → Part → TrainStack → Cosmos3JointFlowMatchSFT`

Model-specific code is confined to `unirl/models/cosmos3/` (bundle, packed joint stage, conditions, track builder) plus one loss in `unirl/algorithms/` and the droid_100 data-prep util. Key behavior:

- Reuses diffusers 0.39's own `Cosmos3OmniPipeline` helpers (`tokenize_prompt`, `_prepare_text/vision/action_segment`, `_encode_video`) so training conditioning is assembled exactly like inference.
- Keeps latent frame 0 clean (the observation) and jointly noises future video plus the action chunk with one sigma; velocity target `ε − x0` matches the UniPC `flow_prediction` convention.
- Stores the transformer fp32 (stable optimizer master) while FSDP gathers bf16 compute weights; the und (text) stream is frozen, only the gen tower trains.
- Resolves flow shift by the official short-edge tiers (256 → 3, 480 → 5, 720 → 10), falling back to the checkpoint scheduler.
- Action BC balances vision/action MSE 1:1 (10x vision + 10x action).

**2026-08 update — ported to the sample-native stack.** #214 removed `RolloutReq`/`RolloutResp`/`RolloutTrack` one day after the previous port, leaving this branch unimportable against main (and, through the old `unirl/train/sft/__init__.py` import, it would have broken every SFT recipe's import on merge). The branch is now merged with current main, and:

- the track builder emits a `Part` and moves into the model package (`unirl.models.cosmos3.track_builder`) — `unirl/train/sft/` keeps only model-agnostic builders per its package contract, and the SFT domain no longer imports cosmos3 code at package-import time;
- `Cosmos3Pipeline.generate` follows the `Sample → Sample` pipeline contract (still `NotImplementedError`: this is an SFT-only integration);
- the recipe headers absorb the deleted `examples/sft/README.md` and gain the previously missing `ENTRY=train_sft` launch prefix.

## Related Issue

N/A (supersedes the closed #181 / #195 lineage).

## Test Plan

CPU-verifiable surface (no test files added, per repo policy):

- `pre-commit run --files <all touched files>` — all hooks passed, including `recipe _target_ paths resolve` and `experimental-tier boundaries hold`.
- Hydra composition: `python -m unirl.train_sft --config-name sft/{cosmos3_droid100_videopred,cosmos3_droid100_action_bc,qwen3_sft,sd3_sft} --cfg job` — all four compose (the two pre-existing recipes prove the SFT domain import stays intact).
- Uncommitted CPU harness against real **diffusers 0.39.0** (tiny `Cosmos3OmniTransformer` + tiny `AutoencoderKLWan` + the real `Cosmos3OmniPipeline`), 12/12 checks passed: package imports (`unirl.train.sft` / `unirl.algorithms` / `unirl.models.cosmos3`); every recipe `_target_` dotpath; the und-freeze regexes classified against the real transformer parameter tree; builder → `Part` → `Part.slice` micro path with packed-condition row alignment; action-BC and videopred micro forward + loss + backward with gradients confined to the gen stream; deterministic `evaluate_loss` under per-sample seeded draws; `_eval_pad` rows carrying zero weight.
- `python -m compileall` on all changed modules.

Not run: accelerator smoke of the ported path — the pre-port integration had 1x8 H20 runs (videopred smoke + 150-step action BC); a fresh 1x8 smoke of both recipes is the remaining merge gate.

## Compatibility / Risk

- Cosmos3 needs `diffusers>=0.39`, exposed as the optional `cosmos3` install extra; the base engine floor stays `>=0.38` (#288 already made the in-tree models 0.39-compatible).
- Purely additive; no shared-code behavior changes. Risk concentrates in the new package until the fresh accelerator smoke passes.
- The supplied `droid_100` recipes are debug-scale and use its collapsed 7-D action layout rather than the canonical 10-D `droid_lerobot` layout; documented in the recipe headers.

## Reviewer Notes

- AI-assisted (Claude Code): the sample-native port, a diffusers v0.39.0 source audit of every touched API (pipeline ctor, `tokenize_prompt`, the `_prepare_*_segment` key contracts, transformer forward kwargs/returns, `_EMBODIMENT_TO_DOMAIN_ID`, viewpoint templates, action tier bins), and the CPU harness above. Duplicate-work check: no other open cosmos PR; supersedes closed #181/#195.
- Suggested review order: `unirl/models/cosmos3/track_builder.py` (Part port + relocation), `unirl/models/cosmos3/pipeline.py`, then the loss.

## Checklist


xshrz (commented):
Consistency nit (non-blocking): the Cosmos3 classes don't inherit the base types the other families use.

- `Cosmos3SFTTaskBase` (unirl/models/cosmos3/sft_task.py) is a plain class, whereas `Qwen3SFTTask` and `SD3SFTTask` both inherit `unirl.train.sft.task.SFTTaskBase`.
- `Cosmos3Bundle` (unirl/models/cosmos3/bundle.py) is a plain class, whereas SD3Bundle / Qwen3Bundle inherit `unirl.models.types.bundle.Bundle`.

Duck-typing works today (SFTPolicy only checks the Protocol, and FSDPBackend only duck-types the bundle), so this is purely a code-health point. But inheriting the base types would:
- keep the Cosmos3 adapters consistent with the qwen3/sd3 ones (one obvious pattern for "how to add a family");
- surface a clear `NotImplementedError` / abstractmethod error early if a required method is missing or renamed, instead of failing deep in the training loop;
- give a single place for shared helpers to grow.

Suggest `class Cosmos3SFTTaskBase(SFTTaskBase)` and `class Cosmos3Bundle(Bundle)` (adjusting for whatever the base ctors require).

Ideny42 (commented):
Aligning the Cosmos3 SFT with the official cosmos-framework recipes

I reviewed this against NVIDIA's [cosmos-framework](https://github.com/NVIDIA/cosmos-framework) SFT recipes — specifically vision_sft_nano (video) and action_policy_droid_nano (action policy), both built on nano_model_config — and made a few changes so the training matches the official flow-matching setup. Since this PR trains through the diffusers Cosmos3OmniPipeline/Cosmos3OmniTransformer, "align" here means matching the numeric/training conventions, not the code.

Changes I adopted

fp32 optimizer master (was bf16). Official uses fsdp_master_dtype=float32 ("fp32 master + bf16 compute"). Note the recipe-level FSDPConfig.master_dtype: fp32 can't be used here — with freeze_understanding, every MoT layer mixes frozen (und) + trainable (gen) params and the wrap only upcasts the trainable ones, so FSDP2 rejects the {bf16, fp32} group. Instead I load the transformer in fp32 (new master_precision, default fp32) and let mixed-precision do the bf16 compute. It matters: bf16-master runs showed ~97% integer-quantized grad-norms on action, and ~1.4 dB lower sampling PSNR on video (16-sample avg).
Resolution-keyed flow_shift. Official keys the σ shift by the short edge — {256:3, 480:5, 720:10}. The recipe hardcoded 5.0 (the 480p value) on tier-256 (192×320) data. I made it a per-sample lookup by resolution (→ 3.0 for tier-256) for both video and action, and also align the eval scheduler's shift to the trained one (the diffusers __call__ has no flow_shift arg, so eval was silently using the checkpoint default).
1:1 vision:action loss weight (action). Official balances the modalities (loss_scale=10 on vision + action_loss_weight=10 → 1:1); the recipe was effectively 1:10, so I set vision_loss_weight=10.
Considered but deliberately NOT adopted

Fresh re-init of the action heads + 5× LR. Official Policy-DROID re-inits the action heads and trains them at 5× LR. UniRL's SFT uses a single LR; in an A/B, fresh-init without the LR boost converged clearly worse than warm-starting the base heads, so I kept warm-start (the PR's original behavior).
waver σ distribution. Official video uses waver (and DROID action shares it via independent_action_schedule=False); this PR uses logit-normal. I tried waver — on this debug set its lower train loss turned out to be a σ-sampling artifact, and actual sampling quality (PSNR/SSIM) was no better (slightly worse), so I left logit-normal.
Known remaining gaps (debug-scale / dataset-limited — intentionally left as-is)

Both tasks train on lerobot/droid_100 (a small robot debug set). Consequences: single-domain short clips, plain task-string prompts instead of the structured-JSON captions the model was trained on, and — for action — droid_100 only ships a 7-D collapsed action instead of the 8-D joint_pos that the official Cosmos3-DROID action recipe trains on (re-training the action head for that layout; not adopted here), plus no proprioceptive state and no concat_view multiview. These are data limitations, not code issues.
No EMA / no CFG-dropout / smaller LR & step budget — debug scale.

The plots below compare baseline (original recipe: bf16 master, hardcoded/checkpoint flow_shift, 1:10 loss) vs fixed (this PR: fp32 master + resolution-keyed flow_shift + 1:1 loss), at matched step counts.

Action BC (500 steps) — baseline = cosmos3_action_bc-v2-baseline, fixed = cosmos3_action_bc-v2-warmstart.

Image

Video prediction (500 steps) — baseline = cosmos3_videopred-v3b-baseline-allmedia, fixed = cosmos3_videopred-v3b-current-allmedia.

Image

Image

Video sampling quality — 16 generated clips at step 500, vs. ground-truth future frames (the per-step val_vision_mse curve is flat/noisy, so this is the metric that actually reflects sample quality):

Clipboard_Screenshot_1784619575

I've pushed these as three commits on my fork, stacked directly on the current PR head (55efc2d), in case it's easiest to pull them in:

Ideny42/UniRL @ feat/cosmos3-support

66fdb1e — fix(cosmos3): fp32 optimizer master (bf16 compute) for stable SFT
3da478e — fix(cosmos3): resolution-keyed flow_shift for train and eval
3e17aa4 — chore(cosmos3): retune droid100 action recipe (1:1 loss, auto flow_shift)

xshrz:
The main blocker is that this PR still uses the obsolete `SFTPolicy` / `SFTTask` architecture from #195. That PR was closed without merging, while the current SFT implementation landed through #209.

Current main uses:

`SupervisedDataSource -> SupervisedTrackBuilder -> RolloutTrack -> TrainStack -> StageAlgorithm`

This PR instead lets each model-specific `SFTTask` load records, run the model, and compute the loss directly. Its recipes also use the old `policy + num_rollouts` schema, so they cannot be instantiated by the current `train_sft.py`.

Requested SFT changes:

1. Rebase onto current main and remove the old `unirl/train/sft/{policy,task,data}.py` implementation.
2. Route Cosmos3 training through the shared bundle, pipeline, model stage, `TrainStack`, and `StageAlgorithm`.
3. Add a supervised video/action track builder using the current `SupervisedDataSource` record format.
4. Add a joint flow-matching algorithm for first-frame-conditioned video and action supervision. The existing `FlowMatchSFT` cannot directly represent the joint video/action objective.
5. Rewrite the recipes using the current `bundle / pipeline / backend / algorithm / stack / track_builder / data_source` schema.
6. Add SFT integration tests covering config composition, record-to-track conversion, loss masking, and train/eval execution.

The Cosmos3-specific model and packing logic can be reused, but the SFT integration layer must be ported to the current architecture before this PR can be merged.

haonan3 (approved):
LGTM

PR #202, opened 2026-07-12
feat(trainer): add periodic eval to PE / ReFL / UnifiedModel

Ideny42:
## Summary

Make in-training **eval** work — and be verifiable — across **all** trainers.
Eval (`evaluate()` + scheduling in the `train()` loop, logged under `eval/*`)
previously existed only in `DiffusionTrainer` and `ARTrainer`. `PETrainer`,
`RewardBackpropTrainer` (ReFL), and `UnifiedModelTrainer` had **no eval path at
all** (0 references — passing `eval_interval` was silently ignored). This PR adds
eval to those three, mirroring `DiffusionTrainer`:

- pull `eval_num_prompts` eval prompts (`run.eval_data_path`), force the diffusion
  sub-block onto a deterministic ODE (`eval_eta`, default 0), score the image track
  with the configured reward (PickScore) → `eval/reward`, and run `evaluate(k)`
  **before** `save(checkpoint-k)` at the same step so a resumed checkpoint
  reproduces the same eval (the basis of the A/B consistency check below).
- entry scripts (`train_pe` / `train_refl` / `train_unified_model`) read `eval_*`
  via `cfg.get`.
- **PE** additionally gains two-side checkpointing (it had none) so its
  consistency run can resume; **ReFL** gains `ReFLPolicy.eval_sample` (an
  eval-mode `no_grad` DRaFT-K sampler, no autograd graph).

It also aligns `ARTrainer`'s eval metric with the others: AR now reports
`eval/reward` alongside its existing `eval/acc` (the MC reward is 0/1, so the two
are numerically equal). This gives every trainer an `eval/reward` key, so AR eval
runs share that axis with the diffusion / PE / ReFL / UnifiedModel curves.

Additive by design: all eval is gated on `eval_interval > 0`, so the default
(`0`) leaves every trainer's training path byte-for-byte unchanged.

## Follow-up commits on this branch (maintainer, post-review)

Four commits added during review (haonan3 + agent), same additive-by-default discipline:

- `6813c29f` **eval CFG knob**: `eval_cfg_text_scale` (default 4.0) now reaches PE / ReFL / UnifiedModel eval, mirroring `DiffusionTrainer` (Bagel-style params take `cfg_text_scale`, standard `DiffusionSamplingParams` fall back to `guidance_scale`). Recipes train at cfg=1.0; eval no longer silently inherits that — set the knob to the train value to eval in the training regime instead.
[...]

PR #117, opened 2026-06-24
perf(vllm-omni): group SD3  and Qwen-Image prompt rollouts 

CjhHa1:
## Summary

- Group SD3 vLLM-Omni rollout requests by prompt and use `num_outputs_per_prompt=samples_per_prompt` instead of sending one diffusion request per image sample.
- Keep downstream UniRL semantics sample-level by preserving sample-level x_T recipes, expanding prompt-level text captures back to sample-level conditions, and validating grouped prompt layout.

## Related Issue

Fixes #108

## Test Plan

- `python3 -m py_compile unirl/rollout/engine/vllm_omni/adapters/sd3.py unirl/rollout/engine/vllm_omni/adapters/dit.py unirl/rollout/engine/vllm_omni/utils/prompts.py unirl/rollout/engine/vllm_omni/utils/noise.py unirl/rollout/engine/vllm_omni/pipelines/_shared/interception.py`
- `python3 -m ruff check unirl/rollout/engine/vllm_omni/adapters/dit.py unirl/rollout/engine/vllm_omni/adapters/sd3.py unirl/rollout/engine/vllm_omni/pipelines/_shared/interception.py unirl/rollout/engine/vllm_omni/utils/__init__.py unirl/rollout/engine/vllm_omni/utils/noise.py unirl/rollout/engine/vllm_omni/utils/prompts.py` → All checks passed
- Pre-commit hooks on commit/push → passed
- Taiji 1x8 H20 SD3 vLLM-Omni E2E alignment smoke, legacy upstream/main, `diffusion/sd3/sd3_vllmomni`, `num_rollouts=1`, PickScore dataset/checkpoints under `/apdcephfs_fsgm3/...`: `unirl-sd3-vllmomni-e2e-legacy-main-zw-p1-0624-v1` → success, `rollout/num_samples=768`, `perf/generate_time_s=43.34967`, `train/grad_norm=1e-05`
- Taiji 1x8 H20 SD3 vLLM-Omni E2E alignment smoke, grouped branch, same recipe/data/checkpoints, `num_rollouts=1`: `unirl-sd3-vllmomni-e2e-grouped-main-zw-p1-0624-v1` → success, `rollout/num_samples=768`, `perf/generate_time_s=34.88510`, `train/grad_norm=1e-05`
[...]

PR #203, opened 2026-07-13
perf(vllm-omni): batch BAGEL grouped t2i rollout into one packed generate_image

zzhuoxin1508:
## Summary

- Group BAGEL vLLM-Omni **t2i** rollout requests by prompt and use
  `num_outputs_per_prompt=samples_per_prompt` (spp) instead of sending one diffusion
  request per image sample. The shared prompt KV is built once and replicated ×spp;
  all spp images are denoised in one packed `generate_image` per step, then the
  engine splits them back per request.
- Keep downstream UniRL semantics **sample-level** by preserving sample-level x_T
  recipes (grouped-span slice of `resolve_request_noise`), splitting log-prob +
  trajectory per image (`[spp, T+1, seq, C]`), and validating the grouped prompt
  layout via `grouped_texts_from_req`.
- **Bugfix ** #202 renamed the
  `pack_initial_noise_extra_args()` keyword `n_prompts=` -> `n_samples=` and updated the
  shared `dit.py` caller, but `BagelInputAdapter` (merged from a side branch) still
  passes `n_prompts=`. Since that rename, every `bagel_t2i` rollout that packs a
  driver-authored x_T crashes on its first generate with `TypeError: pack_initial_noise_extra_args() got an
  unexpected keyword argument 'n_prompts'
## Compatibility / Risk

- Packed path is scoped to **pure bagel_t2i + cfg≤1** (`_is_packable_t2i` /
  `_is_batchable_t2i`). cfg>1 / text-output / i2i keep the **sample-level** layout
  (`num_outputs_per_prompt=1`); a mismatched `num_outputs>1` on the worker raises
  instead of a broken silent fallback. **spp=1 is byte-identical to before.**
- Missing CFG keys are treated as non-batchable (upstream Bagel defaults absent
  keys to CFG-ON @4.0/1.5).
- Packed images are isolated by varlen block-diagonal attention, so a packed forward
  == spp separate forwards. Only per-step SDE RNG differs (<0.1%, no collapse, ratio 1.0).

## Test Plan

- **8×H20 BAGEL vLLM-Omni E2E, 3-way A/B** (spp=16):

  | variant              | e2e s/rollout | ratio  |
  | -------------------- | ------------- | ------ |
  | trainside            | ~187          | 1.0000 |
  | unmodified vllm_omni | ~150          | 1.0000 |
  | batched (this PR)    | ~139          | 1.0000 |

- Confirm bagel_t2i with driver x_T (recipe / `initial_noise_batch`) no longer raises
  on `pack_initial_noise_extra_args`.
- Smoke cfg>1 + spp>1: stays on sample-level path (no pack).

## Reviewer Notes

- Reuses #117's grouped-span `resolve_request_noise` — no new noise code.
- Adapter/pipeline pack gates must stay aligned (`_is_packable_t2i` ↔ `_is_batchable_t2i`).


CjhHa1 (approved):
LGTM

PR #205, opened 2026-07-14
fix(sampling): make omitted AR top-k unrestricted

### LINKED ISSUE #204: Make omitted AR top-k sampling unrestricted
## Problem

`ARSamplingParams.top_k` defaults to `1024`. Once Hydra constructs the dataclass, rollout adapters cannot distinguish that implicit default from an explicitly requested `top_k: 1024`.

For SGLang, this means a recipe that omits `top_k` sends a finite top-k limit of 1024. Explicit `top_k: 0` is already translated correctly to SGLang's `-1` (unrestricted), while explicit positive values pass through unchanged.

The current contract is therefore surprising:

| Recipe input | Runtime AR value | SGLang value |
| --- | ---: | ---: |
| omitted | 1024 | 1024 |
| `top_k: 0` | 0 | -1 |
| `top_k: N` | N | N |

This is not the same as accidentally inheriting Qwen's model default (`top_k=20`), and there is no direct evidence that `1024` alone causes GRPO advantage collapse. However, RLVR research does establish that restricted rollout diversity can increase homogeneous-reward groups and eliminate their advantage signal: [DAPO](https://arxiv.org/abs/2503.14476) and [Advantage Collapse in GRPO](https://arxiv.org/abs/2605.21125).

## Current scope

The standalone AR/VLM SGLang examples set `top_k` explicitly. Three composed PE+SGLang recipes and four trainside/composed recipes omit it and currently inherit `1024`. Their existing behavior should not change accidentally as part of clarifying the default.

## Desired outcome

- Omitting `ARSamplingParams.top_k` has a documented, engine-independent meaning: unrestricted sampling.
- Explicit `top_k: 0` remains unrestricted.
- Explicit positive values, including `1024`, remain unchanged.
- Existing recipes that relied on the old implicit value make that value explicit so the compatibility impact is reviewable.
- Regression tests cover default, zero, positive, and absent-AR translation behavior.

A separate measured tuning change can decide whether existing recipes should move from `1024` to unrestricted sampling.

morluto:
## Summary

Make omitted AR `top_k` mean unrestricted sampling instead of silently inheriting a finite `1024` limit. SGLang keeps its existing adapter translation from the shared Hugging Face/vLLM convention (`0`) to SGLang's disabled value (`-1`). When a request has no typed AR sampling parameters, the SGLang resolver now honors `SGLangEngineConfig.top_k` instead of hard-coding unrestricted sampling.

| Recipe input | Canonical AR value | SGLang wire value |
| --- | ---: | ---: |
| omitted | 0 | -1 |
| `top_k: 0` | 0 | -1 |
| `top_k: N` | N | N |

The seven in-tree recipes that previously inherited `1024` now declare it explicitly, so their rollout behavior is unchanged. The resolver documentation now describes the sampling contract without claiming that every finite top-k value causes GRPO advantage collapse.

## Related Issue

Fixes #204

## Test Plan

- Resolver smoke assertions: missing AR honors configured `top_k=1024`, configured `top_k=0` maps to SGLang `-1`, and typed AR `top_k=32` overrides the config → passed
- `SKIP=no-commit-to-branch uvx pre-commit run --files unirl/rollout/engine/sglang/utils/sampling.py` → all hooks passed
- `python3 scripts/check_recipe_targets.py` → 1,825 targets resolved
- Hydra compose/resolve smoke for the seven compatibility recipes → each resolved `sampling.ar.top_k=1024`
- GPU training/rollout smoke: Not run; this changes configuration defaults while preserving every in-tree recipe's effective sampling behavior. No representative GPU environment was available.

## Compatibility / Risk

Out-of-tree recipes that omit `top_k` now use unrestricted sampling instead of `1024`. Explicit zero and positive values are unchanged. All in-tree AR recipes now specify `top_k`, preserving their current behavior. Requests without typed AR parameters use the SGLang engine config fallback. There is no checkpoint or data-format change.

## Reviewer Notes

Suggested order: review the shared default and SGLang config fallback first, then scan the seven one-line recipe compatibility pins.

AI assistance was used to investigate, implement, test, and review this change. A targeted search found no overlapping open issue or PR before filing #204. Autoreview reported no accepted or actionable findings.

## Checklist


tests/rollout/engine/sglang/test_sampling.py
```diff
@@ -0,0 +1,54 @@
+from types import SimpleNamespace
```
CjhHa1: Can you please remove this test as UniRL does not require tests now

CjhHa1: Have fixed my self

CjhHa1 (approved):
LGTM

PR #207, opened 2026-07-14
refactor(meta-init): unify the meta-init build/restore path across bundles (fixes garbage RoPE on VeOmni / sharded-FSDP)

zzhuoxin1508:
## Summary

This started as a bug fix and turned into a refactor of the meta-init machinery.

**Root cause (unchanged):** under the meta-init path (`meta_init_transformer=true`,
used by the VeOmni / sharded-FSDP `load_sharded` backends) several bundles built the
transformer on `torch.device("meta")`, which forces **non-persistent buffers** onto
meta too — exactly the init-computed tables the checkpoint does not carry
(`WanRotaryPosEmbed.freqs_cos/freqs_sin`, SD3's sincos `pos_embed`, rope tables).
After `to_empty()` these hold **garbage**, and the no-op `init_weights` never rebuilds
them.

For WAN 2.1 this silently breaks RoPE: self-attn q/k get multiplied by junk cos/sin.
Because the multiply is by ~zero garbage, the gradient to the `attn1.to_q` /
`attn1.to_k` LoRA adapters is **killed** — those never train while `attn2` / `v` /
`out` train normally, so LoRA-RL reward plateaus below the classic-FSDP baseline.

## What changed (refactor)

The base already restored captured state via `bundle._meta_init_state` in
`load_trainable_weights` — but only bundles that actually *captured* benefited, and
the capture path was split across two mechanisms. This PR unifies it:

- **`unirl/models/types/meta_init.py`**
  - Add `build_meta_init_transformer(factory, *, dtype)`: builds under
    `accelerate.init_empty_weights(include_buffers=False)` (params on meta,
    non-persistent buffers/attrs real on CPU), captures them via
    `capture_init_state`, casts dtype, and no-ops `init_weights`. Returns
    `(transformer, captured)`.
  - Remove `finalize_meta_init` (folded into the above) and
    `stamp_init_state_restore` (the model-bound deferred-closure restore, which can
    be dropped across Ray-actor boundaries — the bundle carrier already covers it,
    robustly).
  - Keep `capture_init_state` / `restore_init_state`.
- **All 7 meta-init bundles** (`wan21`, `sd3`, `qwen3`, `qwen_vl`, `flux2_klein`,
  `hunyuan_video`, `hunyuan_video15`) now build via `build_meta_init_transformer`
  and stash the capture on `bundle._meta_init_state`.
- **Restore is unchanged in placement:** `load_trainable_weights` replays
  `restore_init_state(model, bundle._meta_init_state)` after the sharded weight load
  (Ray-robust — a plain picklable dict survives crossing actors).
- **`unirl/train/backend/sharded_load.py`** — comment cleanup only.

Net: 9 files, **+128 / −190** — a single build/restore path, less code.

## Related Issue

N/A

## Test Plan

**WAN 2.1-T2V-1.3B FlowGRPO, VeOmni backend, 720p, 2×8 GPU:**
curve in comment
**SD3.5-medium FlowGRPO, VeOmni backend, 8 GPU (`diffusion/sd3_trainside_veomni`):**
- Ran end-to-end on the fix branch and confirmed **reward grows normally** 

## Compatibility / Risk

Behavioral change only on the `meta_init_transformer=true` path (VeOmni / sharded
FSDP). The classic-FSDP `from_pretrained` (eager) path is untouched. No
config/checkpoint/API format changes. Models with no init-computed non-persistent
buffers are unaffected — `restore_init_state` is a no-op when the capture is empty.

## Reviewer Notes

AI-assisted. Root-caused and validated end-to-end on WAN 2.1 (reward alignment +
gradient map). The capturing build path lives in the shared
`build_meta_init_transformer`, so every meta-init model (e.g. SD3's sincos
`pos_embed`) gets the same restore for free.

## Checklist


zzhuoxin1508:
Clipboard_Screenshot_1784083648

leviking98z-rgb:
Reviewed the diff and traced the actual call chain on the branch — the core refactor is sound and I don't see a regression risk. A few points before this lands.

## Correctness — looks good

Removing `stamp_init_state_restore` (the deferred-op restore path) is safe: in both backends the lifecycle calls `load_trainable_weights` → `apply_deferred_ops` unconditionally and in that order (`veomni/backend.py:182,192`, `fsdp/backend.py:105,117`), and `load_trainable_weights` runs `restore_init_state(model, bundle._meta_init_state)` unconditionally (`sharded_load.py:72`). So the bundle-carrier restore always runs, one step earlier than the old deferred op. This matters for `sd3` and `qwen_vl`, which previously had *only* the deferred path (neither set `bundle._meta_init_state`) — the switch to carrier-only is behavior-equivalent. Capture happens before the dtype cast, so the tables are grabbed at init precision and downcast on restore, matching the eager `from_pretrained → .to(dtype)` path.

## Main issue — the "all 7" claim is inaccurate; `qwen_image` is an 8th meta-init bundle left on the old path

There are actually **8** bundles with `meta_init_transformer`, not 7. `unirl/models/qwen_image/bundle.py` is not migrated — it still builds under `torch.device("meta")` (`:161`) with its own `_rebuild_meta_rope_modules` (`:43`) plus the inline non-persistent-buffer warning. That is exactly the "split across two mechanisms" this PR sets out to eliminate — and `qwen_image` is the very model the description cites ("Qwen-Image's complex rope tables") as the motivating `__dict__`-tensor case.

To be clear, `qwen_image`'s existing mechanism is *correct* (it re-instantiates the param-less `QwenEmbedRope` on CPU, which is shard-exempt and survives `to_empty`), so this is a completeness/consistency gap, not an active bug. But it should be resolved one of two ways:

- **migrate it** into `build_meta_init_transformer` — under `init_empty_weights(include_buffers=False)` its `pos_freqs`/`neg_freqs` land real on CPU and are captured as `__dict__` attrs, so `_rebuild_meta_rope_modules` can be deleted (genuinely one path); or
- **explicitly scope it out** and correct the description ("7 of 8", with a note on why).

Either way the "all 7 / every meta-init model gets the same restore for free" wording needs fixing.

## Tests

There's no unit coverage for the capture/restore/build machinery. Given this is silent-fail, load-bearing code (garbage buffer → wrong forward, no exception, reward just plateaus), a cheap CPU-only test would be worth it: build a small module with a non-persistent buffer + a `__dict__` tensor under `init_empty_weights(include_buffers=False)`, capture, `to_empty`, restore, assert values match; plus the `captured=None` / empty-dict no-op case.

## Minor

- Test Plan validates 2 of 7 (WAN2.1, SD3.5). The 4 bundles switching build context from `torch.device("meta")` to `init_empty_weights` (`hunyuan_video`, `hunyuan_video15`, `flux2_klein`) and `qwen_vl` (deferred → carrier) aren't exercised. Worth a smoke test — note that if any non-persistent buffer is param-derived it would now land on meta and trip the `capture_init_state` guard (hard fail at construction) rather than silently degrading.
- Branch is BEHIND `main` — needs a rebase.
- Body says +128/−190; actual is +127/−160.

Overall: solid, correct direction. Happy to approve once the `qwen_image` inconsistency is resolved (migrate or document) and ideally a capture/restore unit test is added.

zzhuoxin1508:
## Smoke-test results: remaining 6 meta-init bundles all covered

Every changed bundle not already validated in the PR (sd3 / wan21 were) ran **10 rollout
rounds on the VeOmni FSDP2 meta-init path** (`meta_init_transformer=true`), 8×H20;
qwen_vl was additionally reproduced on 2×8. Pass = clean 10/10 rounds, no capture-guard
/ meta-tensor errors, coherent generations with healthy reward.

| bundle | rounds | result | reward |
|---|---|---|---|
| flux2_klein (9B full-FT) | 10/10 | ✅ pass | stable 0.75–0.78 |
| hunyuan_video15 | 10/10 | ✅ pass | 0.69 → 0.73 |
| qwen3 (SGLang) | 10/10 | ✅ pass | 0.12 → 0.17, ratio ≈ 1.0 |
| qwen_image (LoRA) | 10/10 | ✅ pass | stable 0.78–0.85 |
| hunyuan_video v1 (LoRA) | 10/10 | ✅ pass | stable 0.66–0.73 |
| **qwen_vl** (before fix) | 10/10 ran | ❌ real bug | reward 0 throughout, garbled generations |
| **qwen_vl** (after fix) | 12/12 | ✅ pass | 0.47 → 0.75, matches native-FSDP control |

### qwen_vl: real bug found and fixed (`b3df6c6`)

This was exactly the deferred→carrier bundle flagged as un-exercised — and it did not
survive contact. transformers 5.x renamed the VLM module tree
(`model.layers.*` → `model.language_model.layers.*`, vision under `model.visual.*`).
The meta-init path loads safetensors directly via `set_model_state_dict(strict=False)`
without `from_pretrained`'s key renaming, so **728/729 checkpoint keys silently failed
to match and were dropped** — LM + ViT kept their `to_empty()` garbage, generations
were pure noise, reward pinned at 0 with zero gradient and no error.

Fix in `sharded_load.py`:
- replay HF's own `WeightRenaming` rules on the checkpoint keys before broadcast;
  adopted only when it strictly increases key matches, so it is a no-op for
  checkpoints that already match (pure LLMs, diffusers transformers);
- a coverage guard that raises when the checkpoint matches <25% of model params,
  turning any future silent no-load into a hard failure at startup.

### Behavior preservation & zero side effects

- Same-seed per-round rewards with meta-init **on vs off** (stock `FSDPBackend`)
  essentially coincide for hunyuan_video15 / qwen_image / hunyuan_video / qwen3;
  fixed qwen_vl matches an upstream/main + native-FSDP control round-for-round.
  No second hidden loading bug.
- flux2_klein has no stock control: 9B full-FT OOMs on 8×H20 without VeOmni's
  sharded build (the meta-init path is what makes it fit); its meta-init reward is
  coherent and stable.
- The fixed loader ran qwen_image (diffusers + LoRA) and qwen3 (pure LLM) with **no
  "key-renaming applied" and no coverage errors** — zero side effects for models
  whose keys already match.

leviking98z-rgb (approved):

PR #208, opened 2026-07-14
fix(sglang): deterministic inference cloned every sample in a group, zeroing DRPO

CjhHa1:
## Summary

`engine_kwargs.rl_on_policy_target: fsdp` makes SGLang 0.5.12.post1 turn on
`enable_deterministic_inference` (`server_args.py:4300-4304`). Every request that
carries no `sampling_seed` then falls back to seed 42
(`sampling_batch_info.py:94-108`), and `n>1` is served by duplicating the request
object *including* its sampling params (`io_struct.py:391-505`). The two recipes
that set that flag — `examples/ar/qwen3_drpo_4b_base_dapo_sglang.yaml` and
`examples/ar/qwen3_ppo_4b_base_dapo_sglang.yaml` — therefore get 8 token-identical
completions for every prompt. Nothing rescues the diversity: candidate `rid`s
differ but `multinomial_with_seed` hashes only `(seed, positions, col_indices)`,
sibling positions are identical, and deterministic mode is batch-invariant.

- **DRPO is a training no-op.** It takes the `grpo` default for `advantage_mode`
  and centres advantages on the group mean, so an all-identical group has
  advantage exactly 0 on every token.
- **PPO still trains but loses exploration.** It sets `advantage_mode: gae`, so
  the gradient is non-zero, but it is computed over one trajectory per prompt
  replicated 8× and the critic sees the same return 8 times.
- **Eval is affected too.** `ARTrainer.evaluate` reuses the same engine with
  `eval_samples_per_prompt: 16`, so the reported `avg@16` is really `avg@1`.

The fix is the shape SGLang's own
`docs/advanced_features/deterministic_inference.md` prescribes for GRPO training
under "Generating Multiple Reproducible Responses": one request per distinct seed.
When deterministic inference is detected, `resolve_sampling` forces `n=1` and sets
a `base_seed` (42, or `sampling.seed` when given); the text and VLM adapters
pre-expand the prompt list `fanout`× and stamp each wire request with
`blake2b(base_seed, sample_id)`. Because `sample_id` carries the rollout id and
the fork index, the seeds are stable under DP sharding, request reordering and
out-of-order completion, and they advance every rollout.

`SGLangRolloutEngine.__init__` now logs a warning at boot when deterministic
inference is on, so the changed wire shape is visible in the run log, and the
SGLang-version constraint is recorded in `unirl/rollout/engine/README.md`.

Two unrelated one-line fixes ride along from this branch's earlier work; both are
described under Compatibility / Risk.

## Related Issue

N/A

## Test Plan

- Lint, exactly as CI runs it:
  `SKIP=no-commit-to-branch pre-commit run --all-files --show-diff-on-failure` —
  all 19 hooks pass, including `recipe _target_ paths resolve`,
  `docstrings are one line` and `core dependency directions hold`.
- The `tests/` tree was removed by policy (#99/#267); this PR adds no test files.
- **Live SGLang smoke: Not run; reason: no GPU available to the author of this
  revision.** This is the gap a reviewer should hold this PR to, and it is three
  measurements, not one. On one GPU with `Qwen/Qwen3-0.6B`:
  1. Raw SGLang A/B, no UniRL code — boot with `--attention-backend triton
     --rl-on-policy-target fsdp`, then compare one seedless `n=8` request against
     8 concurrent `n=1` requests carrying `blake2b(42, sample_id)` seeds, with
     `/flush_cache` between arms. Expect `1/8` distinct on the first and `8/8` on
     the second.
  2. Wall-clock for the same A/B, plus a third already-primed run of arm B. The
     `n>1` path primes the shared prefix with an awaited `max_new_tokens=0`
     request (`tokenizer_manager.py:1438-1449`); 8 concurrent `n=1` requests have
     no such barrier, so prefill may repeat until one of them inserts the prefix.
     That cost is unmeasured.
  3. One UniRL rollout on `main` and on this branch —
     `num_devices=1 batch_size=1 num_rollouts=1 eval_interval=0
     stack.num_updates_per_batch=1`, `max_new_tokens=256`, wandb off. Expect
     `gn≈0.0000` on `main` and non-zero here, with `k3` and `|Δlogp|` holding
     their order of magnitude across the two — the rollout↔replay parity check
     this recipe is tuned around.

## Compatibility / Risk

- **This is deliberately not opt-in.** `resolve_sampling` injects a seed for any
  engine whose `engine_kwargs` set `rl_on_policy_target` or
  `enable_deterministic_inference` — 2 of the 24 sglang recipes in `examples/`
  today, and both are the broken ones. Requiring an explicit `sampling.seed`
  instead would leave a silently zero-gradient run as the default for anyone who
  copies that flag, which the DRPO recipe documents as a correctness requirement.
  The boot warning is what keeps it from being silent.
- **The wire shape changes on those two recipes.** Per rank the DRPO recipe goes
  from 2 requests × `n=8` to 16 requests × `n=1` — the same 16 sequences, more
  requests. `concurrency: 16` absorbs that. See Test Plan item 2 for the prefill
  question, which is unmeasured.
- **`sampling.seed` without deterministic mode raises** rather than being silently
  ignored. Nothing in `examples/` sets it, so nothing changes today.
- No checkpoint, dataset-format or API migration. The other 22 sglang recipes and
  every non-sglang engine are untouched.
- `Part.fork` now rejects empty or duplicate `sample_ids`. Every in-tree caller
  already satisfies that; it is the one hunk outside the sglang seam and can be
  dropped if a reviewer prefers a tighter diff.
- `UniRLWandBLogger.log_rollout_step` now advances the train axis when telemetry
  is disabled. Its only consumer is the resume sidecar at
  `unirl/trainer/base.py:331`, so this affects only a run that trains with wandb
  off and resumes with it on.
- `fingerprint_tensor` reshapes before the `uint8` view, making 0-D multi-byte
  tensors hashable instead of raising from `Tensor.view(dtype)`. The input is
  already `.contiguous()`, so every existing fingerprint is byte-identical.

## Reviewer Notes

Start at `unirl/rollout/engine/sglang/utils/sampling.py`, then the two adapters,
then the boot warning in `unirl/rollout/engine/sglang/engine.py`.

Known gaps left for their own issues. None is reachable from a recipe in
`examples/` today, which is why none blocks this PR:

- `AgenticTrainer` (`unirl/trainer/agentic.py:192`) replicates one prompt
  `group_size` times with *identical* `sample_ids`, each in its own `generate`
  call — so under this scheme an agentic group would derive one seed and clone,
  exactly as the AR path does now. `validated_sample_ids` cannot catch it: it
  checks uniqueness within a call, and each call carries one row.
- `UnifiedModelTrainer._run_rollout_one` (`unirl/trainer/unified_model.py:262-273`)
  mints `r{rid}:a{k}` positionally within a DP shard, so two shards mint the same
  ids. Same underlying invariant as the item above.
- `SGLangEngineConfig.samples_pre_expanded` has had no reader anywhere in the tree
  since #214, and 9 recipes carry a comment describing behaviour no code
  implements.
- The pre-expansion assumes each fan-out group's siblings are contiguous in the
  gen part's `sample_ids`. That is what `Part.fork` produces and what
  `unique_group_indices` assumes, and nothing in-tree violates it, but nothing
  asserts it either — only a same-count permutation could slip past the existing
  count checks, and it would misalign seeds silently rather than raise.
- Deterministic VLM rollout base64-encodes the same image once per sibling, and
  deterministic text rollout applies the chat template once per sibling. Both are
  driver-side waste; the text half was measured at ~0.0023% of a step and the VLM
  half is unreachable until a VLM recipe enables the flag.

An earlier revision of this description claimed a `pytest tests -q` run of 25
tests and a merge of current `main` at `158ae59`. Neither happened — there is no
`tests/` tree at either ref, and this branch has no merge commits. Those claims
are removed, and it also claimed the production recipes stay on
`enable_deterministic_inference: false`, which is wrong: both set
`rl_on_policy_target`, which SGLang treats as the same switch.

AI assistance was used for the analysis, the patch and this description. I checked
the open PRs for overlapping sglang sampling work before revising this.

## Checklist

- [x] I validated code and config behavior; this PR adds no test files, per the
      #99/#267 policy.


examples/ar/qwen3_dppo_4b_base_dapo_sglang.yaml
```diff
@@ -143,6 +143,9 @@ rollout:
       disable_cuda_graph: false
       cuda_graph_max_bs: 16
       enable_lora: false
+      # Keep production sampling on SGLang's fast default backend. The
+      # profiling launcher overrides this to true together with a fixed seed.
+      enable_deterministic_inference: false
```
celve: I think we shouldn’t call this deterministic, since that usually implies batch‑invariant behavior.

celve (changes requested):

examples/ar/qwen3_dppo_4b_base_dapo_sglang.yaml
CjhHa1: done

celve (approved):

PR #209, opened 2026-07-15
feat(sft): SFT training domain — CE + flow-matching over the shared TrainStack (LLM/VLM/diffusion/unified)

haonan3:
## Summary

Adds the **SFT training domain** — supervised finetuning for every bundled model family (LLM / VLM cross-entropy, diffusion / unified-model flow-matching MSE) — as a sibling of the RL trainers that reuses the whole consumer side: the same `TrainStack`, `FSDPBackend`, bundles, and recipe schema. Design follows the cross-framework survey discussed with @haonan3 (verl's engine unification, NeMo-RL's loss-plugin shape, AReaL's `loss_weight_fn`, slime's producer-swap): **SFT = a swapped data producer + an anchor-free loss, never a mode of the RL loop.**

- **Losses as `StageAlgorithm` peers** (`unirl/algorithms/sft.py`): `SFT` (masked CE via the same `ARStage.replay` kernel GRPO uses, temperature pinned 1.0) and `FlowMatchSFT` (σ-sampled velocity MSE via `predict_noise_at_step`, DiffusionNFT's forward-noising recipe; `timestep_shift` defaults to the family's inference schedule). Both declare `requires_advantages=False`.
- **Normalization contract in `TrainStack`**: token-level CE is normalized by the **global** valid-token count of one optimizer step — across micro-batches AND DP ranks (`loss_scale = micro_tokens · dp_world / global_tokens`, SUM-all-reduced) — the invariant transformers/verl/OpenRLHF each shipped wrong for months. The weighting is grouping-invariant by construction, so the update gradient is independent of the micro split / DP layout. `sp_size>1` fails loudly (verl #5983 class). New `eval_track` runs the training loss forward-only (eval mode, full val set).
- **Supervised track builders** (`unirl/train/sft/source.py`): worker-side `DP_SCATTER` builders (`SupervisedTrackBuilder` base + AR / Diffusion variants, bound via the recipe's `track_builder` block) in the seat the rollout engine occupies — prompts via the bundles' own chat-template / `build_conditions` stages (byte-identical to what inference renders), targets via tokenizer+EOS (AR) or the new VAE encode stages (diffusion). Runtime asserts: non-empty supervision, token conservation (AReaL once trained 8× duplicated tokens with a correct-looking loss).
- **Supervised data layer** (`unirl/data/sft.py`): jsonl manifests with a first-class `response` field and `role="target"` media; exact `{epoch, position}` resume cursor checkpointed beside the model. RL's prompt-first data path is untouched.
- **Per-model surface = bundle stages only**: `SD3VAEEncodeStage` / `BagelVAEEncodeStage` (strict inverses of the decode stages, deterministic posterior) + `BagelPipeline.build_conditions`. qwen3/qwen_vl needed **zero** model-package changes.
- Recipes `examples/sft/{qwen3,qwen_vl,sd3,bagel}_sft.yaml` (same block shape as RL recipes; swap algorithm + track_builder + data) + 4 LoRA validation recipes under `examples/sft/validation/` + HF data prep tools (`prepare_sft_{text,vlm,t2i}`).

Relationship to #195/#200: implements the macro design settled in review discussion — same trainer/entry shape, but the loss lives in `unirl/algorithms` (not per-model task files), tracks ride the shared `TrainStack` (packing/EMA/grad-accum for free), and the per-model plug-in surface stays the bundle contract. #195's parity experiments remain the validation reference; those PRs can rebase onto this skeleton.

every changed line reviewed by the submitter. Duplicate-work check: #195/#200 discussed above.

## Related Issue

N/A (design thread: #195 review discussion)

## Test Plan

- **SFT-specific unit tests were dropped from this PR** (per review). The properties they covered — token-weighted global normalization / micro-split gradient invariance, FlowMatchSFT target math + seeded eval determinism, the `requires_advantages` gate, zero-token fail-fast, `eval_track` weighted mean, builder packing/EOS/truncation/pad-mask, and data-cursor exact resume — are enforced by construction and by runtime fail-fast guards, and are exercised end-to-end by the GPU runs below.
- `python scripts/check_recipe_targets.py` — all `_target_` paths resolve (incl. the 8 new SFT recipes: 4 full + 4 LoRA validation).
- `ruff check unirl/` + `ruff format --check` — clean.
- GPU e2e (1×8 H20, in progress — results will be posted on this PR): qwen3 alpaca SFT, qwen_vl llava-mix SFT, sd3 naruto-captions LoRA SFT, bagel T2I LoRA SFT; success = train loss falls, eval/loss falls, no NaN.

## Compatibility / Risk

- Purely additive at the entrypoint/recipe level; two shared-file changes: (1) `StageAlgorithm` gains two defaulted class attrs (`requires_advantages=True`, `loss_weighting="sample"`) — every existing algorithm keeps today's behavior; (2) `TrainStack._run_update` refactors its loss-scale computation behind `_resolve_loss_scales` — the `"sample"` path reproduces the existing arithmetic exactly.
- New optional flow: SFT checkpoints use the existing backend save/load + export tooling (SFT→RL handoff unchanged).


haonan3:
## GPU e2e results — all 4 model families pass (1×8 H20)

Suite: `ENTRY=train_sft bash examples/run_experiment_single_node.sh sft/<recipe>` per family, small-step smokes, full-set eval loss via `stack.eval_track` (fixed-σ for flow matching so the number is comparable across evals). **Success criteria met everywhere: train loss falls, eval/loss falls, no NaN, rc=0.**

| family | model | data | steps | eval/loss trajectory | rc |
|---|---|---|---|---|---|
| LLM (CE) | Qwen3-4B-Base (full FT) | alpaca-cleaned 7.8k | 30 | **0.9702 → 0.8898 → 0.8877 → 0.8834** | 0 |
| VLM (CE) | Qwen2.5-VL-3B (full FT, ViT frozen) | llava-instruct-mix 2k | 30 | **1.3869 → 1.2828 → 1.2832 → 1.2807** | 0 |
| diffusion (flow MSE) | SD3.5-medium (LoRA r32) | naruto-blip-captions 1.1k | 40 | **0.3336 → 0.3194** → 0.3203 → 0.3226 → 0.3228 | 0 |
| unified (flow MSE) | BAGEL-7B-MoT (LoRA r64, gen experts) | naruto-blip-captions | 20 | **0.3127 → 0.3093** → 0.3103 | 0 |

Train-loss samples: qwen3 `1.036 → 0.857`; qwen_vl noisy-down (short answers); sd3 `0.222 → 0.181` @ step15 (random-σ MC, oscillates as expected); bagel `0.291 → 0.194` @ step8.

Two bugs found and fixed by the runs themselves:
1. `eval_num_samples` wasn't declared in the recipes (Hydra strict mode) — added to all four.
2. **Bagel NCCL deadlock at smoke step 18**: `build_conditions` initially reused the memoized T2I context prefill (#145). Under the FSDP-wrapped transformer the prefill is collective, and a *per-rank* LRU hit (first recurring naruto caption) skipped one rank's all-gathers while peers ran them → 100%-util wedge (py-spy: all ranks pinned at the varlen `.item()` sync). Fixed by making `build_conditions` deliberately uncached — supervised batches have no GRPO siblings to keep hits rank-symmetric, and each prompt recurs once per epoch anyway. The GRPO path keeps the cache (sibling expansion makes hits symmetric). Commit `fix(bagel): uncached prefill in build_conditions`.

VLM note: this box's venv lacked a matching flash-attn wheel at suite start, so the qwen_vl run used the recipe's `sdpa` fallback path (padded replay); flash-attn finished compiling before the bagel run and its FA path is what bagel exercised.

Logs: `share_305110755/hunyuan/haonan/mmgrpo/_sft_smoke_209/logs/` (box `unirl-smoke-20260713-c0591a`, registry experiment `7fd929b1`, pass).

xshrz:
Clipboard_Screenshot_1784187276
Clipboard_Screenshot_1784187280
87658b38a68bf9fb5f7ea720da6ea2f3
fea2f1ef2505429b2c6e29fa7662563e

sd3.5:
Clipboard_Screenshot_1784187300

bagel:
Clipboard_Screenshot_1784187314

unirl/train/stack/base.py
```diff
@@ -317,6 +324,145 @@ def on_rollout_end(self) -> None:
...
+        weights = [self._micro_loss_weight(resp_track, start, end) for start, end in micros]
+        local_total = float(sum(weights))
+        (global_total,) = self._all_reduce_sums([local_total])
+        if global_total <= 0.0:
+            raise ValueError(
+                f"{type(self).__name__}: zero valid tokens in this optimizer step "
+                "(fully-masked batch?) — the data source must not emit steps with no "
+                "supervision (0/0 loss NaNs destroyed checkpoints in verl #785)."
+            )
+        dp_world = self._loss_weight_world()
```
CjhHa1, line 391: This is only True in FSDP situation, better get FSDP world

unirl/algorithms/sft.py
```diff
@@ -0,0 +1,412 @@
...
+
+        A seeded generator de-randomizes the noising, so the same validation
+        set yields a comparable number every eval (random-σ losses only
+        oscillate). ``segment.loss_mask`` (``[B]``) excludes padded eval rows.
+        """
+        x0 = self._clean_latents(segment)
+        if x0 is None:
+            return 0.0, 0.0
+        generator = torch.Generator(device=x0.device)
+        generator.manual_seed(self.eval_seed)
```
CjhHa1: The seed is per batch not sample level, is this right?

unirl/train/stack/base.py
```diff
@@ -251,21 +253,17 @@ def _run_update(
...
             result = self.algorithm.compute_loss_and_backward(
                 conditions=micro_track.conditions,
                 segment=micro_track.segment,
                 advantages=micro_track.advantages,
                 training_progress=training_progress,
-                loss_scale=loss_scale,
+                loss_scale=loss_scales[i],
             )
             micro_results.append(result)
             total_loss += result.loss
```
CjhHa1: FlowMatching SFT will be wrong

unirl/train/sft/source.py
```diff
@@ -0,0 +1,334 @@
...
+
+def _sample_ids(records: Sequence[Record]) -> List[str]:
+    return [str(r.get("sample_id", f"sft:{i}")) for i, r in enumerate(records)]
+
+
+def _pad_flags(records: Sequence[Record]) -> List[bool]:
+    return [bool(r.get("_eval_pad", False)) for r in records]
+
+
+class ARSupervisedSource(Remote):
```
CjhHa1: Better renamed to SupervisedTrackBuilder as this is not source

xshrz:
# review reply

## ① `unirl/train/stack/base.py` (loss_scale) — "FlowMatching SFT will be wrong"

The `loss_scales[i]` refactor doesn't change FlowMatchSFT's behavior — it's identical to the previous inline `(end - start) / update_total` for this algorithm. `FlowMatchSFT` declares `loss_weighting = "sample"`, so `_resolve_loss_scales` takes the sample branch and returns `[(end - start) / update_total]` per micro (the micro's share of the update's samples) — exactly what the old code computed.

What the refactor actually fixes is **AR SFT** with `loss_agg_mode="token-mean"` (`loss_weighting="token"`), which previously got sample-share weighting even though it needs the global-token denominator.

Math for FlowMatchSFT: each micro returns `per_sample.mean()` (a mean over its `n_i` samples); scaling by `n_i / N` and summing gives `(1/N) · Σ per-sample-losses` = the global sample-mean, and FSDP's DP-mean over equal `DP_SCATTER` shards makes it exact across ranks — so the accumulated gradient equals the full-batch sample-mean regardless of micro packing or DP layout. Happy to add a micro-invariance unit test if that would help.

---

## ② `unirl/algorithms/sft.py` (`evaluate_loss`) — "The seed is per batch not sample level, is this right?"

Good catch — fixed in the follow-up commit. The eval draw is now **per-sample**: `eval_track` threads `sample_ids` into `evaluate_loss`, and each sample seeds its `(σ, ε)` from `sha256(f"{eval_seed}:{sample_id}")` (a dedicated generator per sample), so the validation loss is invariant to eval batch size, batch order, and DP sharding. A single batch-level seed tied a sample's noising to its position in the batch, which is exactly the fragility you flagged. `SFT.evaluate_loss` takes the same `sample_ids` arg for a uniform signature but ignores it (CE is deterministic).

---

## ③ `unirl/train/sft/source.py` — "Better renamed to SupervisedTrackBuilder as this is not source"

`Source` is intentional: these classes are the SFT swap-in for the rollout engine as the *data producer*. In the RL loop the producer is the rollout engine (`TrainsideRolloutEngine`); SFT replaces it with a dataset-backed producer that emits the same `RolloutTrack` the stack consumes — a `Remote` sibling with one `DP_SCATTER` `build`, mirroring the engine's shape (see the module docstring). Naming it `...Source` keeps that RL/SFT symmetry explicit (it's the *source* of tracks, in place of the rollout source). `TrackBuilder` reads as a stateless helper and loses the "producer that mirrors the rollout engine" framing. I'd lean toward keeping `Source` for that parallel, but happy to revisit if it's not landing.

unirl/algorithms/sft.py
CjhHa1: Done

unirl/train/stack/base.py
CjhHa1: Done

unirl/train/sft/source.py
CjhHa1: Done

unirl/train/stack/base.py
CjhHa1: Done

CjhHa1 (approved):
LGTM

PR #210, opened 2026-07-15
feat(experimental): add WAN ReFL as a self-contained training package

YSunLIN:
# feat(experimental): add WAN ReFL as a self-contained training package

## Summary

This PR moves algorithm-specific training flows out of UniRL core and organizes them as self-contained **recipes**, landing `experimental/refl` — WAN ReFL/BPTT (direct differentiable-reward backprop) for Wan 2.1 T2V and Wan 2.2 I2V — as the first recipe. The recipe is built purely on existing core seams (`placement()` + `remote_hydra` + `_target_` late binding); it introduces **no new orchestration layer**.

- **`experimental/refl` package** (trainer + roles + models + reward + examples + README, self-contained):
  - `REFLTrainer` subclasses `unirl.trainer.base.BaseTrainer` directly and wires two colocated roles (actor + differentiable reward), mirroring `RewardBackpropTrainer` (the SD3 image-ReFL driver).
  - `ReflActorRole` mirrors `ReFLPolicy`'s family-agnostic contract: `pipeline_target` + `model_config` → `Pipeline.from_config`, with `FSDPBackend` composed in `initialize()`. Payloads are sample-native primitives (`Texts` / `Images` + per-sample metadata records) — no request objects.
  - Recipe-local `Wan21ReflPipeline` / `Wan22ReflPipeline` subclass the mainline pipelines and swap in a `diffuse_with_grad` BPTT stage (truncated `mid_timestep..final_timestep` grad window, single-branch CFG, optional per-step KL against the LoRA-disabled reference). The BPTT stage contract (`DiffuseWithGradResult`) is **recipe-local by design**: concrete stages inherit the `DiffusionStage` Protocol explicitly, so a protocol-level stub would become a real `None`-returning method on every stage and defeat `hasattr` capability checks; the promotion path is a separate opt-in protocol (the `DifferentiableReward` idiom) once a second out-of-recipe BPTT consumer exists. Mainline WAN diffusion/pipeline code is untouched.
  - Recipe-local rewards with isolated dependencies: VideoAlign (Qwen2-VL VQ/MQ/TA) and Face-identity scorers under `experimental/refl/reward/`, each with its own `requirements.txt` (core env stays clean).
  - Flat configs (same schema as `examples/`): `experimental/refl/examples/{wan21_t2v_videoalign_refl,wan22_i2v_face_refl}.yaml`.

- **Core enablers** (framework-level, shared with future BPTT recipes):
  - Differentiable path: cross-RPC grad-input handling fix in `Worker.call`; `RewardService.score_differentiable(media_tensor, prompts, records)` widened to video tensors + per-sample metadata; `DifferentiableReward` protocol updated to match.
  - Memory-optimized `WanVideoVAE` (nested grad checkpoint + activation-grad-only conv) **deliberately replaces the WAN decode implementation in core** — memory/numeric VAE optimizations are model assets shared by GRPO and ReFL, not algorithm property; forking the decode contract per algorithm was rejected in review. Hub repo-id checkpoints (the mainline `wan21_t2v.yaml` default) resolve through the HF cache (`snapshot_download`), preserving the previous `from_pretrained` loading semantics.
  - `LoraConfig.module_prefix` (Wan 2.2 trains only the `low_noise` DiT), `linear_warmup` LR schedule, `MultimodalRLDataSource` `run.shuffle`.

## Review adjustments (2026-07-30, maintainer push at `ec561c7a`)

Fast-forward on top of the author's head `e3c6b940` (nothing rewritten), applying the review consensus and re-aligning with main:

1. **Rebased onto the sample-native core**: upstream #214 removed `RolloutReq`/`RolloutInputs` after this PR's last merge; the recipe now consumes `Sample` primitives end-to-end.
2. **Dropped the roles DSL** (`recipes/common/`, −422 lines) per review: `REFLTrainer(BaseTrainer)` + `remote_hydra` replace the role-list orchestration; configs re-rooted from `roles:` to the repo-wide flat schema. Net −322 lines vs the previous head while keeping all functionality.
3. **P1 KL DP-aggregation fix**: `kl_loss` was a per-shard *scalar* `shared_field` — DP collect kept only rank 0's KL and re-broadcast it, which only lined up on the verified `batch == actor_dp == 8` topology (logs duplicated rank 0; other B/dp splits risked shape mismatch). Now a per-sample `[B]` concat column; `scripts/verify_refl_kl_batching.py` pins the wire-layer invariants across B==dp / B>dp / non-power-of-two / dp==1 / unequal actor-reward dp (standalone script per the #99/#267 no-unenforced-test-tree policy).
4. **Seed scheme**: kept your fixed-noise semantics (`params.seed` used verbatim, every rollout/rank — the regime your 835-rollout curve was trained in). An earlier adjustment decorrelated noise per rollout/rank; a checking pass flagged that as an unevidenced semantics change and reverted it (`fe2788c6`).
5. **I2V condition assembly** moved from the role into `Wan2xReflPipeline.build_refl_conditions` (mirrors each mainline `generate`); the role has no per-model imports.

**Expected behavior deltas vs the curves below**: per-shard-correct KL values (the Wan 2.2 `kl_weight=1.0` loss/KL curves may shift slightly — that is the correction) and the bf16 LoRA master (fp32 master trips a torch 2.7 FSDP2 uniform-dtype assert; a dedicated fp32 LoRA param-group fix is planned as a follow-up). Curve *shapes* are otherwise expected to match.

Second pass (`e0497399`), closing the remaining review findings:

6. **Dependency contract**: colocated reward+actor share one Python process, so a recipe `requirements.txt` can only *add* packages — it can never version-"isolate". The old file pinned `transformers==4.45.2` / `peft==0.10.0` / `flash-attn==2.5.8`, which would downgrade the core stack (`transformers>=5.6`, `peft>=0.14`) on install. VideoAlign now runs on the core stack: the wrapper's documented 5.x shim is enabled (explicit `mm_token_type_ids` black-list pop — never signature filtering, which PEFT breaks), flash-attn falls back to SDPA when absent, and `requirements.txt` is additive-only.
7. **Hub VAE loading**: `WanVideoVAE.load_from_diffusers` reads local files only; `WAN21Bundle` now resolves non-local paths through the HF cache first, so mainline repo-id configs keep loading.
8. **Recipe-local BPTT contract**: `DiffuseWithGradResult` + the `diffuse_with_grad` contract moved out of the core `DiffusionStage` Protocol (explicitly-inherited protocol stubs made `hasattr` capability checks vacuously true on every stage) into `experimental/refl/models/types.py`.
9. **Single KL knob**: `actor.kl_weight` owns on/off + weight and is injected into the stage's `sampler_kwargs`; a stale `sampling.sampler_kwargs.kl_weight` raises instead of being silently overridden.
10. Hygiene: BPTT window validated (`0 <= mid <= final < T`) at diffuse time; launch scripts run under `set -euo pipefail`; the Face reference-embedding cache is a bounded LRU (64).
11. **Layer + naming decisions (maintainer review)**: the top-level layer is `experimental/` — core carries the official training flows, and a layer named `recipes` read as the official usage (the word also already means the flat YAML here). `experimental/` names the incubation side of the intended two-way flow: mainstream packages get absorbed and solidified into core; internal/private packages live here uncommitted. Inside a package, **directories are named after the core home their content graduates into** (structural no-op promotion): `examples/` mirrors the top-level `examples/`, `reward/` mirrors `unirl/reward/`, `models/` mirrors `unirl/models/` (the BPTT pipeline/stage subclasses graduate by merging into the matching model packages). Launch `.sh` wrappers were deleted — the launch surface is one documented command, now in `experimental/refl/README.md` together with the layout map, environment policy, and the verification table.

## Test Plan

> 8xH20 with a Ray cluster started (`RAY_ADDRESS=auto`); replace `/path/to/...` with real paths.

```bash
export DATA_PATH=/path/to/prompts.txt PRETRAINED_MODEL=/path/to/Wan2.1-T2V-1.3B-Diffusers \
       VIDEOALIGN_MODEL_PATH=/path/to/VideoReward
RAY_ADDRESS=auto python -m experimental.refl.run --config-name=wan21_t2v_videoalign_refl num_devices=8
```

```bash
pip install -r experimental/refl/reward/face/requirements.txt
export PRETRAINED_MODEL=/path/to/Wan2.2-I2V-A14B-Diffusers DATA_PATH=/path/to/i2v_prompts.jsonl \
       FACE_MODEL_PATH=/path/to/antelodev2
RAY_ADDRESS=auto python -m experimental.refl.run --config-name=wan22_i2v_face_refl num_devices=8
```

Static validation at `e0497399` (CPU, 2026-07-30):

- compose-check both configs (`python -m experimental.refl.run --config-name=... --cfg job --resolve`) — rc=0 / rc=0;
- `scripts/check_recipe_targets.py` — 2270 `_target_` paths resolve (incl. `recipes.*`);
- worker-walker simulation instantiates every nested config dataclass and binds the `ReflActorRole` ctor for both configs;
- `python scripts/verify_refl_kl_batching.py` — all topology checks pass;
- full `pre-commit run --all-files` passes; `unirl/` imports nothing from `experimental/`.

Maintainer GPU smoke (2026-07-30, head `40b3f4c9`, 8xH20 fleet image `unirl-v1d`: torch 2.7.1 / transformers 4.56.0.dev0 / peft 0.17.1):

- `wan21_t2v_videoalign_refl` (`num_rollouts=2`, full 81f/480x832 geometry, local WAN ckpt + VideoReward-2B): **PASS**, clean exit — `rollout 1/2 reward=1.9839 loss=-0.4960 grad_norm=31.1 (230s)`, `rollout 2/2 reward=3.1245 grad_norm=25.3 (228s)`. Finite nonzero grad_norm = gradients flow reward → VAE → DiT LoRA across the RPC chain.
- Two portability fixes found by this smoke and pushed (`d96800dc`, `40b3f4c9`): `REFLTrainer` now stores `cfg` (BaseTrainer does not retain it), and the configs drop `master_dtype: fp32` — fp32 LoRA over the bf16 base trips torch 2.7 FSDP2's uniform-original-dtype assert (the documented `refl_sd3.yaml` note); LoRA master weights are now bf16, matching the SD3 refl recipe.

Still pending on GPU (owner: @YSunLIN):

- [ ] `wan22_i2v_face_refl` smoke at the current head (needs the face env extras + your I2V dataset with `ref_video_path` metadata);
- [ ] the ≥4.58/5.x `mm_token_type_ids` path of the VideoAlign shim (fleet image is 4.56 → the pop is a no-op there; needs a transformers>=4.58 env);
- [ ] longer-run curve-shape comparison vs your original curves (expected deltas: per-shard-correct KL, bf16 LoRA master);
- [ ] one mainline WAN sanity: a short official `wan21_t2v` trainside run (or Hub repo-id load + decode eyeball) covering the `WanVideoVAE` swap;
- [ ] provenance: upstream source + revision + license for `WanVideoVAE` and the VideoAlign port, per the repo `vendor/VENDOR_COMMIT.txt` convention.

## Verification Status

| Config Name | Model | Hardware | Status | Notes |
| --- | --- | --- | --- | --- |
| `wan21_t2v_videoalign_refl` | Wan2.1-T2V-1.3B + VideoAlign reward | 8xH20 | Verified (pre-adjustment head) | ReFL/BPTT; `num_rollouts=835`; `batch_size=8`; `num_inference_steps=25`; final-step differentiable path |
| `wan22_i2v_face_refl` | Wan2.2-I2V-A14B + Face reward | 8xH20 | Verified (pre-adjustment head) | ReFL/BPTT; `num_rollouts=690`; `batch_size=8`; `module_prefix=low_noise`; `num_inference_steps=8` |

### Wan 2.1 T2V ReFL training reward curve

Clipboard_Screenshot_1784087515

### Wan 2.2 I2V ReFL training reward curve

Clipboard_Screenshot_1784087567

## Compatibility / Risk

- `experimental/refl` is a new, self-contained path; existing entrypoints and the mainline WAN GRPO flow are unaffected. `experimental/` is repo-run only for now (not in wheel packaging) — a follow-up recipes-contract PR will settle packaging + import-direction CI.
- `RewardService.score_differentiable` signature changed (`media_tensor, prompts, records`); the two core ReFL call sites (`unirl/trainer/refl.py`) are updated in this PR. A follow-up PR will remove the legacy core ReFL path (`train_refl.py` / `trainer/refl.py` / `train/refl/`) so refl lives only in recipes.
- The WAN VAE decode flow is replaced by the memory-optimized `WanVideoVAE`; dtype/tiling/recompute may differ slightly from the old path numerically — monitored via reward curves and sample quality.
- `MultimodalRLDataSource.run.shuffle` defaults to `true` (previous behavior preserved); `module_prefix` is opt-in and backward compatible — confirm trainable params via startup logs.
- Recipe reward `requirements.txt` files are **additive-only by contract** (same-process colocation makes version isolation impossible); rewards that genuinely need a conflicting stack belong in the out-of-process `unirl-reward-service`, which is only an option for non-differentiable rewards.

AI-assisted: the 2026-07-30 adjustment was implemented and statically validated with Claude Code under maintainer direction (haonan3); every change reviewed against the review thread consensus. Duplicate-work check: supersedes the orchestration-layer portion only; overlaps were reconciled with #188's direction in the review discussion.


haonan3:
The tiled VAE decode path hardcodes the output buffer batch dimension to 1, while hidden_states may have B > 1. The decoded tile therefore has batch size B and cannot be accumulated in-place into values when B > 1.

Since decode() now defaults to tiled=True, this can break existing WAN recipes such as wan21_t2v.yaml, which uses forward_batch_size: 4. Please allocate these buffers using the actual input batch size, apply the same fix to tiled_parallel_decode, and add a regression test covering batch size greater than one.

celve:
I support moving WAN ReFL algorithm code into `recipes/`, but I don't think the new `Trainer + Role` abstraction is justified. `Remote`, `Handle`, `DevicePool`, `placement`, `remote_hydra`, and `BaseTrainer` already provide role lifecycle, placement, dispatch, and resource management.

Currently there is only one `Trainer` subclass and one `Role` subclass, while the proposal adds roughly 590 lines and a second orchestration API. The recipe still hardcodes `actor` and `reward`, so the `roles` list does not make the algorithm genuinely generic. There are also design concerns: `Role` assumes a fixed `bundle → pipeline → backend` lifecycle, and `RewardRole` bypasses `RewardService.__init__` while changing its method contract.

I suggest removing `unirl/trainer/trainer.py` and `recipes/common/roles.py`, having `REFLTrainer` inherit `BaseTrainer`, and making the recipe worker inherit `Remote` directly. Existing `placement`/`remote_hydra` APIs can build the topology, while `RewardService` can be extended for differentiable image/video scoring with metadata. The WAN ReFL functionality should remain; the generic role DSL can be reconsidered once multiple independent recipes demonstrate the need.

haonan3:
**Why now**: upstream `#214` retired `RolloutReq`/`RolloutInputs` minutes after your last main merge, so the branch stopped importing on current main. While re-aligning, we also applied the review consensus discussed with @celve, and the team direction for `recipes/` (extension layer over existing core seams; no second orchestration API).

**What changed** (functional payload — WAN pipelines, both rewards, `WanVideoVAE`, config semantics — is untouched):
1. Sample-native rebase: the actor role consumes `Texts`/`Images` primitives + per-sample metadata records; no request objects left.
2. `recipes/common/` roles DSL dropped (−422 lines): `REFLTrainer` now subclasses `BaseTrainer` and wires actor+reward with `placement()`+`remote_hydra`, mirroring `RewardBackpropTrainer`; `ReflActorRole` mirrors `ReFLPolicy`'s `pipeline_target`+`model_config` contract. Configs re-rooted to the flat schema (values preserved).
3. **P1 KL DP-aggregation fix**: `kl_loss` was a per-shard scalar `shared_field`, so DP collect kept only rank 0's KL — it only lined up on the batch==dp==8 topology. It is now a per-sample `[B]` concat column; `scripts/verify_refl_kl_batching.py` covers B==dp / B>dp / non-power-of-two / dp==1 / unequal actor-reward dp.
4. Seed now varies per rollout/dp-rank (matches `ReFLPolicy`); previously every rollout redrew the same initial noise.
5. I2V condition assembly moved into `Wan2xReflPipeline.build_refl_conditions`; the role has no per-model imports.

**Ask**: could you re-run the two 8xH20 smokes at this head? Expected deltas vs your curves: per-shard-correct KL values (wan22 `kl_weight=1.0` loss/KL curves may shift slightly — that is the correction) and per-rollout-varying init noise; curve *shapes* should match. Static validation at this head is recorded in the PR body. If anything in the adjustment looks wrong for your use case, flag it and we'll fix on the branch.

After merge, follow-ups (separate PRs): remove the legacy core ReFL path (`train_refl.py` / `trainer/refl.py` / `train/refl/`) so refl lives only in recipes, and a small recipes-contract PR (packaging + import-direction CI + docs).

haonan3:
Second maintainer pass at `e0497399` (fast-forward; responds to the external review findings). PR title updated to match the converged scope — this PR adds `recipes/refl`, it does not move official trainers.

**Fixed in this pass:**
1. **Dependency contract** — colocated reward+actor share one Python process, so recipe `requirements.txt` can only *add* packages, never re-pin the core stack (the old pins `transformers==4.45.2` / `peft==0.10.0` / `flash-attn==2.5.8` would have downgraded the core env on install). VideoAlign is forward-ported to the core stack: the wrapper's own documented 5.x shim is now enabled (explicit `mm_token_type_ids` pop — never signature filtering, which PEFT breaks) and flash-attn falls back to SDPA when absent.
2. **Hub VAE loading** — `WAN21Bundle` resolves non-local paths through the HF cache before `WanVideoVAE.load_from_diffusers`, so mainline repo-id configs (`Wan-AI/...` defaults) keep loading. The decode swap itself stays in core deliberately: VAE memory/numeric optimizations are model assets shared by GRPO and ReFL; forking the decode contract per algorithm was rejected in earlier review.
3. **Recipe-local BPTT contract** — `diffuse_with_grad` / `DiffuseWithGradResult` moved out of the core `DiffusionStage` Protocol into `recipes/refl/models/types.py`. Concrete stages inherit that Protocol explicitly, so the protocol-level stub was inherited as a real `None`-returning method on every stage, making `hasattr` capability checks vacuously true. Promotion path: a separate opt-in protocol (the `DifferentiableReward` idiom) once a second out-of-recipe BPTT consumer exists.
4. Single KL knob (`actor.kl_weight`; stale `sampler_kwargs.kl_weight` now raises), BPTT window validation, `set -euo pipefail` launch scripts, bounded Face ref-embedding LRU.

**updated pre-merge checklist (now in the PR body):** re-run both smokes on the **core stack** (transformers>=5.6; this also validates the VideoAlign forward-port), one mainline `wan21_t2v` sanity for the VAE swap (or Hub-load + decode eyeball), and upstream source/revision/license for `WanVideoVAE` + the VideoAlign port per the repo `vendor/VENDOR_COMMIT.txt` convention — you know the exact origins, we did not want to guess them.

haonan3:
**transformers 5.6 validation round at `811be0f9`** — the declared core stack (`>=5.6,<5.7`) is not what the fleet image ships (4.56.0.dev0), so we validated in an isolated `transformers==5.6.2 + peft==0.20.0` venv on 8xH20, with the real VideoReward checkpoint. Three 5.6 breaks found and fixed:

1. **Vision features** (`reward_model.py`): 5.6's vision tower returns `BaseModelOutputWithPooling` where `pooler_output` = merged features (what `get_video_features` reads) and `last_hidden_state` = PRE-merger states — the old preference for `last_hidden_state` would silently scatter wrong features into the placeholders. Now `pooler_output` wins; remaining fallbacks removed (pin-one-version policy), raw-Tensor branch kept only for the 4.56 fleet image with a removal TODO.
2. **Checkpoint loader** (`checkpoint.py`): the old→new Qwen2VL layout remap now also applies to the LoRA-split branch (LoRA keys embed module paths). Note: exercised statically only — no LoRA-split VideoAlign checkpoint was available; the full-`model.pth` branch is what the gate ran.
3. **`use_cache` forwarding** (`factory.py`): 5.x forwards unknown `from_pretrained` kwargs to the model ctor → `TypeError`; it is now set on the config after load (the fix the file's own disabled-shim note prescribed).

Gate evidence (isolated venv, GPU): VideoReward load → differentiable forward → backward — `rewards=[-3.6562, -3.3438]` (random-noise input), `grad_abs_mean=2.2e-3`, all finite. Env findings for the record: the fleet image carries transformers 4.56.0.dev0 / peft 0.17.1 (incompatible with 5.6) and lacks diffusers — the declared 5.6 stack currently exists in no image; and `peft>=0.14` admits combinations that cannot import against transformers 5.6. A lockfile-driven image rebuild proposal will follow separately.

haonan3 (approved):

PR #211, opened 2026-07-15
refactor(rollout): extract generic async runtime

CjhHa1:
## Summary

- Extract non-blocking Ray dispatch, in-flight scheduling, launch clamping, quiescing, and versioned group buffering from `AsyncARTrainer` into a trainer-agnostic `unirl.rollout.async_runtime` module.
- Buffer complete `RolloutResp` groups so future multi-track trainers can reuse the runtime, while preserving the existing single-track AR concat and canonical `ar` response path.
- Keep filtering opt-in at the buffer level and inactive in the scheduler; engines, entrypoints, recipes, and weight-sync semantics are unchanged.

## Related Issue

N/A. This is the behavior-preserving foundation for follow-up resident/admission and resumable-rollout work.

## Test Plan

- [x] `ruff check unirl/rollout/async_runtime.py unirl/trainer/async_ar.py`
- [x] `ruff format --check unirl/rollout/async_runtime.py unirl/trainer/async_ar.py`
- [x] Repository pre-commit hooks
- [x] GPU parity run for the Qwen3 async GRPO/DRPO recipes

## Compatibility / Risk

- `BaseRolloutEngine.generate`, all rollout engines, Hydra entrypoints, and existing async recipes are untouched.
- The load-bearing launch-clamp formula and `drain -> weight sync -> version increment` ordering are preserved.
- `RayGenerationDispatcher` intentionally mirrors the current private `Handle` dispatch/rebind methods; this coupling is documented in the implementation.
- Signal filtering remains disabled by default, so existing training-batch selection does not change.

## Checklist

- [x] I kept agentic session/workflow/environment logic out of the generic layer.


zzhuoxin1508:
## GRPO Async A/B Test — main vs #211

Verified that the async-runtime refactor in #211 is **behavior-preserving** (no regression) on the AR async path.

**Setup:** Qwen3-4B-Base, DAPO-math, GRPO, 1×8 H20 (4 train / 4 rollout, disaggregated async), `num_devices=8`. Two configs run as A/B (main vs #211), otherwise identical to the original recipe.

### Experiment 1 — on-policy (`max_inflight=1`, `weight_sync_interval=1`, `buffer_max_staleness=0`), 10 steps
| metric | main | #211 |
|---|---|---|
| reward (first→last) | 0.113 → 0.275 | 0.102 → 0.311 |
| ratio (mean) | 0.996–1.000 | 0.995–1.000 |
| clip | 0.00–0.01 | 0.00–0.02 |
| grad_norm | ~0.011 | ~0.012 |
| lag | 0 | 0 |

### Experiment 2 — buffered async (`max_inflight=2`, `weight_sync_interval=4`, `buffer_max_staleness=2`), 12 steps
| metric | main | #211 |
|---|---|---|
| reward (first→last) | 0.117 → 0.178 | 0.113 → 0.199 |
| ratio (mean) | 0.982–1.002 | 0.978–1.007 |
| clip | 0.00–0.05 | 0.01–0.06 |
| grad_norm | ~0.010–0.014 | ~0.010–0.026 |
| per-rollout time | ~2.0 min | ~2.0 min |

### Verification (buffered arm)
- **launch ceiling** — bounded by `max_inflight` and the staleness window in `next_batch`; ~2.0 min/rollout (vs ~3.5 min on-policy) from generation/train overlap; no buffer underflow.
- **freshness drain** — `drain_freshest` evicts stale groups then takes the newest; ratio returns to ≈1 right after each sync.
- **cross-sync buffer** — `VersionedGroupBuffer` carries groups across syncs (up to `staleness=2`); no underflow.
- **lag ≤ 2** — eviction rule `current_version - weight_version ≤ 2`; ratio stays bounded (0.978–1.007), never diverges.
- **drain → sync → version++ ordering** — structurally guaranteed in the train loop; both arms completed all steps (syncs at step 4/8/12) with no in-flight/weight-sync corruption.

### Result
Both on-policy and buffered A/B runs show **numerically consistent behavior between main and #211** (differences within seed noise): bounded ratio, finite loss/grad, reward increasing, no dropped/repeated groups, and both arms finished cleanly. **No regression — safe to merge.**

> Note: #211 only affects the AR async path (`async_ar.py` + new `async_runtime.py`); the diffusion (colocate) path is unaffected and was not part of this test.

unirl/rollout/async_runtime.py
```diff
@@ -0,0 +1,356 @@
...
+                ready.append(job)
+            else:
+                still.append(job)
+        self._inflight = still
+
+        first_error: Optional[BaseException] = None
+        for job in ready:
+            try:
+                self._complete(job, on_complete)
+            except BaseException as exc:
```
zzhuoxin1508: When two ready jobs both fail,
   only `first_error` is re-raised — the second exception hits the `else`
   branch and vanishes with no log line.

unirl/rollout/async_runtime.py
```diff
@@ -0,0 +1,356 @@
...
+                gen_id=job.gen_id,
+            )
+
+    def reap_ready(self, on_complete: CompleteGeneration) -> None:
+        """Collect every ready generation; leave unresolved jobs in flight."""
+
+        ready: List[InflightGeneration] = []
+        still: List[InflightGeneration] = []
+        for job in self._inflight:
+            if self._dispatcher.is_ready(job):
```
zzhuoxin1508: Since ready jobs are removed from `_inflight` before `_complete`, a job whose scoring fails here is permanently lost .The old inline `_reap_ready` left `_inflight` untouched on error, so the `finally`-path `_drain_all` could still re-collect and re-score it.Intentional trade-off?

CjhHa1: done

unirl/rollout/async_runtime.py
CjhHa1: done

unirl/rollout/async_runtime.py
```diff
@@ -0,0 +1,369 @@
...
+        extraction does not change existing AR batch selection.
+
+        Returns ``None`` without consuming eligible groups when fewer than ``n``
+        remain after eviction/filtering.
+        """
+
+        if max_staleness is not None and current_version is not None:
+            self._items = [item for item in self._items if current_version - item.weight_version <= max_staleness]
+        if has_signal is not None:
+            self._items = [item for item in self._items if has_signal(item.resp)]
```
zzhuoxin1508: `has_signal` has no caller anywhere in the repo (the only call site, `next_batch`, never passes it), so this is untested speculative API. Its semantics also contradict the docstring: groups failing the predicate are permanently deleted from `self._items` — even when the call returns `None` because fewer than `n` remain — while the docstring promises "Returns None without consuming eligible groups". A future caller passing a transiently-false predicate (e.g. "reward not attached yet") would silently lose completed rollouts on the first drain.

unirl/rollout/async_runtime.py
```diff
@@ -0,0 +1,369 @@
...
+    @property
+    def inflight_count(self) -> int:
+        return len(self._inflight)
+
+    @property
+    def launch_id(self) -> int:
+        return self._launch_id
+
+    @property
+    def buffer_size(self) -> int:
```
zzhuoxin1508: These three properties have no readers anywhere on the branch — they were only used by the unit tests that were removed from this PR. As shipped they're dead public surface: they suggest external observers exist, so future refactors of `_inflight`/`_launch_id`/`_buffer` have to preserve accessors nobody uses.

unirl/rollout/async_runtime.py
CjhHa1: Removed in `ed6a200` — dropped the unused `has_signal` parameter (and its destructive in-place filter) so `drain_freshest` only does staleness eviction + freshest-n selection.

unirl/rollout/async_runtime.py
CjhHa1: Removed in `ed6a200` — deleted `inflight_count` / `launch_id` / `buffer_size` (and the now-unused `VersionedGroupBuffer.size`) so they are no longer dead public surface.

zzhuoxin1508 (approved):
LGTM

PR #213, opened 2026-07-16
perf(async): refill AR rollouts before scoring

CjhHa1:
## Summary

- Refill the next admissible AsyncAR rollout immediately after collecting the current batch and before reward scoring/training, extending generation overlap without adding another controller, queue, or runtime mode.
- Count the trainer-leased batch against the publication freshness window so early refill cannot cross weight-sync, eval, checkpoint, or final boundaries.
- Keep AsyncDiffusion's score-before-refill ordering through a trainer hook; its cross-slab transfer invariant is unchanged.
- The branch includes current `main`, where [#358](https://github.com/Tencent-Hunyuan/UniRL/pull/358) fixes the missing AsyncAR input-primitive preflight through the shared model-blind `ar_preflight()` path. The final merge therefore contains both fixes without duplicating that logic in this PR diff.

The partial-localization transport optimization previously on this branch was split into [#354](https://github.com/Tencent-Hunyuan/UniRL/pull/354).

## Related Issue

Related to the rollout/training overlap goal in [#93](https://github.com/Tencent-Hunyuan/UniRL/issues/93). This PR deliberately reuses the existing `RolloutManager` / `RolloutPool` control path rather than adding a second controller.

## Design

- `RolloutPool` remains the only background dispatcher; `RolloutManager` remains the publication and grouping authority.
- AsyncAR collects one completed generation, accounts for that leased batch in admission, submits the next allowed generation, then scores and trains the collected batch.
- `max_inflight` remains the dispatch concurrency bound. Publication freshness and hard-boundary math remain the outer admission constraints.
- AsyncDiffusion uses the shared loop's safe default ordering: score and complete cross-slab transfer before launching replacement work.
- No `dual` mode or `max_pending_generations` configuration remains.
- AsyncAR prompt-input validation comes from main's shared `ar_preflight()` implementation (#358), which is called explicitly because `AsyncARTrainer` skips `ARTrainer.__init__`.

## Test Plan

- Local-only focused contract suite — 11 passed. Test files are intentionally retained outside the PR diff.
- GRPO and DRPO async Hydra composition — passed.
- Ruff and full pre-push hooks — passed.
- GPU workload matrix on the isolated 8×H20 worker of `debug-16gpu-0807`:
  - GRPO 128-token ABBA: extra pending was 1.85% slower.
  - GRPO 1024-token ABBA: extra pending improved the mean by 3.04%, below the 5% retention threshold; median and rollout idle did not improve.
[...]

PR #212, opened 2026-07-15
feat(rollout): add bounded async prefetch residency

CjhHa1:
## Summary

- **#211 is now in `main`.** Current head `d06898b` merges the latest `main` and resolves the async-runtime conflicts.
- Add `prefetch_batches` as a bounded request/group-residency axis independent from `max_inflight` concurrency and the existing policy-version launch clamp.
- Reserve capacity before consuming data, require the fixed AR contract of one generation per training batch, and consume bounded resident work FIFO so paid-for groups cannot starve into stale eviction.
- Keep checkpoint/resume deterministic with hard launch ceilings, empty-runtime assertions, and save/weight-sync alignment validation.
- Add per-step runtime depth, admission/reap/age/eviction metrics plus train, generation-wait, quiesce-wait, and weight-sync timings.
- Preserve `main`'s complete-or-nothing buffer insertion, failed-job retry bookkeeping, and teardown exception safety.
- Enable one successor batch in the Qwen3 GRPO/DRPO async recipes; with the default `weight_sync_interval=1`, the existing sync-window clamp still prevents prefetch.

## Related Issue

Built on the async runtime introduced by #211, which is now in `main`. This is request/group-level residency only; token/KV resumability remains a separate follow-up.

## Test Plan

- [x] Original async-residency regression suite — 20 passed before removal; all 20 cases re-executed in memory on current head
- [x] Current repository suite — 25 passed
- [x] Retry/atomic-insertion failure paths — inline validation passed
- [x] GRPO and DRPO Hydra configs compose and fully resolve
- [x] Pre-commit hooks on all PR-owned files
- [x] Final diff against `main` contains no test-file changes
- [ ] GPU parity: `prefetch_batches=0`, `max_inflight=1`, `buffer_max_staleness=0`
- [ ] GPU overlap: `prefetch_batches=1`, `max_inflight=1`, `weight_sync_interval=2`

## Compatibility / Risk

- Python/API default `prefetch_batches=None` preserves #211's legacy admission behavior; bounded mode is opt-in outside the updated example recipes.
- This does **not** claim fully-async token generation: every AR request is completed under one policy version, and the mandatory drain-before-weight-sync barrier remains.
- Bounded mode deliberately uses FIFO residency; legacy mode keeps freshness-first selection.
- Checkpointing with a separate rollout engine now requires actual periodic/final save steps to align with `weight_sync_interval`, avoiding resumed-vs-uninterrupted rollout-weight divergence.
[...]

PR #197, opened 2026-07-09
feat(agentic): fully-async resident pool  on top of half-async

CjhHa1:
## Summary

- Stacked on [#121](https://github.com/Tencent-Hunyuan/UniRL/pull/121) (`feat/agentic-half-async`): colocate **fully-async (v5) resident pool** for multi-turn think→gen→obs Bagel agentic RL.
- Sessions carry across `generate()` calls with policy-version staleness (`max_staleness`); trainer bumps version after each optimizer step.
- Packs per-token / per-SDE-step version metadata (`TextSegment.token_versions`, `LatentSegment.sde_versions`) for audit/backpressure only — GRPO / FlowGRPO unchanged.
- Additive surface: `FullyAsyncPool` + `version.py`, `BagelAgenticEngineFullyAsync`, `FullyAsyncAgenticTrainer`, entry `train_agentic_fully_async.py`, recipe `examples/agentic/bagel_thinkgen_fully_async.yaml`, CPU tests.

**Out of scope:** disaggregated (`*_disagg*`) path is intentionally not included.

## Related Issue

Depends on #121.

## Test Plan

- [x] CPU: `pytest tests/agentic/test_fully_async.py tests/agentic/test_engine_fully_async.py`
- [x] Config compose: `python -m unirl.train_agentic_fully_async --config-name=agentic/bagel_thinkgen_fully_async --cfg job --resolve`
- [ ] GPU E2E (colocate): recipe `agentic/bagel_thinkgen_fully_async` with `max_staleness>=1` to observe carry-over / version spread metrics

## Compatibility / Risk

- Purely additive on top of #121; no changes to existing half-async trainer/engine behavior beyond optional segment metadata fields (default `None`).
- Keep `batch_size` divisible by rollout DP size (pool harvests whole groups per shard).

## Checklist

- [x] I reviewed the changed code and removed unrelated/generated artifacts.
- [x] I updated tests, docs, and configs where needed, or explained why not.

### Maintainer validation (3991583)

Merged the updated #121 base and removed both test files added by this stacked PR; the final diff against its base has no `tests/**` changes. The 11 original fully-async CPU cases passed before removal. The recipe composes/resolves, and `compileall` plus full pre-commit pass. This still cannot merge independently: it depends on draft #121, has no GPU E2E, and both are superseded architecturally by the Sample-native work in #214.

PR #121, opened 2026-06-24
feat(agentic): half-async multi-turn think→gen→obs RL for Bagel

CjhHa1:
## Summary

New `unirl/agentic/` package — a general-purpose, **half-async multi-turn agentic RL** layer (think→gen→obs), validated end-to-end on **Bagel-7B-MoT**. Built on the proven `AsyncARTrainer` half-async model and the multi-track `RolloutResp` lineage.

Backend-agnostic core + Bagel integration (zero changes under `unirl/models/bagel/`):
- `session.py` — `Session`/`MsgNode` linear delta chain; the "obs not in segment" invariant (obs text is prefix-only context for the next turn, never a training segment row).
- `env.py` — `AgenticEnv` protocol + `RuleEnv` (deterministic, model-free).
- `workflow.py` — `AgenticWorkflow` + `ThinkGenWorkflow` + a model-agnostic `AgenticBackend` seam.
- `buffer.py` — `AgenticBuffer` (multi-track groups, freshness/staleness eviction, slime-style zero-variance dynamic filter).
- `engine.py` — `BagelAgenticEngine` (trainside; generalizes Bagel's single-turn `t2ti` to a `T`-turn think→gen loop; returns the 2-track `{ar, image}` resp).
- `trainer.py` — `AsyncAgenticTrainer` (colocate; GRPO(think) + FlowGRPO(image) → one `UnifiedModelTrainStack` step; over-sampling + dynamic filter, with a no-buffer fast path).
- Entry `train_agentic.py` + recipe `examples/agentic/bagel_thinkgen_rulenv.yaml`.

## Related Issue

N/A

## Test Plan

- **CPU unit/integration (29 passing):** `pytest tests/agentic/` — session build/`obs-not-in-segment`, RuleEnv, ThinkGenWorkflow over a fake backend, AgenticBuffer (freshness/staleness/variance filter), full `BagelAgenticEngine.generate → 2-track resp` via a fake pipeline, and `propagate_rewards` + per-prompt/root GRPO grouping.
- **Config compose:** `python -m unirl.train_agentic --config-name=agentic/bagel_thinkgen_rulenv --cfg job --resolve`
[...]

PR #214, opened 2026-07-16
feat(agentic): add Sample-native multi-turn rollout and training

celve:
## Summary

This draft proposes a Sample-native agentic RL integration for UniRL.

- Replace the rollout request/response boundary with the lineage-preserving `Sample -> Sample` / `Sample -> list[Sample]` contract and migrate existing model, rollout, reward, and training paths onto `Sample` / `Part`.
- Add synchronous multi-turn agent execution with role-aware conditioning, tool/environment observations, a rank-zero rollout coordinator, and thread-safe SGLang inner engines.
- Add barrier, colocated partial-rollout, and disaggregated asynchronous agentic trainers, with ALFWorld and calculator/deep-research workflows.
- Add migration and agent-loop documentation plus runnable example recipes.

The draft is intentionally opened for architecture and sequencing review before the remaining final-HEAD GPU matrix and patch-size cleanup are complete.

## Test Plan

Completed on the current draft tip (`159627bd`):

- `python3 -m compileall -q unirl`
- `python3 scripts/check_recipe_targets.py` — 1,965 Hydra `_target_` paths resolved
- Hydra API composition plus `OmegaConf.resolve` for all six barrier/partial/async ALFWorld and calculator/deep-research recipes

Real-GPU evidence from integration ancestors:

- Qwen3-8B agentic rollout on two H20 workers: multi-turn calculator execution, ragged trajectories, cross-worker aggregation, and driver hydration passed (`b4ed9375`).
- Qwen3-8B ALFWorld GRPO: fixed eight-game set, up to ten environment turns, reward approximately 0.53 -> 0.85 with a 0.94 peak (`9cd7f3ab`).
- Qwen3-1.7B deep research with a Qwen2.5-72B judge: batch 128, eight siblings per prompt, reward approximately 0.22 -> 0.50 over 19 rollouts before an infrastructure failure (`ce13a48d`).
- Stable Diffusion 3 was exercised end-to-end through all three rollout paths: [trainside](https://wandb.ai/linyuwus/unirl-lin560-e2e/runs/ceqzx0zv), [vLLM-Omni](https://wandb.ai/linyuwus/unirl-lin560-e2e/runs/so1kvhvp), and [SGLang diffusion](https://wandb.ai/linyuwus/unirl-lin560-e2e/runs/ut6xvkim).
- The original autoregressive (AR) recipe completed multi-rollout training and produced an increasing reward trend ([W&B](https://wandb.ai/linyuwus/unirl-lin560-pe-ar-e2e/runs/ig1mu5ny)).
- The original Prompt Enhancement (PE) recipe completed multi-rollout joint training and produced an increasing reward trend ([W&B](https://wandb.ai/linyuwus/unirl-lin560-pe-ar-e2e/runs/qjrmmriu)).
- FLUX.2 Klein 4B image editing (IT2I) passed on the current draft tip (`159627bd`) on 8× H20 with a local PickScore smoke override: 8 source-image prompts × 2 edits, the three-Part text → source image → generated image lineage and typed `image_latent`/`image_latent_ids` replay conditions passed on every rank, and one optimizer step completed with `train/has_backward=1` and nonzero grad norm ([W&B](https://wandb.ai/linyuwus/unirl-lin560-it2i-e2e/runs/n11ypp12)).
- The original checked-in BAGEL + EditReward IT2I recipe (`diffusion/bagel/bagel_editreward`) completed on 8× H20 with EditReward isolated on a second pod: the default step-0 deterministic evaluation ran, followed by 3 full rollouts at the default 8 prompts × 8 edits (192 training edits total), 14 denoising steps, and 2 optimizer updates per rollout (6 total). All 32 recipe scoring batches, plus the separate two-turn preflight, returned HTTP 200; rollout rewards were `-0.7940`, `-0.9308`, and `-0.7629`; and eight 1024×512 source|edited previews synced successfully ([W&B](https://wandb.ai/linyuwus/unirl-lin560-it2i-original/runs/jujmld3s)). This validation used two working-tree fixes not yet committed to the PR: hydrating the source-image `TensorRef` before media-preview indexing and treating EditReward's blank `output_dir` as its `/tmp` inference default.

Pending before ready-for-review:

- Re-run the agentic barrier, partial-rollout, and asynchronous recipes on the final draft tip after the sync-only engine refactor and latest-main merge.
- Re-run a final-tip AR job covering the merged unrestricted `top_k` behavior.
- Re-run the original BAGEL IT2I recipe on the committed current-head fixes so that the successful hardware result is final-tip evidence.
- Record exact supported Python/CUDA/dependency environments, commands, datasets, hardware, and logs for the final GPU matrix.

## Compatibility / Risk

- Breaking public API change: `RolloutReq`, `RolloutResp`, and `RolloutTrack` are removed without compatibility aliases. See `unirl/types/README.md` for field/helper mappings.
- The patch spans the Sample/Part foundation plus every existing rollout consumer; choosing either side of the migration piecemeal would leave incompatible engine/model/trainer contracts.
- SGLang and vLLM-Omni require separate pinned CUDA environments. The final clean-install matrix is pending.
- Agentic evaluation is not implemented. Barrier/partial variants require evaluation disabled; asynchronous variants currently force it off.
- Partial carry is supported only for resumable/stateless environments. Stateful ALFWorld episodes and stateful tool sessions use the drop policy.
- Checkpoints restore model/optimizer state, not in-flight rollout buffers, carried environment episodes, or tool sessions.

## Reviewer Notes

Suggested review order:

1. `unirl/types/` Sample/Part lineage and migration contract.
2. Existing model, rollout, reward, and trainer migrations.
3. SGLang concurrency plus `unirl/rollout/loop/` and tools.
4. Agentic barrier/partial/async trainers, environments, and recipes.

Overlap/sequence needs maintainer agreement:

- #121 and #197 implement agentic paths on the retired `RolloutResp` architecture; this draft proposes the Sample/Part direction instead.
- #211, #212, and #213 contain overlapping async-runtime work that would need sequencing or a Sample-native port.
- #203 overlaps BAGEL vLLM-Omni grouping paths.

Temporary integration-only smoke scripts and the top-level test tree used during development were intentionally removed from the upstream patch; the retained validation evidence is summarized above. Before ready-for-review, we will complete the final-tip GPU matrix and perform a subsystem-by-subsystem human review. This integration was developed with substantial AI assistance (Codex); the human submitter will review and defend the final diff before the draft is promoted.

## Checklist

### Current-head maintainer validation (28a27e4)

- Merged current `main` (`fce4f4b`) and resolved the Sample/Part migration conflicts; GitHub now reports the PR mergeable.
- Committed the two fixes used by the successful BAGEL/EditReward run: hydrate the source-image transport reference before preview indexing, and use `/tmp/editreward_output` when the vendored config leaves `output_dir` blank.
- Ported current-main Qwen3 agent-SFT message/tool support and prompt-content eval seeding to the Sample-native API. Explicit eval noise keys are a CONCAT field, so DP split/select preserves row alignment.
- Removed all nine test files introduced by this PR. The final diff against `main` contains no `tests/**` changes.
- Validation: `pytest tests -q` (25 passed); 32 deleted historical PR regression cases executed in memory (all passed, no files restored); `compileall`; full pre-commit; all seven new agentic recipes and four modified diffusion recipes composed and resolved.
- Plain repository-wide `pytest` also runs the same 25 tests successfully, then mis-collects six existing command-line functions in `unirl-reward-service/scripts/test_videoalign.py` because they require CLI arguments; that script is unchanged relative to `main`.

This removes the textual merge conflict and known uncommitted BAGEL fixes, but does not make the patch low risk. Final-tip agentic/AR/BAGEL GPU reruns, dependency-matrix evidence, patch-size cleanup, and subsystem human review remain required before merge.


celve:
image

celve:
DRPO: 
image

celve:
PE trainside: 
image

unirl/trainer/agentic_async.py
```diff
@@ -0,0 +1,463 @@
...
+        finite = torch.isfinite(rewards)
+        mean_reward = float(rewards[finite].mean().item()) if bool(finite.any()) else 0.0
+        advantages = self._group_advantages(rewards, group_ids)
+
+        train_parts: List[Part] = []
+        for i, tr in enumerate(trajs):
+            adv_i = float(advantages[i].item())
+            for gp in tr.gen_parts():
+                gp = _part_with_field(gp, "advantages", torch.full((gp.batch_size,), adv_i, dtype=torch.float32))
+                gp = _part_with_field(gp, "primitive", None)
```
CjhHa1: Blocking: `Part` has a `primitives` field, not `primitive`. `_part_with_field` forwards this key into `Part(**kwargs)`, so the first non-empty async batch raises `TypeError: Part.__init__() got an unexpected keyword argument 'primitive'` before `train_track`. This should mirror `AgenticTrainer` (`"primitives", {}`), with a test that executes `_train_on_groups`.

unirl/rollout/engine/agentic/engine.py
```diff
@@ -0,0 +1,506 @@
...
+            return self._attach_env_reward(sample, env_reward), True
+        finally:
+            # Guaranteed teardown (LIN-533): end any open tool sessions / episodes for
+            # this trajectory — on success, crash, AND abort. Duck-typed like
+            # ``tool_schemas`` so envs without ``close`` are unaffected, and wrapped so
+            # a teardown error can never re-raise (``_run_one`` must not raise).
+            close = getattr(self._env, "close", None)
+            if close is not None:
+                try:
+                    close(sample)
```
CjhHa1, line 428: This also closes the environment when `_run_one` returns `(sample, False)` for a checkpointed/carry trajectory. On resubmit, `AlfworldEnv.reset` and stateful `ToolEnvironment.reset` create a new episode/session and return a root-only `Sample`, so the carried turn history is lost (async ALFWorld restarts at every weight sync). Checkpointing needs to avoid teardown for resumable state, or stateful tails must be dropped rather than carried.

unirl/trainer/agentic_async.py
```diff
@@ -0,0 +1,463 @@
...
+                return picked
+            if self.rollout.drained()[0]:
+                refills += 1
+                if refills > self._MAX_REFILLS:
+                    raise RuntimeError(
+                        f"async-agentic rollout {rollout_id}: buffer underflow after {refills} refills "
+                        f"(buffer={self._buffer.size()} < batch={self.batch_size}); raise "
+                        f"oversample_batch_size or buffer_max_staleness."
+                    )
+                self._submit_drive(self._pending_carried, rollout_id)
```
CjhHa1: There is a completion-loss race here: `_pump()` may poll just before the final worker appends to `_completed`; `drained()` can then become true, and this new `submit()` calls `reset_round()`, clearing that unpolled completion. The same pattern exists in the partial trainer. Please make join+drain atomic, or perform a final poll after observing `drained` and before resetting worker buffers.

unirl/types/sample.py
```diff
@@ -0,0 +1,797 @@
...
+        Input Parts (the prompt head and any :meth:`Part.input_child`) have
+        ``sampling_params is None`` and are skipped."""
+        return [p for p in self.parts if p.is_gen]
+
+    def gen_part(self, params_type: type) -> Part:
+        """The first gen Part whose ``sampling_params`` is an instance of
+        ``params_type`` (e.g. ``ARSamplingParams`` / ``DiffusionSamplingParams``)
+        — locating a stage by TYPE, not position (the migration's convention;
+        mirrors the engine-side ``ar_gen_part`` / ``diffusion_gen_part`` readers)."""
+        for p in self.parts:
```
CjhHa1: Selecting the first matching sampling-param type is ambiguous once a trajectory has repeated stages, e.g. `[input, ar1, observation, ar2]`. A supported agentic inner such as vLLM-Omni uses type-based lookup for write-back, so turn 2 can select/overwrite `ar1` and leave the frontier shell empty. This API should either reject multiple matches or expose an explicit frontier/last-occurrence lookup for generation paths.

unirl/trainer/agentic_async.py
```diff
@@ -0,0 +1,578 @@
...
+                return picked
+
+            refills += 1
+            if refills > self._MAX_REFILLS:
+                raise RuntimeError(
+                    f"async-agentic rollout {rollout_id}: buffer underflow after {refills} refills "
+                    f"(buffer={self._buffer.size()} < batch={self.batch_size}); raise "
+                    f"oversample_batch_size or buffer_max_staleness."
+                )
+            self._submit_drive(self._pending_carried, rollout_id)
```
CjhHa1, line 434: A refill reuses the same `rollout_id`, so fresh samples receive the same `r{rollout_id}:<source-id>` root namespace as the drained drive. This is reachable with `DefaultDataSource`, whose IDs restart at `prompt:0:sample:0` on every `get_samples` call: a pending sibling from prompt A can be combined with a newly generated sibling for prompt B into one GRPO group (I reproduced this locally). Please include a monotonically unique drive/refill nonce in fresh root IDs; `_GroupAssembler` should never see the same root identify different source rows.

unirl/models/hunyuan_image3/modes/t2ti.py
```diff
@@ -99,40 +98,49 @@ def _cot_stop_tokens(bundle, bot_task: str) -> List[int]:
...
+        )
 
-    ar_sp = req.sampling_params.get("ar")
-    require(
-        ar_sp is not None,
-        "HunyuanImage3Pipeline.generate (t2ti): AR sampling params missing — t2ti needs "
-        "a sampling dict with both 'ar' (ARSamplingParams) and 'diffusion' (DiffusionSamplingParams) entries.",
-    )
-    diff_sp = req.sampling_params.get("diffusion")
+    conditioning = sample.conditioning()
```
CjhHa1: This conditioning is aligned to the final diffusion frontier, not the AR Part. For a normal hierarchical request `P=1, N=2, M=3`, `sample.conditioning()` returns 6 prompt rows while `ar_part` has only 2 IDs; AR generation/fill therefore produces a deterministic 6-vs-2 batch mismatch when `M > 1`. Build AR conditioning from `sample.parts[: ar_idx + 1]`, fill the AR Part, and only then expand conditioning for the diffusion frontier.

unirl/rollout/engine/agentic/engine.py
```diff
@@ -0,0 +1,539 @@
...
+                if isinstance(info, dict) and info.get("reward") is not None:
+                    env_reward = float(info["reward"])
+                if done:
+                    return self._attach_env_reward(sample, env_reward), True
+                if observation is not None:
+                    sample = sample.observe(observation)  # +[obs(1)]
+            return self._attach_env_reward(sample, env_reward), True  # max_turns reached = terminal
+        except Exception as exc:  # noqa: BLE001 — isolate: one bad trajectory must not sink the drain
+            logger.warning("AgenticRolloutEngine: trajectory failed, returning partial: %s", exc, exc_info=True)
+            return self._attach_env_reward(sample, env_reward), True
```
CjhHa1: This converts every reset/decode/environment exception into a normal terminal trajectory. There is no failed status, so a backend outage before the first token is counted as a completed sibling and enters reward/GRPO as an empty answer; a later failure can train on a partial trajectory/reward. Infrastructure failures should propagate, or be marked explicitly failed and excluded from group assembly/training rather than returned with `done=True`.

unirl/rollout/engine/trainside/engine.py
```diff
@@ -90,34 +90,77 @@ def __init__(
...
                     end = min(start + fbs, bs)
-                    outputs.append(self.pipeline.generate(req.slice(start, end)))
+                    chunk = self.pipeline.generate(Sample(parts=[*input_parts, gen.slice(start, end)]))
+                    gen_chunks.append(chunk.parts[-1])
                     # LIN-387: no per-chunk empty_cache() — it forced allocator
                     # re-warm on the next chunk (decode 0.87s -> 2.76s spikes).
                     # Chunking alone bounds the live-tensor peak; cached blocks
                     # are reused, not leaked.
-                return RolloutResp.concat(outputs)
+                return Sample(parts=[*input_parts, Part.concat(gen_chunks)])
```
CjhHa1, line 158: This chunking is only correct for a single generated Part. For `[input, AR, diffusion]`, each chunk keeps the full AR shell, slices only diffusion, lets the pipeline fill both stages, then retains only `chunk.parts[-1]`; the generated AR output is discarded and the returned Sample keeps the original empty AR shell. Please either reject `forward_batch_size` for multi-stage Samples or slice/merge complete lineage subtrees and preserve every filled Part.

unirl/trainer/agentic_async.py
celve: Confirmed and fixed in `8375cd1b`.

Reproduced exactly as you described. The root cause is that the ID scheme's uniqueness rests on an invariant nothing enforced — *one `get_samples()` per `rollout_id`*. `_next_batch` breaks it: a refill re-submits under the same `rollout_id`, up to `_MAX_REFILLS` times per rollout step. Combined with a data source that numbers by batch position rather than prompt identity (`DefaultDataSource`, and `MultimodalRLDataSource`'s `prompt:{idx}` fallback), two drives namespace different source rows identically.

Worth noting there are two distinct corruptions, and the second needs no half-finished root — the refill fires *precisely* when the buffer is short but non-empty, so a completed group is normally still buffered:

1. `_build_tasks` overwrites `_gt_by_root`, so a buffered group is graded against another prompt's answer.
2. Two buffered groups sharing a root id collapse into one GRPO group in `_train_on_groups`.

Fixed with a monotonic per-drive nonce applied to fresh roots only; carried partials keep the ids they were submitted under and still rejoin their own siblings. Regression test: `tests/trainer/test_agentic_async_root_ids.py`.

unirl/models/hunyuan_image3/modes/t2ti.py
celve: Confirmed and fixed in `30a1fb21`.

Verified your exact case: at P=1, N=2, M=3 `conditioning()` returns 6 rows while `ar_part` holds 2, and the fill raises `Part.primitives['text'] batch 6 != sample_ids count 2`. M=1 turns out to be the only width that happened to work.

One clarification on the mechanism: the AR Part contributes nothing to that 6-row view — it is an unfilled shell and `turns()` skips ancestors with no populated primitives, so the 6 rows are the root prompt gathered onto the 6 diffusion-frontier samples.

Took your fix. Rather than open-code it, added `Sample.conditioning_at(index)` naming the `parts[:i+1]` replay idiom that the `conditioning()`/`turns()` docstrings already document, and rebuilt t2ti on the shape `BagelPipeline` t2ti already used: resolve at the AR frontier, assert the width, fill the AR Part, then re-read `conditioning()` so the walk broadcasts prompt *and* CoT onto the image frontier. Bagel's hand-rolled equivalent now uses the helper so the two implementations cannot drift. Test: `tests/models/test_hunyuan_image3_t2ti.py`.

unirl/rollout/engine/agentic/engine.py
celve: Confirmed and fixed in `7ad5b349` — flagging explicitly, since this reverses a deliberate choice.

The mechanism is as you describe: a fault before the first token returns `done=True` with zero gen Parts and reaches the grader as an empty answer, and a later fault trains on the partial trajectory. What made it concrete is that `_group_advantages` already documents *"NaN reward = crashed trajectory: excluded from the group's mean/std"* — but nothing ever produced that NaN, so the branch was unreachable, while `AgenticEnvTrainer` deliberately scored failures `0.0` (*"stays a legit group member"*). The two halves disagreed with each other.

Resolved toward exclusion, since that is what the documented contract promises. `_run_one` now attaches NaN — dropping any partial `env_reward`, which does not describe a complete trajectory — and both reward paths honour it. Measured on a group of 3 correct siblings + 1 crashed, advantages go from `[0.577, 0.577, 0.577, -1.732]` to all-zero: the fault no longer manufactures a gradient for its siblings.

Kept the failure *isolated* rather than propagating it, since one bad trajectory should not sink the drain; the warning now carries the count so the rate is measurable in a run. Test: `tests/rollout/test_agentic_failure_marking.py`.

Happy to revert to the previous `0.0` behaviour if you would rather keep failures as scored group members — in that case the fix becomes deleting the unreachable NaN branch instead, so the two halves still agree.

unirl/rollout/engine/trainside/engine.py
celve: Confirmed and fixed in `fe5fe9c1`, taking the first of your two options.

Traced it end to end: each chunk re-runs the interior stage and the engine keeps only `chunk.parts[-1]`, so the AR output is generated and then discarded and the Sample returns with its original empty shell. For PE it is worse than a dropped stage — `PEPipeline.generate` runs both levels inside one call and derives `M = len(diff) // len(ar)` from row counts, so slicing `diff` but not `ar` either raises a confusing arity error or silently mis-pairs prompts to images.

Chose rejection over lineage-aware chunking deliberately: slicing whole subtrees fixes the mis-pairing, but the interior stage still re-runs per chunk, so a temperature-sampled AR stage would give sibling images inconsistent conditioning — and that stage's peak memory is unchanged anyway, so chunking does not even achieve its own goal there. The knob belongs on the stage, which `pe_sglang_*` already does via the SGLang diffusion sub-engine's own `forward_batch_size`; the error message points there.

The guard keys on gen-Part count rather than Part count (V2V is legitimately 3 Parts / 1 gen Part and sets `forward_batch_size: 1`), and fires whenever the knob is set rather than only when this batch would chunk, so a config cannot pass at one fan-out and silently corrupt at a larger one. Swept all 113 recipes: 44 pair trainside with `forward_batch_size`, none of them multi-gen, so no existing recipe changes behaviour. Test: `tests/rollout/test_trainside_chunking.py`.

Also checked `SGLangDiffusionRolloutEngine` since the trainside comment says it mirrors that path — its `_generate_batch` only ever fills the frontier, so it never regenerates and discards an interior stage. No guard added there.

unirl/rollout/engine/agentic/engine.py
```diff
@@ -0,0 +1,546 @@
...
+        (the ``NCCLWeightSync.set_rollout_targets`` shape). The handles are plain
+        picklable Ray actor handles, so they survive the ``Worker.call`` arg path.
+        """
+        self._workers = list(actor_handles)
+        self._role = str(role_name)
+
+    def _split_request(self, request: Sample) -> List[Sample]:
+        """Fan a batched request into ``n × P`` single-trajectory tasks (the ``n``
+        siblings of a prompt share its slash-free root id — design §3.1)."""
+        return [prompt for prompt in request.split() for _ in range(self._n)]
```
CjhHa1, line 179: These `n` stochastic siblings reuse the same `Sample` and root IDs, and `_run_one` later calls `fork(1)` for each task. Consequently every sibling gets the same descendant IDs (`root/0`, then `root/0/0`, etc.) and is distinguishable only because it stays in a separate `Sample` object. Once generated Parts are concatenated for training, those IDs collide; any future ID-keyed conditioning, noise, logging, or merge cannot identify the trajectory unambiguously. The GRPO root/group ID should remain shared, but each trajectory/node needs a distinct identity (for example an explicit trajectory/branch ID, or fork `n` once and split the unique branches). The async/partial `_build_tasks` paths repeat the same pattern.

celve:
PE: 
image

celve:
qwen omni: 
image

celve:
DRPO: 
image

celve:
SD3 vLLM Omni: 
image

PR #215, opened 2026-07-16
feat(hunyuan_video): trainside recipe with climbing PickScore curve (LoRA alpha=256)

leviking98z-rgb:
## Summary

Adds a HunyuanVideo-1.0 T2V trainside recipe (`hunyuan_video_t2v_trainside_pickscore_climb.yaml`) whose PickScore reward actually **climbs** (0.70 → 0.83 over 150 rollouts), vs the default trainside recipe where the reward stays **flat**.

The single decisive change is **LoRA `alpha`: 64 → 256** (scale 1.0 → 4.0). Everything else is the previously-validated reference sampling recipe (10 inference steps, 4 SDE steps, 480×640, sps=8, lr=3e-4, clip=1e-4).

## Root cause: LoRA output scale, not precision

With the default `alpha = rank = 64` (scale 1.0), the LoRA branch's contribution to the model output is too small for the policy-gradient signal to move the reward: `grad_norm` shows ~0 and the reward is flat for the whole run. Raising `alpha` to 256 (scale 4×) makes the *same* gradient produce a 4× larger output change, and the reward climbs.

We initially suspected bf16 rounding of the tiny LoRA updates, but **fp32 training with alpha=64 did NOT fix it** (+0.011 only). So the lever is the LoRA **output scale**, not numeric precision.

## Evidence

**Controlled alpha sweep** (all else fixed: 8 GPU, batch=8, sps=8, lr=3e-4, 10 steps, 4 SDE, 480×640, clip=1e-4, PickScore on `datasets/pickscore/train.txt`):

| alpha | scale | reward trend (≈50 rollouts) |
|---|---|---|
| 64 (default) | 1.0 | flat (~+0.01) |
| 128 | 2.0 | ~flat |
| **256** | **4.0** | **climbs (+0.05)** |
| 512 | 8.0 | climbs (overshoots at high lr) |
| 64 + fp32 | 1.0 | still flat (+0.011) |

**Confirmation run** (`alpha256_32gpu_500roll`, 32 GPU, batch=32/sps=8, alpha=256): reward **0.703 → 0.831** over 150 rollouts, window-by-window monotonic:
```
r1-20:  0.705
r21-40: 0.745
r41-60: 0.776
r61-80: 0.789
...
r150:   0.831
```
This matches the historical reference runs `xt8m67nc` / `0w3vg80u` (which climbed to ~0.78-0.81).

## Reproduce
```
  --config-name=diffusion/hunyuan_video/hunyuan_video_t2v_trainside_pickscore_climb \
  num_devices=8 num_rollouts=150 \
  data_source.args.run.data_path=datasets/pickscore/train.txt \
  data_source.args.run.eval_data_path=datasets/pickscore/test.txt
```
(For 32-GPU multinode, set `num_devices=32 batch_size=32` to keep the per-worker mini-batch size.)


celve:
Should we replace the original recipe instead of adding a new one?

leviking98z-rgb:
Consolidated into #220. The validated alpha=256 HunyuanVideo trainside recipe is now included there as commit ddeefaf, alongside the canonical text-conditioning fix.

PR #216, opened 2026-07-16
feat(ltx2): raise GRPO sampling scale (sps 4->16) to unstick flat reward

leviking98z-rgb:
> **⚠️ DRAFT / RFC — preliminary signal only, NOT yet confirmed.** Opening to discuss the root-cause framing and gather a confirming long run. Do not merge yet.

## Summary

The default `ltx2_t2v_trainside` recipe leaves the reward **flat**. This RFC proposes that the cause is the **GRPO sampling scale** (`samples_per_prompt=4`), not the reward or the model, and adds `ltx2_t2v_trainside_sps16_rfc.yaml` (sps 4→16, num_updates 1→2) as a candidate fix.

## Why LTX2 differs from HunyuanVideo

Both show a flat reward, but the mechanism is **different**:

| | HunyuanVideo | LTX2 |
|---|---|---|
| grad_norm | ~0 (gradient vanishes) | **non-zero (0.001–0.006)** |
| ratio | exactly 1 | **deviates (0.9999)** |
| meaning | not training | **IS training, but wrong direction** |
| fix | LoRA alpha ↑ (output scale) | GRPO sampling scale ↑ (this RFC) |

LTX2 *is* taking gradient steps (non-zero grad_norm, ratio ≠ 1) — but the reward doesn't improve. That points to an **unreliable advantage direction**, not a vanishing gradient.

## Hypothesis: samples_per_prompt=4 is too few

GRPO advantage = `(reward − group_mean) / group_std`, where the group is the `samples_per_prompt` samples of one prompt. With **sps=4**, the group mean/std are estimated from only 4 samples → the advantage **sign is often wrong** → the gradient points in random directions → reward doesn't climb.

The default `batch=8 / sps=4 / num_updates=1` are **single-8-GPU memory-safe** values for the ~19B LTX2 (the header says *Launch: 1 node × 8 GPUs*), not GRPO-tuned. By contrast the HunyuanVideo trainside recipe uses `batch=48 / sps=16 / num_updates=2` (v1 parity). This RFC brings LTX2's GRPO signal scale back toward that.

## Preliminary evidence (why it's still a draft)

Two 32-GPU runs at **sps=16** both showed the reward window **rising at rollout ~29** (+0.018), where **every sps=4 variant had plateaued** (tested: alpha 64–1024, plain + directional PickScore, num_frames 9 & 33).

**Not yet confirmed.** In this same investigation, sps=4 runs faked a climb at rollout ~35 (+0.017) then **collapsed** to −0.024 by rollout 50. So a +0.018 bump at rollout 29 is not proof. This needs a long run confirming a monotonic climb **past rollout 50** before it can be claimed as a fix.

## Open questions for reviewers
1. Is `sps=4` in the default recipe a deliberate choice or a memory-driven default? Should the canonical recipe assume multinode?
2. Does the reward have enough headroom for LTX2? (PickScore on LTX2-2's already-strong base frames saturates near ceiling; directional PickScore was tried to add headroom.)
3. Needs multinode (32 GPU) — `sps=16` won't fit the single-node budget.

### DIFF
--- a/examples/diffusion/ltx2/ltx2_t2v_trainside_sps16_rfc.yaml
+++ b/examples/diffusion/ltx2/ltx2_t2v_trainside_sps16_rfc.yaml
@@ -1,0 +1,164 @@
+# @package _global_
+# LTX-Video-2 T2V GRPO — trainside (in-process FSDP).
+#
+# LTX-2 is a ~19B video DiT (ltx-2-19b-dev; HF Lightricks/LTX-2, diffusers layout)
+# with flow matching and Gemma3 text encoding.
+# Supports T2V (this recipe), I2V (set primitives["image"]), and T2AV
+# (enable_audio: true, LTX-2.3 checkpoint).
+#
+# Architecture:
+#   - 48 LTX2VideoTransformerBlock layers (unified video+audio attention)
+#   - 3D VAE: 32x spatial, 8x temporal, 128 latent channels
+#   - Gemma3 text encoder + connector projections
+#   - FlowMatch Euler Discrete scheduler
+#
+# Launch (1 node × 8 GPUs):
+#   bash examples/run_experiment_multinode_taiji.sh diffusion/ltx2/ltx2_t2v_trainside
+
+num_devices: 8
+batch_size: 8
+num_rollouts: 10000
+
+logging:
+  report_to_wandb: true
+  project_name: ${oc.env:WANDB_PROJECT,unirl-ltx2-t2v}
+  run_name: null
+  entity: ${oc.env:WANDB_ENTITY,null}
+  tags: [ltx2, t2v, grpo, trainside, dancegrpo]
+  log_media: true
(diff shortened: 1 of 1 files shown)


leviking98z-rgb:
Superseded by #233. The SPS16 multinode RFC recipe is now included there as `examples/diffusion/ltx2/ltx2_t2v_trainside_sps16_rfc.yaml`, updated to the aligned fp32 scheduler-state boundary and canonical per-prompt GRPO normalization. It remains explicitly labeled RFC; consolidation does not claim the long-run reward hypothesis is confirmed.

leviking98z-rgb:
Scope correction: the SPS16 RFC recipe was subsequently removed from #233 to keep that PR limited to the core LTX trainside/SGLang correctness alignment. #216 remains closed because its long-run reward hypothesis is unconfirmed; it is no longer represented as content consolidated into #233.

PR #218, opened 2026-07-17
feat(bagel): add native vLLM-Omni T2TI UniGRPO rollout

celve:
## Summary

Add BAGEL T2TI UniGRPO support using a native two-stage vLLM-Omni rollout:

- run Stage 0 autoregressive thinking followed by Stage 1 image diffusion, carrying native KV-transfer and replay metadata across the stage boundary;
- synchronize full BAGEL weights into both Omni stages and replay AR/image tracks with one shared prompt-level advantage;
- preserve strict production geometry `P=32, N=24, M=1, U=2`, with separate production and single-GPU smoke profiles;
- use exact layer-major replay to preserve native chunk/cache semantics while keeping FSDP2 collective ordering independent of variable thinking-trace depth;
- validate replay-lineage, stage-boundary, anchor-parity, scale-profile, lifecycle, and full-weight-sync behavior without retaining PR-added test files.

The colocated trainer lifecycle is also hardened:

- park only sharded Adam state and PickScore during bounded phases while FSDP parameters and shards remain GPU-resident;
- stage detached replay data on CPU and reclaim image-micro caches;
- propagate a fail-closed 90% trainer allocator cap through Ray device-pool workers;
- isolate vLLM-Omni allocator settings and restore optimizer/device placement exactly across rollout boundaries;
- make proxy injection opt-in so the launcher remains portable outside internal environments.

## Related Issue

N/A

## Test Plan

PR-added test files were exercised during review and then removed. The final PR diff against current `main` contains no test files.

- Current upstream-merged head: `python -m pytest tests -q` — 25 passed.
- Focused historical BAGEL request/adapter/scale-profile validation executed from a temporary directory outside the worktree — 43 passed.
- Original PR head before test-file removal, excluding the workstation-only `torchvision` collection gap — 218 passed, 8 skipped.
- Historical suite against the upstream-merged implementation — 217 passed, 8 skipped. One old constructor case expected a schedule failure after backend boot; current `main` intentionally resolves diffusion schedules before boot. Equivalent current behavior was checked inline: invalid diffusion schedules fail before boot, AR-only adapters do not resolve a diffusion schedule, and post-boot weight-sync failure shuts the backend down exactly once.
- The two local-reward parking cases blocked only by the workstation's missing `torchvision` import were rerun with a minimal import stub — both passed.
- Exact vLLM pod suite at allocator-hardening revision `2a9f6d11`: 58 passed in 17.57 seconds, covering scale profiles, constructor cleanup, flow-many replay, reward parking, and optimizer parking/restoration.
- Ruff/pre-commit, Python compilation, `bash -n scripts/launch_bagel_vllmomni_t2ti.sh`, recipe target validation, and `git diff --check`: passed.
- Real BAGEL/FlashAttention H20 parity checks: exact layer-major replay matched chunk-major replay for K/V, velocity, transition mean, log-probability, and selected gradients.
- Two-H20 CUDA/NCCL FSDP2 check: unequal replay depths completed with bf16 compute, fp32 masters, activation checkpointing, and `reshard_after_forward=true`.
- End-to-end R12 acceptance:
  - one node with 8 H20 GPUs;
  - BAGEL-7B-MoT, PickScore prompts/reward, Torch 2.11.0+cu129, vLLM 0.20.0, and vLLM-Omni 0.20.0;
  - global `P=32, N=24, M=1, U=2`, producing 768 pairs and two disjoint updates per round;
  - two finite rounds completed in 3,141.71 s and 3,130.27 s;
  - minimum physical headroom remained 7,312 MiB and 7,320 MiB;
  - all ranks crossed the third Omni wake and resumed generation.

  Run: https://wandb.ai/linyuwus/bagel-unigrpo/runs/s3x6q1h5

## Compatibility / Risk

- Image fanout is intentionally strict `M=1`; general `M>1` support is outside this PR.
- The production profile requires vLLM and vLLM-Omni 0.20.0 with Torch 2.11/CUDA 12.9.
- The checked-in production topology is 32 GPUs. Runtime acceptance used one 8-H20 node with the same global workload, so each rank carried four times the samples and roughly four times the FSDP/Adam shard bytes. Real 32-GPU production remains unmeasured.
- The H20 evidence predates the final current-main conflict resolution; the merged head was CPU/static-regression validated but was not rerun end to end on BAGEL hardware.
- Exact replay remains the production default. Collapsed replay is experimental and disabled because captured-trace numerical parity did not pass.
- The allocator constraint applies only to GPU-resident BAGEL T2TI flow-many training and fails closed for missing, invalid, non-finite, or greater-than-0.90 limits.
- Shared trainer, FSDP lifecycle, reward-service, device-pool, and full-weight-sync behavior is config-gated. There is no checkpoint or dataset-format migration.
- Proxy settings are no longer installed by default. Deployments that need a proxy can set `STAR_PROXY_URL` and optionally `STAR_NO_PROXY`.

## Reviewer Notes

Suggested review order:

1. Native Stage 0-to-Stage 1 contract and replay-track lineage.
2. Exact layer-major replay and FSDP collective ordering.
3. Old-policy anchor and detached stage-boundary semantics.
4. Optimizer/reward parking and allocator propagation.
5. Production/smoke recipes and scale assertions.

Current `main` was merged in commit `a6dcad0`. Conflict resolution preserved both native BAGEL T2TI behavior and upstream packed-T2I/scheduler behavior; deprecated request metadata access was migrated from `stage_config` to `task_config`, and constructor cleanup follows current pre-boot schedule resolution.

The end-to-end R12 run used runtime revision `b086fd1e`. The later fail-closed allocator hardening at `2a9f6d11` was verified in a detached exact-pod worktree but was not substituted into that live process. The detailed incident history, rejected alternatives, and measured acceptance evidence are recorded in `docs/bagel_vllm_omni_scale_mismatch_report.md`.

AI assistance was used during implementation, investigation, test development, and review. The resulting behavior was validated with unit, parity, distributed CUDA/NCCL, and end-to-end GPU runs.

## Checklist

- [x] I validated the implementation, docs, and configs; this PR adds no test files.


CjhHa1:
I reviewed the final PR diff against the current `upstream/main` (`641f49d6`). My recommendation is **not to merge or conflict-rebase this PR as-is; supersede it with focused PRs on current main**.

The remaining product value is real but narrower than the current title suggests: main already has trainside BAGEL Prompt → Thinking → Image UniGRPO (`#148`), while it still lacks the native two-stage vLLM-Omni `bagel_t2ti` rollout, exact Stage-0 KV-lineage replay, and shared full-weight publication. vLLM-Omni 0.27 still supports `bagel_think`, so that capability remains worth implementing.

Main blockers for this branch:

- It targets vLLM/vLLM-Omni 0.20 and Torch 2.11/cu129, including the hard-coded `bagel_vllm_omni_0_20` load plan. Main now uses vLLM 0.27, vLLM-Omni 0.27rc1, and Torch 2.13/cu130.
- The adapter/trainer path uses `RolloutReq`/`RolloutResp` and `DiffusionOutput.custom_output`; main is `Sample`/`Part` native and uses the current capture/postprocess metadata path.
- Main is 98 commits beyond the merge base, and a trial merge reports 28 content-conflict files across the algorithm, BAGEL model, rollout, trainer, backend, and distributed layers. Whole-file conflict resolution would regress newer BAGEL it2i conditioning, ordered AR interleaves, current FSDP wrapping, and vLLM-Omni lifecycle handling.
- `UnifiedModelTrainStack` uses `1 / len(micro_slices)` for every micro-batch. For uneven micro sizes this overweights smaller micros; current main correctly uses `(end - start) / update_total`.
- `_move_bundle_module_state()` recursively moves every module discovered under the bundle, including auxiliaries such as VAE/text/vision encoders whose residency may be owned elsewhere. This is unsafe as a generic backend lifecycle change.
- The final head removed 21 feature-test files (7,943 test lines) from its pre-merge parent, so the final PR diff contains no T2TI/replay/lifecycle tests. The final conflict-resolved head was not rerun end-to-end on hardware; full RatioNorm/FSDP parity and the encoded 32-GPU production topology also remain untested.
- The known parity-failing `collapsed` replay mode remains exposed, and several BAGEL-only capabilities are added as universal base flags/no-op hooks. There is also dead `_snapshot_reference` compatibility code.

Parts worth preserving in new work:

1. A current `Sample`-native `bagel_t2ti` adapter with strict Stage-0/Stage-1 pairing, emitted-token log-prob validation, and typed KV replay lineage.
2. Exact layer-major replay for variable thinking depths, with the historical tests recovered and adapted to current APIs.
3. Selected-stage full-weight publication with fresh per-stage serialization, ACK/name/checksum validation, and failure propagation, rebuilt on the current sync abstractions.
4. Full-FT immutable reference persistence as a separate correctness fix. Main can currently capture already-trained weights as the new reference after resume; the sidecar approach here should instead be integrated atomically and topology-independently with checkpoint state.
5. Exact optimizer-state parking only if fresh CUDA-13 measurements still require it. The broad bundle lifecycle and reward parking should not be carried over; current managed scorers already support per-call residency.

The historical H20 work and incident analysis are valuable design evidence, but the current branch has low operational merge value and high regression risk. A small v0.27-compatible implementation with reproducible tests would be substantially safer to review and maintain.

PR #219, opened 2026-07-17
feat(boogu_image): Boogu-Image-0.1-Base T2I bundle + trainside FlowGRPO recipe

celve:
## Summary

Adds RL post-training support for **Boogu-Image-0.1-Base** (T2I): a new
`unirl/models/boogu_image/` bundle + a trainside FlowGRPO recipe.

- **Vendored DiT** (`vendor/`, pinned at boogu-project/Boogu-Image `434fb56`,
  documented in `VENDOR_COMMIT.txt`): the 10.29B Lumina-2-lineage
  single/double-stream transformer is not in diffusers, and the HF
  checkpoint's "custom code" file is a shim importing the GitHub package, so
  vendoring (bagel precedent) is the only clean path. Mechanical edits only:
  flattened imports, SDPA processors pinned (upstream selects attention
  backends from a `device` env var — unacceptable for training numerics),
  torch RMSNorm/swiglu default branches, raising TaylorSeer/TeaCache stubs
  (RL must be cache-free).
- **Bundle package**: Qwen3-VL chat-template embed stage (fixed T2I system
  prompt; the CFG negative `""` routes to the DROP prompt per the reference's
  dataset logic), diffusion step/stage with the t-convention sign flip
  (t = 1 − σ), sequential CFG gated `> 1.0` with plain linear combine,
  resolution-independent cached RoPE tables, FLUX.1 `AutoencoderKL` fp32
  decode, trivial-case meta-init (the vendored tree has zero registered
  buffers), and `build_schedule_policy = static_only(e^1.15)` — Boogu's
  released static-v1 time shift (seq_len 4096) is algebraically the standard
  static σ-shift with `s = e^1.15` (verified to 1 fp32 ulp; Boogu's
  scheduler JSON uses custom field names the generic loader would silently
  ignore).
- **Recipe** `examples/diffusion/boogu_image/boogu_image_trainside.yaml`:
  FlowGRPO + LoRA r64 + PickScore at 512², CFG-off (Boogu's off value is
  1.0). FSDP `block_class_names` lists the 5 concrete instantiated block
  classes (exact type-name matching; the DoubleStream block is a separate
  class, not a subclass). LoRA targets include the double-stream
  joint-attention projections that live *inside* the attention processors
  (`img_to_q`/`instruct_to_q`/… — dot-boundary suffix matching cannot reach
  them via `to_q`). `master_dtype: fp32`, `old_logp_source: replay`.
- `pyproject.toml`: `einops>=0.7` (imported by the vendored DiT; pure-python,
  engine-agnostic). `.pre-commit-config.yaml`: vendored tree joins
  `bagel/vendor/` in the hook exclude.

**Update (2026-07-27):** rebased onto current main and migrated to the
Sample-native rollout model (#214): `generate()` is now the
`Sample -> Sample` endomorphism — prompt via `sample.conditioning()`,
`DiffusionSamplingParams` (with pinned `sigmas`) from the frontier gen Part,
noise groups from `frontier.group_ids`, x_T via `NoiseRecipe.from_sample`,
frontier filled with segment + decoded images + replay conditions. Also
adopts the new `build_meta_init_transformer` API and adds `load_vae`
(mirrors the z_image migration). Re-verified on 8×H20: the full parity gate
re-PASSED with byte-identical metrics (teacher-forced velocities still
bit-exact, replay ratio exactly 1), and the Sample-native trainer loop runs
clean (pre-check rewards match the pre-port run's same-seed values).

## Related Issue

N/A (README roadmap invites model requests; no existing issue or PR for
Boogu-Image — checked open PRs/issues).

## Test Plan

Current-head refresh (`ca2b647f954698467d1a743af79d7feba04a4299`): merged current `main` and resolved the two shared-file conflicts by retaining both the Boogu vendor exclusion and the benchmark-data exclusion, plus both the current dependency floors and Boogu’s `einops` dependency. On the merged head: 25 existing repository tests passed; `python -m compileall -q unirl/models/boogu_image` passed; the checked-in recipe composed and resolved; 1,997 Hydra `_target_` paths resolved; and pre-commit passed on every file in the final PR diff. The final diff adds no test files. The GPU evidence below is the contributor’s original Boogu checkpoint validation and was not re-run during this current-main refresh.

All run on 8×H20-96G (gz), checkpoint `Boogu/Boogu-Image-0.1-Base` @ HF
revision `7f464b5cabb4b0699b85409ab942337d8188124c` staged pod-local;
venv torch 2.11.0+cu129 / diffusers 0.38.0 / transformers 5.11.0.

- `python -m unirl.train_diffusion --config-name=diffusion/boogu_image/boogu_image_trainside --cfg job --resolve` — passes with and without `BOOGU_IMAGE_PATH`.
- `python -m compileall -q unirl`; `bash -n examples/*.sh`; `SKIP=no-commit-to-branch pre-commit run --all-files` — green (incl. the recipe-`_target_` hook).
- **Step-level parity vs the reference `BooguImagePipeline`** (1024², 50
  steps, CFG 4.0, seed 0, SDPA both sides): σ grid max|Δ| 1.19e-7;
  state-dict keys exactly match the checkpoint index; chat-template
  attention masks byte-equal; **teacher-forced per-step velocity bit-exact
  (cosine 1.0, max|Δ| 0 at all 50 steps)**; free-running from shared x_T
  (η=0): final-latent cosine 0.9953, decoded PSNR 32.7 dB; isolated VAE
  decode PSNR 56 dB (fp32 stage vs the reference's bf16 decode);
  diffuse↔replay ratio exactly 1; replay-backward smoke: 942 finite grads.
- **Wiring pre-check** `num_rollouts=2 batch_size=8 sampling.samples_per_prompt=4`:
  full loop clean, first-update ratio 1.0000±0.0000.
- **RL smoke** `num_rollouts=100` (48 prompts × 16 samples @ 512², 12 steps,
  LoRA) — **completed 100/100**: PickScore reward mean rose monotonically
  from **0.7519 ± 0.018 (first 20) to 0.8687 (last 20)** — +0.117 ≈ 6.5×
  the early-window std (max 0.8857) — with ratio_mean ≡ 1.0000 on first
  updates, approx_kl ≈ 0, and zero degenerate groups throughout
  (W&B run `unirl-boogu-image/c3h5ddt4`).

## Compatibility / Risk

Additive: new model package + new recipe dir. Shared-file touches are one
dependency line (`einops`, engine-agnostic pure python) and one pre-commit
exclude entry. No runtime/trainer/SDE/reward changes. Note: validation ran
on transformers 5.11.0 (venv) vs the pyproject `>=5.6,<5.7` pin — Qwen3-VL
encoding was bit-exact vs the reference on 5.11; flagging in case the pin
should be revisited separately.

## Reviewer Notes

- Start with `unirl/models/boogu_image/vendor/VENDOR_COMMIT.txt` (every
  mechanical edit + re-vendor recipe), then `diffusion.py` (the convention
  adapter: sign flip, timestep dtype, bare-tensor return, CFG gate).
- Follow-ups out of scope here: Edit (TI2I) overlay, Turbo (DMD few-step)
  RL, SGLang/vLLM serving (fork-side model support required), VeOmni SP wrap.
- AI-assisted implementation; the full diff was reviewed and every gate
  above executed on hardware before publishing. Duplicate-work check: no
  open upstream PRs/issues mention Boogu.

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,9 +8,11 @@ default_stages: [pre-commit, pre-push, manual]
 #   - unirl/models/bagel/vendor/ : pristine official ByteDance-Seed/Bagel modeling
 #     (see VENDOR_COMMIT.txt); kept byte-identical to upstream, so the whitespace/format
 #     hooks must not rewrite it either.
+#   - unirl/models/boogu_image/vendor/ : pristine boogu-project/Boogu-Image model code
+#     (see VENDOR_COMMIT.txt); same re-vendor rationale as bagel/vendor/.
 #   - benchmarks/*/*/data/ : vendored upstream benchmark prompt sets (GenEval,
 #     PartiPrompts, VBench, ...); kept byte-identical so upstream checksums verify.
-exclude: ^(unirl-reward-service/|unirl/models/bagel/vendor/|benchmarks/[^/]+/[^/]+/data/)
+exclude: ^(unirl-reward-service/|unirl/models/bagel/vendor/|unirl/models/boogu_image/vendor/|benchmarks/[^/]+/[^/]+/data/)
 
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
--- a/examples/diffusion/boogu_image/boogu_image_trainside.yaml
+++ b/examples/diffusion/boogu_image/boogu_image_trainside.yaml
@@ -1,0 +1,215 @@
+# @package _global_
+# Boogu-Image-0.1-Base + trainside (FSDP) colocate, v2 trainer, FlowGRPO.
+#
+# Trainside = rollout runs in-process on the FSDP-wrapped training pipeline
+# (no separate rollout actor, no weight sync). Strongest colocate form.
+#
+# This recipe targets Boogu/Boogu-Image-0.1-Base: a 10.29B vendored
+# single/double-stream DiT (Lumina-2 lineage) + frozen Qwen3-VL-8B mllm
+# instruction encoder + FLUX.1 16-channel AutoencoderKL. The released Base
+# scheduler is a STATIC v1 time shift (seq_len=4096) whose sigma-space form is
+# the standard static shift with s = e^{1.15} = 3.158192909689768 — pinned via
+# BooguImagePipeline.build_schedule_policy (Boogu's scheduler_config.json uses
+# custom field names the generic JSON loader would silently ignore).
+#
+# CFG is intentionally OFF here (guidance_scale=1.0 — Boogu's off value; the
+# combine is pred + (g-1)*(pred - pred_neg)). Same rationale as z_image: CFG
+# sharpens the per-prompt distribution (less in-group diversity -> weaker GRPO
+# advantage) AND doubles the per-step forward cost. To re-enable: set
+# sampling.guidance_scale to 2.0-4.0 (reference inference uses 4.0); the
+# pipeline then auto-adds ""-negatives, which the embed stage routes to the
+# DROP system prompt (dataset logic).
+#
+# Resolution note: Boogu's native band is 1K/1.5K/2K; 512^2 is off-band but
+# the sigma schedule is resolution-invariant (config-pinned static shift), and
+# 512^2 keeps the smoke cheap. If early rollout/generated_media look too
+# degenerate for PickScore to rank, fall back to 1024^2:
+#   sampling.height=1024 sampling.width=1024 rollout.forward_batch_size=1 \
+#   stack.micro_batch_size=1 batch_size=16 sampling.samples_per_prompt=8
+#
+# Scale is set by the launcher (num_devices = cluster GPU count). For a quick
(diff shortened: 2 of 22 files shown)


celve:
image

Ideny42 (approved):
I've got a question — Boogu leans pretty hard on the agentic rewriter
(Turbo-Thinking is their strongest variant), but this recipe RLs directly on
raw prompts. Should we also train in PE/agentic mode (frozen rewriter in the
loop, reward still against the original prompt) to match the model's actual
use case? Or is raw-prompt trainside just meant as the first step? Rest looks
fine to me.

celve:
> I've got a question — Boogu leans pretty hard on the agentic rewriter (Turbo-Thinking is their strongest variant), but this recipe RLs directly on raw prompts. Should we also train in PE/agentic mode (frozen rewriter in the loop, reward still against the original prompt) to match the model's actual use case? Or is raw-prompt trainside just meant as the first step? Rest looks fine to me.

Yes I think we should support the PE/agentic things. Will try to do that when #214 is merged.

Issue #217, opened 2026-07-17
[rollout, ckpt] RFC: Multi-GPU SGLang Rollouts with TP, EP, and Checkpoint-Engine Weight Sync


NancyFyong:
- **Status:** need review
- **Related Draft PR:** [#191 — WIP: split rollout TP/EP/PP parallelism into focused PRs](https://github.com/Tencent-Hunyuan/UniRL/pull/191)
- **Split review PRs (review in order):**
  - #227 — `review/rollout-tp-ep-lora-sync` — AR rollout layout, SGLang TP/EP topology, LoRA sync
  - #225 — `review/weight-sync-ckpt-engine-ipc` — checkpoint-engine IPC weight sync
  - #226 — `review/ar-colocate-memory-qwen` — colocate-memory safeguards, Qwen MoE support, TP payload handling

## What does this RFC do?

This RFC proposes a rollout-side parallelism model for SGLang in colocated
autoregressive reinforcement learning. It lets one SGLang rollout engine span
multiple GPUs with tensor parallelism (TP), forwards expert-parallel (EP)
settings to SGLang, keeps unsupported pipeline parallelism (PP) fail-closed,
and adds a checkpoint-engine IPC transport for efficient full-weight updates.

The proposal is deliberately delivered as four reviewable changes rather than
one large feature PR. The existing integration branch remains useful for
end-to-end regression testing, but it is not the preferred review or merge
vehicle.

## Motivation

Before this work, colocated SGLang rollout assumed one rollout engine per GPU:
`dp_size == world_size`. That model cannot serve rollout models that need more
than one GPU per engine, such as sparse MoE models with long-context sampling.
It also makes the training and rollout memory peaks compete on the same device.

Supporting multi-GPU rollout requires more than passing `tp_size` to SGLang:

1. UniRL must assign logical rollout ranks to contiguous TP groups.
2. Exactly one worker in each TP group must own the SGLang server and its
   ports; the other training workers must remain collective participants
   without starting duplicate servers.
3. All weight-sync transports must retain collective lockstep while avoiding
   duplicate pushes from non-owner workers.
4. GPU identifiers must remain valid when Ray restricts
   `CUDA_VISIBLE_DEVICES` per worker.
5. Colocated AR training must release training-side memory before rollout and
   reclaim transport buffers between full-weight updates.

## Goals

1. Support `tp_size > 1` for SGLang rollout in colocated training.
[...]

PR #225, opened 2026-07-20
feat(weight-sync): add SGLang checkpoint-engine IPC sync

NancyFyong:
## Summary

Add an opt-in SGLang checkpoint-engine transport for colocated full-weight synchronization.

- add `CkptEngineIPCWeightSync` and a checkpoint-engine 0.4.2-compatible trainer sender;
- create one CUDA IPC/ZMQ route per colocated SGLang scheduler using GPU UUIDs;
- keep train ranks in collective lockstep through prepare, handshake, tensor staging, bucket, release, and completion phases;
- poison and terminate SGLang when a transfer failure can leave checkpoint-engine REP workers blocked;
- fail closed for unsupported PP, active speculative workers, server-level DP, cross-host TP, and GPU/endpoint mismatches;
- keep checkpoint-engine imports explicit so ordinary full-sync modules do not load optional `pyzmq`;
- add a 21-line Hydra delta recipe that inherits the existing TensorWeightSync recipe.

The CUDA driver / torch-memory-saver scheduler-spawn compatibility work remains split into #468. Tensor, NCCL, and generic IPC transports are unchanged by the final diff.

## Related Issue

- RFC: #217
- TP/EP full-weight prerequisite: #226
- CUDA scheduler-spawn compatibility split: #468
- Integration history: #191

## Test Plan

Validated at current head `219b7b65`:

- `python3 -m compileall` on changed Python modules — passed;
- Ruff check and format check — passed;
- `python3 lint/check_recipe_targets.py` — 2716 recipe `_target_` paths resolve;
- Hydra composition of `ar/qwen3_cppo_30b_a3b_base_dapo_sglang_ckpt_engine_ipc` — passed;
- pre-commit on every changed file — all hooks passed;
- focused smokes: optional `pyzmq` import isolation, lazy public export, empty-handle rejection, stalled-receiver shutdown, and ProcessGroup abort fallback — passed;
- no test file is added by this PR.

A final-head 8×H20 CE smoke at `219b7b65` completed all three rollouts in 112s with two successful `/update_weights_from_ipc` updates. The process imported SGLang `0.5.12.post1` from the benchmark venv and checkpoint-engine `0.4.2`, validating the socket-init consensus, single-socket transfer, buffer lifecycle, and final HTTP join behavior against the declared public dependencies. Focused fault-path smokes additionally cover constructor failure consensus, handshake-error ACK delivery with positive linger, raw collective error translation, and sender cleanup before complete consensus.

## Compatibility / Risk

The transport is same-node colocate and HTTP-backend only. It pins `checkpoint-engine==0.4.2` through the SGLang extra. One logical tensor must fit in one sender bucket;the Qwen3-30B delta recipe uses 640 MiB because its bf16 embedding is about 594 MiB. Unsupported PP>1, active speculative decoding, SGLang server-level DP>1, incomplete TP metadata, and cross-host TP fail closed.

`asdict_drop_none` retains both dataclass and msgspec.Struct serialization because the public pinned SGLang release and the current deployment image expose different io_struct representations.

## Reviewer Notes

Review scope is the checkpoint-engine wire protocol, distributed orchestration, and minimal SGLang HTTP receiver seam. Deployment-specific CUDA/`LD_PRELOAD` handling is intentionally absent. The existing TensorWeightSync recipe is unchanged;the checkpoint-engine recipe inherits it and overrides only the transport, required bucket size, timeout, and logging identity.

## Checklist

- [x] I reviewed the final diff against its real upstream merge base.
- [x] I removed unrelated native Tensor/NCCL session wrapping and redundant PYTHONPATH forwarding.
- [x] I verified callers, defaults, optional capability boundaries, and nested failure propagation.
- [x] I updated the opt-in config and validation notes without duplicating the base recipe.


CjhHa1 (changes requested):
## Review

Request changes. The checkpoint-engine IPC transport is the right split and the fail-closed / per-GPU sender design is sound, but the advertised native TP8 path cannot boot on this head, and the branch did not apply cleanly to current `main`.

### Blockers

1. **Recipe / test plan vs native TP gate.** The original PR switched `qwen3_cppo_30b_a3b_base_dapo_sglang.yaml` to `CkptEngineIPCWeightSync` while leaving `backend: native` + `tp_size: 8`. `NativeBackend.boot` still raises `NotImplementedError` for `tp_size>1`. The claimed 8×H20 native TP8 smoke cannot start as written. Either finish native TP or document/test HTTP (the existing full-weight recipe already uses HTTP + TP8).

2. **Rebase required.** GitHub reported `CONFLICTING` against current `main`. `full/__init__.py` had a real conflict; the sglang backends have drifted. Automerge of “changed in both” files is not safe.

### Should-fix

3. **Do not replace the existing 30B tensor recipe.** `examples/ar/qwen3_cppo_30b_a3b_base_dapo_sglang_full_weight_sync.yaml` is already the opt-in full-weight file. Add a new ckpt-engine recipe instead of changing the default 30B yaml.

4. **CUDA / torch-memory-saver monkeypatches are out of scope** (~160 lines, commits from `review/ar-colocate-memory-qwen`). `HookUtilModePreload.get_path_binary` is patched process-wide and never restored. Keep if H20 spawn actually needs them, but call that out; prefer a follow-up PR.

5. **No tests for the new transport.** Handshake / oversized tensor / topology fail-closed / thread placement are all unit-testable.

6. **`track_prefix` was accepted then dropped** on the engine method, and the trainer-side call did not forward `self._track_prefix`.

7. **`weight_version` skew.** Non-TP-zero and TP-zero incremented on different success paths (or only TP-zero incremented). A receiver-only failure (e.g. `flush_cache`) leaves the TP group with mismatched versions.

### Nits

- Prefer current GPU UUID over `list(zmq_handles.items())[tp_rank]`.
- `weight.view(-1).view(uint8)` needs a contiguous tensor.
- Server-level DP was only read from `cfg.dp_size`; `engine_kwargs["dp_size"]` slipped through.

### What looks good

- Per-GPU reusable IPC buffer (avoids concentrating flatten copies on TP-zero).
- Prepare all-reduce + per-phase sender consensus so FSDP all-gather stays aligned.
- Oversized tensor uses `raise`, not `assert`.
- Native vs HTTP thread placement matches LoopThread parking; pyzmq sockets stay on the sender thread.
- PP>1, cross-node TP, incomplete TP metadata, speculative kwargs fail closed.
- `checkpoint-engine>=0.4.2,<0.5` is scoped to the sglang extra.

I rebased this onto current `main` locally and applied the recipe / version / topology / test fixes on `review/pr-225` at `/data/home/aimicahchen/UniRL_Project/UniRL-pr225-review` (`7155c51`). Happy to push that branch if you want it as a follow-up.

CjhHa1 (approved):

PR #226, opened 2026-07-20
feat(weight-sync): support SGLang TP/EP full-weight sync

NancyFyong:
## Summary

Independent split for SGLang AR rollout TP/EP full-weight sync.

Review scope:
- support full-weight tensor/NCCL sync with SGLang TP/EP rollout layout;
- keep non-host TP rollout workers participating in train collectives while `tp_rank==0` owns the SGLang HTTP server;
- send per-device SGLang TP tensor payloads so each scheduler receives an IPC handle from the correct device;
- reject unsupported generic IPC TP layouts instead of silently using the wrong local-rank path;
- pack split Qwen MoE expert checkpoint keys during FSDP2/meta-init sharded load;
- support Qwen3/Qwen3.5 MoE packed export naming in full-weight sync;
- fail closed when required SGLang `ServerArgs` fields such as `ep_size`, `enable_expert_parallel`, `enable_memory_saver`, `enable_weights_cpu_backup`, or `skip_server_warmup` are explicitly requested but unsupported by the installed SGLang build;
- add a Qwen full-weight TensorWeightSync recipe using environment variables for model and data locations.

Current strategy:
- rollout TP/EP is supported through the HTTP backend only;
- native backend `tp_size>1` fails fast;
- child-local `CUDA_VISIBLE_DEVICES` is applied before importing SGLang or touching `torch.cuda`;
- this PR intentionally does not patch SGLang scheduler internals, torch-memory-saver, `LD_PRELOAD`, or `LD_LIBRARY_PATH`.

Out of scope:
- LoRA hot-load / adapter sync changes;
- checkpoint-engine IPC transport implementation;
- diffusion / `sglang_diffusion` TP>1 support;
- deployment-specific CUDA forward-compatibility shims for a CUDA runtime newer than the host driver.

## Related Issue

- RFC: #217 — Multi-GPU SGLang Rollouts with TP, EP, and Checkpoint-Engine Weight Sync
- Integration / regression reference: #191 — the original unsplit rollout branch; this split supersedes it for review.

## Test Plan

Validated at current head `5e116e5ecc30c17a7bb5cc917e51c317e7b47f07`:

- `SKIP=no-commit-to-branch pre-commit run --all-files --show-diff-on-failure` — every configured hook passed.
- Ruff check and format check on the four changed source files in the final follow-up — passed.
- `python -m compileall -q unirl` — passed.
- `python scripts/check_recipe_targets.py` — 2303 recipe `_target_` paths resolve.
- `git diff --check` — passed.
- GitHub checks — 5/5 successful (lint, PR body, title, status label, and open-source scan).
- The final PR diff contains no `tests/` paths, per maintainer direction.

Current-head 8-GPU smoke:

- Hardware/runtime: one node with 8× NVIDIA H20 96 GB, SGLang `0.5.12.post1`, PyTorch `2.11.0+cu130`, Qwen3-30B-A3B-Base.
- Workload: 2 rollouts, 8 prompts × 2 samples, batch size 8, horizon/max-new-tokens 32, one optimizer update per batch, rollout concurrency 1, and 512 MiB full-weight buckets.
- The launcher recorded the exact head above before starting.
- SGLang settled `tp_size=8`, `ep_size=8`, `enable_memory_saver=True`, and `enable_weights_cpu_backup=True`; all scheduler ranks initialized as `TP0..TP7 / EP0..EP7`.
- UniRL reported TensorWeightSync layout `world=8, dp=1, tp=8, ep=8`.
- Rollout 1 completed with a non-zero update: `reward=0.6875`, `loss=0.0619`, `gn=3.2042`, `lr=1e-6`.
- The subsequent full-weight sync completed 114/114 `/update_weights_from_tensor` requests with HTTP 200.
- Memory release/resume completed, second-round generation returned HTTP 200, and rollout 2 completed: `reward=0.5625`, `loss=-0.0256`, `gn=5.3634`, `lr=1e-6`.
- The job exited 0 and cluster cleanup released training GPU memory.

The H20 validation node has driver 535 while the pinned environment is CUDA 13.0. The validation launcher therefore supplied `cuda-compat-13` and a validation-only, checkout-external shim that preserved its preloaded compatible driver through torch-memory-saver's scheduler spawn. Nothing from that shim is committed or claimed as functionality of this PR. On an older-driver forward-compat deployment, the image/launcher must provide the same property; otherwise use a driver natively compatible with the pinned CUDA runtime.

The earlier [same-token replay and deliberately stale-weight A/B validation](https://github.com/Tencent-Hunyuan/UniRL/pull/226#issuecomment-5146016622) was run at `f76dbfd`: normal TP8 sync reduced stale-weight mean `|Δlogp|` by about 251× and mean K3 by about 3731×. It remains numerical evidence for the unchanged normal SGLang full-weight transport path, but is not presented as a current-head rerun or as proof of bitwise/long-run curve identity.

No dataset, checkpoint, log, or generated output is added to this branch.

## Experiment Snapshot

Longer-run reference, not a substitute for the current-head validation above: Qwen3-30B-A3B, CPPO, a single 8-GPU node, all transformer weights trainable, LR `1e-6`, SGLang HTTP TP=8 / EP=8, 16 prompts × 16 samples/prompt, rollout concurrency 64, sync every rollout, `mem_fraction_static=0.15`, `cuda_graph_max_bs=64`, `max_new_tokens=4096`, `temperature=1.0`, `top_p=1.0`, `top_k=0`, and eval every 10 rollouts with avg@8.

### Curves

> **Full weights:** train/eval reward curve from the longer run.

Full-weight train/eval reward curve

## Compatibility / Risk

This branch duplicates the minimum rollout TP/EP layout needed to run full-weight sync independently from the LoRA PR. The full-weight changes tune memory and sync defaults for large MoE AR rollout. Generic IPC TP now raises instead of silently using an unsupported local-rank-0-only path. The tensor payload change depends on all TP ranks participating in payload gather so each scheduler receives a CUDA IPC handle from the correct device.

Explicit first-class SGLang flags now fail closed if the installed runtime's `ServerArgs` does not accept them. This may require a newer SGLang build for recipes that rely on memory saver, weights CPU backup, or expert parallelism.

CUDA compatibility is an explicit deployment boundary. With a driver natively compatible with the pinned CUDA runtime, no driver shim is needed. With an older driver plus NVIDIA's forward-compatibility layer, the Ray daemon and SGLang scheduler descendants must inherit a consistent compatible library path, and torch-memory-saver must not discard a required driver preload while replacing `LD_PRELOAD` with its own hook. This PR does not monkey-patch those third-party internals; that setup must be supplied by the image/launcher and was supplied externally for the H20 validation above.

The meta-init loader change is limited to Qwen-style MoE checkpoints that store per-expert `gate_proj`, `up_proj`, and `down_proj` weights while the model expects packed expert tensors. Missing or inconsistent expert keys fail closed instead of being silently ignored. The normal Transformers `from_pretrained` path is unchanged.

Diffusion trainer behavior is intentionally left unchanged.

## Reviewer Notes

The CUDA-driver-preload / torch-memory-saver compatibility commits in the branch history were deliberately superseded, not accidentally lost during conflict resolution. The final implementation keeps the child-local device mapping at the spawn boundary and avoids patching installed SGLang or torch-memory-saver internals. The forward-compatibility requirement still exists on mismatched driver/runtime hosts and is documented above as a deployment prerequisite rather than claimed as UniRL functionality.

The `CUDA_VISIBLE_DEVICES` ordering invariant is documented immediately at `_launch_server_with_env`: the override must run before importing SGLang or touching `torch.cuda`, because CUDA visibility can be cached after initialization.

This branch is no longer stacked on the LoRA or checkpoint-engine branches. It intentionally carries only the full-weight TP/EP sync path plus the minimum shared rollout layout needed to run that path.

Duplicate-work check: #191 is the original unsplit rollout branch; this PR is the focused full-weight split. AI assistance was used to prepare the split and PR text; the branch diff and validation were checked before publication.

## Checklist


unirl/rollout/engine/sglang/backends/base.py
```diff
@@ -35,6 +34,34 @@
 )
 
 
+def _normalize_cuda_visible_devices(
+    cuda_visible_devices: Optional[Sequence[str]],
+    *,
+    tp_size: int,
+) -> Optional[List[str]]:
```
CjhHa1, line 72: **Blocking: the merge commit looks like it dropped the entire content of three commits on this branch.**

Three commits here are titled `fix(sglang): preserve preloaded CUDA driver shims` (0ae9c2c), `fix(sglang): retain CUDA shims through memory-saver spawn` (a4ea0aa), and `fix(sglang): retain CUDA driver for scheduler descendants` (0ea1566). Comparing this file between that last commit and the current head:

| | `0ea1566` | head `f7d1f8e` |
|---|---|---|
| `backends/base.py` | 335 lines | 147 lines |
| helpers present | `_split_preloads`, `_python_cuda_library_dirs`, `_preloaded_cuda_driver_libraries`, `_preloaded_cuda_driver_dirs`, `_preserve_cuda_driver_preloads`, `_configure_torch_memory_saver_composite_preload`, `_run_sglang_scheduler_with_cuda_driver_preload`, `_scheduler_spawn_environment`, `_normalize_cuda_visible_devices` | `_normalize_cuda_visible_devices` only |

Those three commits contribute `+151/-26` to this file, `+28/-2` to `backends/http.py`, and `+17/-3` to `backends/native.py`. The merge commit `f7d1f8e` lists `backends/base.py` and `backends/http.py` among its conflicts, and after it, essentially none of that remains: this file is down to the single `_normalize_cuda_visible_devices` helper, `http.py` has a 6-line `_launch_server_with_env` in place of `_launch_server_with_cuda_driver_preloads`, and `native.py`'s share of the PR diff is only the new 6-line `tp_size>1` guard. So the branch history advertises work that is no longer in the tree.

Two things this affects:

1. **The Test Plan no longer describes this head.** It says "Validated at current head `0ea1566`" and lists "Inline scheduler/TMS compatibility check — passed: the CUDA driver preload is retained for scheduler descendants while the trainer process can restore its environment." Neither half of that sentence is true of `f7d1f8e`: there is no driver preload retention, and the parent no longer mutates and restores its own environment (the new approach sets the child's env inside the child, which is genuinely cleaner, but it is not what was validated).

2. **More than the `LD_PRELOAD` handling was lost.** `_scheduler_spawn_environment` also quarantined `LD_LIBRARY_PATH`; its docstring notes that Ray Workers inherit environment variables from the already-running Ray daemon rather than from the submitting command, so a stale `LD_LIBRARY_PATH` can make a scheduler load the wrong CUDA runtime. That protection is gone too. And the recipe added in this PR sets `enable_memory_saver: true`, which is precisely the torch-memory-saver path those commits were written against.

Could you clarify whether this was an accidental conflict resolution or a deliberate revert? Either way I think two things are needed before merge: the PR description and Test Plan should describe the code that is actually here, and the 8xH20 smoke should be re-run on the current head rather than on `0ea1566`. If the simplification was deliberate, it would also help to say why the torch-memory-saver interaction those three commits fixed no longer needs handling — from the outside it reads as a regression rather than a cleanup.

unirl/rollout/engine/sglang/backends/http.py
```diff
@@ -127,6 +129,16 @@ def _import_sglang_runtime() -> Dict[str, Any]:
...
 
 
+def _launch_server_with_env(server_args: Any, env_overrides: Dict[str, str]) -> Any:
+    """HTTP server target with child-local launch environment overrides."""
+    if env_overrides:
+        os.environ.update(env_overrides)
+
+    from sglang.srt.entrypoints.http_server import launch_server
+
+    return launch_server(server_args)
```
CjhHa1, line 203: This is the replacement side of the dropped commits — see the detailed comment on `backends/base.py`. At `0ea1566` this spawn target was `_launch_server_with_cuda_driver_preloads`, which wrapped the launch in `_preserve_cuda_driver_preloads(driver_libraries)` and passed `run_scheduler_process_func=_run_sglang_scheduler_with_cuda_driver_preload` so the driver shim survived into SGLang's own scheduler subprocesses. The current version only propagates `CUDA_VISIBLE_DEVICES`.

Setting the child's environment inside the child is the right shape and I like it better than mutating and restoring the parent's env — my concern is only that it silently replaced the torch-memory-saver interaction that three commits on this branch were written to fix, while the recipe added here runs with `enable_memory_saver: true`.

One narrower point on the mechanism itself, independent of the above. This relies on `CUDA_VISIBLE_DEVICES` being set before CUDA is initialised in the child. That holds today: `spawn` re-imports this module to unpickle the target, and neither this module nor `_import_sglang_runtime`'s import list initialises a CUDA context before `os.environ.update` runs. But it is an ordering invariant that nothing enforces — if anything in the module-import path ever touches `torch.cuda.device_count()` (which is `lru_cache`d) or otherwise initialises CUDA, the override becomes a silent no-op and the schedulers land on the wrong devices. A comment stating the invariant, or an assertion that CUDA is uninitialised at that point, would make it robust to a future import creeping in.

unirl/rollout/engine/sglang/config.py
```diff
@@ -246,19 +316,39 @@ def server_intent(
...
+        if self.dp_size is not None:
+            intent["dp_size"] = int(self.dp_size)
+        if self.enable_expert_parallel is not None:
+            intent["enable_expert_parallel"] = bool(self.enable_expert_parallel)
+        if self.enable_memory_saver is not None:
+            intent["enable_memory_saver"] = bool(self.enable_memory_saver)
+        if self.enable_weights_cpu_backup is not None:
+            intent["enable_weights_cpu_backup"] = bool(self.enable_weights_cpu_backup)
+        if self.skip_server_warmup is not None:
+            intent["skip_server_warmup"] = bool(self.skip_server_warmup)
```
CjhHa1, line 346: **Blocking: these newly forwarded flags are load-bearing, but the transport to `ServerArgs` drops unknown keys silently.**

`HTTPBackend.boot` filters the intent like this:

```python
allowed = {f.name for f in dataclasses.fields(rt["ServerArgs"])}
server_kwargs = {k: v for k, v in server_intent.items() if k in allowed}
```

with the docstring noting that "non-ServerArgs escape-hatch keys drop harmlessly". That is a reasonable contract for optional tuning knobs, which is what `engine_kwargs` carried until now. It is not a safe contract for the four flags this PR promotes to first-class config fields — `ep_size`, `enable_memory_saver`, `enable_weights_cpu_backup`, `skip_server_warmup` — because each of them changes correctness or memory ownership rather than performance.

`enable_weights_cpu_backup` is the one that worries me. The recipe's comment states the invariant it is meant to guarantee: "Preserve the latest rollout weights across rounds that intentionally skip a full sync (including rollout 0 when baseline eval is disabled)." If the installed SGLang build does not have that field, the flag is dropped with no log line, the invariant quietly stops holding, and the failure mode is not a crash — it is a rollout reading weights that were never restored after `sleep`/`wake_up`, i.e. silently corrupted training with a plausible-looking loss curve. That is the hardest possible class of bug to attribute later.

This is pre-existing machinery, so I am not asking you to redesign the filter. But since this PR is what starts routing correctness-critical flags through it, could it validate the ones it explicitly opted into? Something like: after filtering, diff the explicitly-set flags against `allowed` and raise (or at minimum `logger.warning`) for any that were dropped. Failing closed here would be consistent with the rest of this PR — `pp_size>1`, `dp_size>1`, non-divisible `ep_size`, and incomplete TP groups all fail closed rather than silently doing the wrong thing, and this is the same category of problem.

examples/ar/qwen3_cppo_30b_a3b_base_dapo_sglang_full_weight_sync.yaml
```diff
@@ -0,0 +1,175 @@
...
+    temperature: 1.0
+    top_p: 1.0
+    concurrency: 64
+    samples_pre_expanded: true
+    # The trainer sleeps SGLang before restoring FSDP onto the same GPUs.
+    # Preserve the latest rollout weights across rounds that intentionally
+    # skip a full sync (including rollout 0 when baseline eval is disabled).
+    enable_memory_saver: true
+    enable_weights_cpu_backup: true
+    skip_server_warmup: false
```
CjhHa1, line 112: Two questions on this block, both tied to the silent-drop issue flagged on `config.py:327-332`.

**`ep_size: 8` with no `enable_expert_parallel`.** The config adds both fields, but this recipe only sets `ep_size`. In several SGLang releases expert parallelism is gated on `enable_expert_parallel`, which is also what derives/overrides `ep_size` in `ServerArgs.__post_init__` — setting `ep_size` alone can leave EP off while the value is accepted. The Test Plan reports "SGLang initialized all ranks TP0-TP7 / EP0-EP7", but rank labels appearing in logs does not by itself prove the MoE layers were sharded expert-wise rather than tensor-wise. Could you confirm against the SGLang version you ran, and set `enable_expert_parallel: true` explicitly if it is required? I could not verify locally — no `sglang` in my environment.

**`enable_weights_cpu_backup: true` as a stated invariant.** The comment above this block is the load-bearing claim about weights surviving rounds that skip a full sync. If the flag does not exist in the installed build it is dropped silently, so the comment would describe behaviour that is not actually in effect. Worth asserting the flag survived rather than documenting it as guaranteed.

Minor, on the same block: given the trainer now calls `wake_up(tags=["weights"])` before the sync and a full `wake_up()` after `backend.offload()`, `mem_fraction_static: 0.15` is doing a lot of load-bearing work for a 30B MoE on H20s. A one-line comment on how that number was arrived at would help the next person who changes `tp_size` or the model and wonders whether 0.15 was measured or guessed.

CjhHa1 (commented):
Reviewed the full diff (16 files, +1469/-157) against `main` and against the intermediate commits.

The core design holds up: hosting one SGLang engine per TP group on `tp_rank==0` while the other TP ranks stay as no-op shells that still participate in the train collectives lines up cleanly with the existing `_collect_dp_merge` filter (`tp_rank == 0 and is_pipeline_last_stage and sp_rank == 0`), and `RankInfo` already carried the tp/pp/ep fields, so the layout change grafts on without touching the dispatch contract.

I have two blocking concerns, both left as inline comments:

1. **The merge commit `f7d1f8e` appears to have dropped the entire content of three of the seven commits on this branch** — the CUDA-driver-preload / torch-memory-saver compatibility layer. The Test Plan explicitly claims that layer was validated. Details inline on `backends/base.py`.
2. **Newly added, load-bearing SGLang flags travel through a filter that drops unknown keys silently.** `enable_weights_cpu_backup` is the dangerous one: if the installed SGLang lacks the field, the recipe's stated invariant breaks with no error. Details inline on `config.py` and the recipe.

I also have a set of non-blocking findings (an unrelated `logprob_start_len` removal in the text adapter, a dead branch in the `ar.py` bootstrap recovery ladder, an unimplemented `weight_payload_fanout` hook, an incorrect `fanout > 1` non-SGLang payload branch, the per-expert MoE collective count, and the new `skip_lm_head` behaviour change for all tied-embedding models). Happy to write those up separately so this thread stays focused on what blocks the merge.

One process note that cuts across both blocking items: this branch adds 1469 lines and zero tests, while several of the new code paths are pure functions that need no GPU — `_parallel_shape_from_init_kwargs`, `_build_rank_infos`, `_build_tp_visible_device_map`, `_normalize_cuda_visible_devices`, `_pack_qwen_moe_expert_keys`, `_iter_rollout_tensors`, and the new `SGLangEngineConfig.__post_init__` requirements. "25 existing tests passed" establishes no regression, but none of the new topology validation or the MoE packing fail-closed behaviour is covered, and those are exactly the paths the PR description highlights as failing closed rather than silently taking a wrong path.

leviking98z-rgb:
## Maintainer TP8 validation

I tested the previous head (`74c7e02`) on a single node with 8× NVIDIA H20 96 GB GPUs, using the repository-pinned SGLang `0.5.12.post1`, PyTorch `2.11.0+cu130`, and Qwen3-30B-A3B-Base. I found two runtime blockers and pushed fixes in `f76dbfd`:

1. The TP8 recipe explicitly passed `enable_expert_parallel`, but the pinned SGLang `ServerArgs` has no such field, so UniRL correctly failed closed before boot:

   ```text
   RuntimeError: SGLang HTTP backend cannot apply required ServerArgs fields: ['enable_expert_parallel']
   ```

   The recipe now uses `ep_size: 8` as the EP switch. The live topology confirmed SGLang ranks `TP0..TP7 / EP0..EP7`.

2. The spawned SGLang process inherited the launcher's process group, while `HTTPBackend.shutdown()` tears the server tree down with `killpg()`. On any downstream failure, cleanup therefore also SIGTERMed the Ray worker/training driver and masked the original exception. The SGLang child now starts a new session with `os.setsid()`, with a regression test for the launch ordering.

### TP8/EP8 end-to-end result

Base recipe: `examples/ar/qwen3_cppo_30b_a3b_base_dapo_sglang_full_weight_sync.yaml`

Smoke overrides: 2 rollouts, 8 prompts/rollout, 2 samples/prompt, batch size 8, horizon/max-new-tokens 32, one optimizer update per batch, rollout concurrency 1, and 512 MiB full-weight buckets.

- Train/FSDP layout: world=8, DP=8.
- SGLang rollout layout: world=8, DP=1, TP=8, EP=8.
- TensorWeightSync layout: world=8, DP=1, TP=8, EP=8.
- Settled SGLang args included `tp_size=8`, `ep_size=8`, `enable_memory_saver=True`, and `enable_weights_cpu_backup=True`.
- Rollout 1 completed a real non-zero update:

  ```text
  rollout 1/2  reward=0.5625  loss=0.0230  gn=3.7858  lr=1.00e-06  ratio=0.9919  |Δlogp|=2.94e-02
  ```

- The subsequent full-weight sync completed **114/114** `/update_weights_from_tensor` buckets with HTTP 200, after that non-zero update.
- Memory release/resume completed successfully, then all second-round generation requests returned HTTP 200.
- Rollout 2 completed:

  ```text
  rollout 2/2  reward=0.4375  loss=0.0510  gn=4.1184  lr=1.00e-06  ratio=0.9838  |Δlogp|=7.04e-02
  ```

- The job exited 0 and released all training GPU memory.

The final closed-loop rerun kept normal CUDA graphs enabled but disabled piecewise CUDA graphs to shorten repeated startup. Separately, the recipe-default standard + piecewise graph capture completed on all eight TP ranks in two setup runs, so the final override was not hiding a graph-capture failure.

### Other validation

- `python -m pytest -q tests` — **8 passed**.
- Ruff check and format check on all changed Python files — passed.
- `python scripts/check_recipe_targets.py` — **2303** UniRL `_target_` paths resolved.
- `git diff --check` — passed.
- GitHub lint, PR-body/title, status-label, and open-source scan checks — all passed at `f76dbfd`.

The test node's host driver needed an external CUDA 13 forward-compat preload for the repository's cu130 environment. That validation-only shim was not committed and does not alter the TP/EP or weight-sync implementation.

This is strong coverage of topology, SGLang lifecycle, memory handoff, and full-weight transport after an actual optimizer step. It is intentionally not a claim of long-run curve parity. A train-side-vs-SGLang comparison is still useful as a longer regression, but with stochastic sampling the curves should be statistically comparable across seeds, not point-by-point identical.

leviking98z-rgb:
## Follow-up: numerical correctness / stale-weight A/B

I added a numerical validation pass on the current head (`f76dbfd`), beyond the earlier topology/lifecycle smoke test.

### Method

For every generated token, I compared the rollout-time log-probability with a train-side teacher-forced replay at the weights that should have generated that rollout:

- `|Δlogp| = |logp_train_replay - logp_rollout|`
- K3 = `exp(log_r) - log_r - 1`, where `log_r = logp_train_replay - logp_rollout`

The reported mean/max values below were all-reduced over all 8 FSDP data-parallel ranks and all generated tokens. I set `stack.micro_batch_size=2` for this short workload (two samples per rank), so each rank had exactly one replay micro: the mean is token-weighted globally and the max is the true global maximum rather than a per-micro aggregate. I also ran the recipe-default `micro_batch_size=1`; it exited 0 and gave the same conclusion, while the table uses the exact-global rerun. The all-reduce/logging instrumentation was validation-only and has been reverted; the PR worktree is clean at `f76dbfd`.

Environment and workload: 8× H20 96 GB, Qwen3-30B-A3B-Base, repository-pinned SGLang `0.5.12.post1`, SGLang TP8/EP8, 8 prompts × 2 samples, max 32 generated tokens, 2 rollouts, one optimizer step per rollout, fixed SGLang seed.

### Intended recipe learning rate (`1e-6`)

| Rollout engine | Point | global mean `|Δlogp|` | global max `|Δlogp|` | global mean K3 | global max K3 |
|---|---:|---:|---:|---:|---:|
| SGLang TP8/EP8 + full sync | before first update | 0.0637 | 1.12 | 0.00975 | 0.946 |
| SGLang TP8/EP8 + full sync | after update + 114/114 synced buckets | 0.0761 | 0.987 | 0.00847 | 0.360 |
| pure train-side rollout control | rollout 1 | 0.0795 | 1.28 | 0.0159 | 1.33 |
| pure train-side rollout control | rollout 2 | 0.0794 | 0.923 | 0.0107 | 0.594 |

Both jobs exited 0. The TP8 post-sync mean and maximum remain in the same numerical range as its pre-sync baseline and the pure train-side autoregressive-vs-replay control. The train-side control samples are stochastic and not token-identical to the SGLang samples; they calibrate the expected engine-internal generation/replay gap. This is **not** evidence of bitwise identity or literally zero precision loss, but it is evidence against a mean- or tail-level regression in the intended two-step workload.

The pure train-side FSDP autoregressive control otherwise hit an unrelated native cuDNN-SDPA worker crash. Its successful control run disabled cuDNN SDP plus activation/gradient checkpointing for that control only. No such workaround was used by the SGLang TP8 runs, and none was committed.

### Fault injection: synchronized vs deliberately stale weights

To distinguish a real update from “the HTTP calls returned 200 but SGLang kept old weights”, I amplified only the learning rate to `1e-4` and ran a controlled A/B:

- **sync:** normal `weight_sync_interval=1`, 114/114 update requests returned HTTP 200.
- **stale:** `weight_sync_interval=999`, zero update requests.
- All other inputs/configuration were identical.
- The first-rollout dumps were byte-identical in the two runs (SHA-256 `4f9c976cfa5eaaa5aa31306c319b1185061e654989b6f471d48c20a8c6c80438`) and their first-step metrics were identical, so the A/B baseline is reproducible.

| Second rollout | global mean `|Δlogp|` | global max `|Δlogp|` | global mean K3 | global max K3 |
|---|---:|---:|---:|---:|
| normal TP8 full sync | 0.00765 | 0.251 | 0.000402 | 0.0289 |
| deliberately no sync | 1.92 | 16.7 | 1.50 | 15.7 |

Normal sync reduced the stale-weight global mean `|Δlogp|` by approximately **251×** and mean K3 by approximately **3,731×**. Both jobs exited 0. This rules out the full-weight update being a no-op and shows that the synced SGLang policy tracks the updated train policy materially better than stale weights.

### Interpretation

- The full-weight TP8/EP8 sync is numerically effective after a real optimizer update.
- At the recipe learning rate, rollout/replay mean error does not show a catastrophic precision regression after sync.
- I would not claim “zero precision loss” from two rollouts: long-horizon learning behavior still needs a larger multi-seed test with global p95/p99 gates.
- SGLang and train-side reward curves should not be required to match point-by-point. Their samplers consume RNG differently, so even equal policies can generate different completions. For a long-run regression, compare reward mean/variance and KL/log-prob distributions over multiple fixed seeds. For this single-step sync question, replaying the **same SGLang-generated tokens** on the train model is the more direct correctness test.

leviking98z-rgb:
## Follow-up: four review findings fixed

Pushed [`374d6c89`](https://github.com/Tencent-Hunyuan/UniRL/commit/374d6c89bde26cc6c9f4e14e252ce407a43c832c).

1. **Train-side VeOmni EP expert export** — packed Qwen MoE expert tensors now stay fused until `FullWeightSync._iter_full_tensors_ep` gathers the rank-local `[E/ep, ...]` blocks and assigns global expert IDs. Ordinary non-EP Qwen export still uses the existing per-expert unpacking path.
2. **SGLang boot/shutdown cleanup** — a failed health check now reaps the started SRT process tree, safely handles the pre-`setsid()` race, still reaches an orphaned group after its session leader exits, and escalates SIGTERM to SIGKILL on timeout.
3. **TP rank-local serialization failure** — every rank now enters the same `all_gather_object` with either a payload or a serialized error, so all ranks receive the same failing-rank error instead of successful peers hanging in the collective.
4. **AsyncAR TP divisibility** — prompt divisibility is now checked against rollout **DP size**, derived from the same SP/TP/PP layout resolver used by `Handle`; e.g. 8 rollout GPUs with TP8 correctly means DP1, not DP8.

### Validation at this head

- `PYTHONPATH=. /root/.venvs/unirl-pr-review/bin/python -m pytest -q tests` — **16 passed**, with one existing PyTorch allocator deprecation warning.
- Ruff check and format check on all 8 changed/new Python files — passed.
- `/root/.venvs/unirl-pr-review/bin/python -m compileall -q unirl tests` — passed.
- `PYTHONPATH=. /root/.venvs/unirl-pr-review/bin/python scripts/check_recipe_targets.py` — **2303** UniRL `_target_` paths resolve.
- `git diff --check` — passed.
- `SKIP=no-commit-to-branch pre-commit run --all-files --show-diff-on-failure` — every configured hook passed.
- GitHub checks at `374d6c89` — **5/5 successful** (lint, PR body, title, status label, open-source scan).

### GPU/numerical context

The earlier [8×H20 TP8/EP8 end-to-end validation](https://github.com/Tencent-Hunyuan/UniRL/pull/226#issuecomment-5145126493) and [same-token replay + stale-weight A/B numerical validation](https://github.com/Tencent-Hunyuan/UniRL/pull/226#issuecomment-5146016622) were run at `f76dbfd` and remain the GPU evidence for the normal SGLang TP8/EP8 full-weight transport path. That workload was not rerun for this follow-up commit: the SGLang success-path transport/numerics are unchanged. The newly corrected **train-side VeOmni EP** export branch is covered here by a focused key/layout regression test, not by a fresh real-GPU train-side-EP run.

This was an AI-assisted follow-up; I reviewed the full staged diff and checked the related open PRs before publishing it.

leviking98z-rgb:
## Follow-up: test files removed

Per maintainer direction, pushed [`5e116e5e`](https://github.com/Tencent-Hunyuan/UniRL/commit/5e116e5ecc30c17a7bb5cc917e51c317e7b47f07) removing all five `tests/` files introduced on this PR. The final PR diff against `main` now contains **no `tests/` paths**; the four source fixes in `374d6c89` are unchanged.

Validation after the cleanup:

- Ruff check and format check on the four changed source files — passed.
- `/root/.venvs/unirl-pr-review/bin/python -m compileall -q unirl` — passed.
- `PYTHONPATH=. /root/.venvs/unirl-pr-review/bin/python scripts/check_recipe_targets.py` — **2303** UniRL `_target_` paths resolve.
- `git diff --check` — passed.
- `SKIP=no-commit-to-branch pre-commit run --all-files --show-diff-on-failure` — every configured hook passed.
- GitHub checks at `5e116e5e` — **5/5 successful**.

The **16 passed** result in the preceding comment was collected at `374d6c89` before those validation-only regression files were removed; the tests are intentionally not retained in the PR.

unirl/rollout/engine/sglang/backends/base.py
leviking98z-rgb: Thanks — I confirmed this was a deliberate supersede, not an accidental conflict resolution. I updated the PR Summary, Test Plan, Compatibility/Risk, and Reviewer Notes so the current tree and its support boundary are explicit.

I also reran the exact current head `5e116e5ecc30c17a7bb5cc917e51c317e7b47f07` on 8×H20 with SGLang `0.5.12.post1` and PyTorch `2.11.0+cu130`. The run initialized `TP0..TP7 / EP0..EP7`, settled `enable_memory_saver=True` and `enable_weights_cpu_backup=True`, completed two rollouts, completed 114/114 full-weight update requests with HTTP 200 after the first optimizer step, completed release/resume, and exited 0.

The important caveat is now stated rather than hidden: the torch-memory-saver interaction has not disappeared on a mismatched driver/runtime host. This validation node has driver 535, so its launcher supplied CUDA 13 forward compatibility and a checkout-external preload-preservation shim. This PR neither commits nor claims that shim. The supported boundary is either a driver natively compatible with the pinned CUDA runtime, or an image/launcher that preserves the compatible driver through the torch-memory-saver scheduler spawn and gives Ray/SGLang a consistent library path. The PR intentionally does not monkey-patch installed SGLang/TMS internals.

That removes the stale Test Plan claim, documents exactly what was dropped and why, and adds the requested current-head 8-GPU evidence.

unirl/rollout/engine/sglang/backends/http.py
leviking98z-rgb: Addressed on the current tree. `_launch_server_with_env` now documents the ordering invariant immediately before applying the override: it must execute before importing SGLang or touching `torch.cuda`, because CUDA visibility can be cached after initialization. The target performs `setsid()`, applies `os.environ.update(env_overrides)`, and only then imports `launch_server`.

I chose the explicit invariant comment instead of an assertion because PyTorch exposes no stable public predicate for “CUDA has never been initialized”; importing or probing `torch.cuda` solely to assert it would itself enlarge the import boundary that must remain CUDA-free.

The current-head 8×H20 rerun exercised this path and initialized exactly `TP0..TP7 / EP0..EP7`. The separate forward-compatibility/TMS limitation is now documented in the PR body and in my reply on the `base.py` thread rather than being implied to be handled here.

unirl/rollout/engine/sglang/config.py
leviking98z-rgb: Fixed in `193f1efe` and retained at the current head. `SGLangEngineConfig.server_intent()` records every explicitly requested load-bearing typed field in `_unirl_required_server_args`; `_filter_server_args_or_raise()` then raises before boot if any such field is absent from the installed `ServerArgs`. Both HTTP and native backends use that filter. Optional `engine_kwargs` escape-hatch keys may still drop, but the typed correctness/memory-lifecycle fields no longer can.

The required set covers `ep_size`, `enable_expert_parallel`, `enable_memory_saver`, `enable_weights_cpu_backup`, and `skip_server_warmup`. The current-head 8×H20 run also confirmed that the installed runtime settled `ep_size=8`, `enable_memory_saver=True`, `enable_weights_cpu_backup=True`, and `skip_server_warmup=False` rather than dropping them.

examples/ar/qwen3_cppo_30b_a3b_base_dapo_sglang_full_weight_sync.yaml
leviking98z-rgb: All three points are addressed.

- The pinned SGLang `0.5.12.post1` `ServerArgs` used here has `ep_size` but no separate `enable_expert_parallel` field. The recipe therefore uses `ep_size: 8` as the switch. UniRL still exposes the optional boolean for other builds, but now fails closed if it is explicitly requested and unsupported. The exact current-head rerun settled `ep_size=8` and initialized all eight ranks as `TP0..TP7 / EP0..EP7`.
- `enable_weights_cpu_backup` is part of the required-field fail-closed set, and the same run printed `enable_weights_cpu_backup=True` in settled `ServerArgs`; it was not silently dropped. The run then completed release/resume, a real optimizer step, 114/114 weight updates, and the second rollout.
- The recipe now comments that `mem_fraction_static: 0.15` is measured headroom for FSDP offload/onload plus staged SGLang wake on this 8-GPU 30B MoE layout, and must be retuned when changing the model or TP layout.

leviking98z-rgb:
## Current-head TP8/EP8 review-closure validation

Re-ran the exact current head `5e116e5ecc30c17a7bb5cc917e51c317e7b47f07` on one 8× NVIDIA H20 96 GB node with SGLang `0.5.12.post1`, PyTorch `2.11.0+cu130`, and Qwen3-30B-A3B-Base.

- SGLang settled `tp_size=8`, `ep_size=8`, `enable_memory_saver=True`, `enable_weights_cpu_backup=True`, and `skip_server_warmup=False`.
- All scheduler ranks initialized as `TP0..TP7 / EP0..EP7`; UniRL reported TensorWeightSync `world=8, dp=1, tp=8, ep=8`.
- Rollout 1 completed a non-zero update: `reward=0.6875`, `loss=0.0619`, `gn=3.2042`, `lr=1e-6`.
- The subsequent full-weight sync completed 114/114 `/update_weights_from_tensor` requests with HTTP 200.
- Memory release/resume and second-round generation completed; rollout 2 finished with `reward=0.5625`, `loss=-0.0256`, `gn=5.3634`, `lr=1e-6`.
- The job exited 0, node cleanup released training memory, and the cluster claim was released.

Compatibility caveat: this node has driver 535 while the pinned environment is CUDA 13.0, so the validation launcher supplied `cuda-compat-13` plus a checkout-external torch-memory-saver preload-preservation shim. No part of that shim is committed. The PR body now states that a native-compatible driver or an image/launcher providing this property is a deployment prerequisite; the PR does not claim to patch SGLang/TMS internals.

I updated the PR Test Plan and Compatibility/Risk sections to the current head, removed stale references to deleted test files, replied to all four review comments with the corresponding code/runtime evidence, and resolved all four threads. The final PR diff still contains no `tests/` paths.

leviking98z-rgb:
## Architecture / change map

The final PR diff is one new recipe plus changes to 15 existing files (`+1711/-194`). The central change is to remove the former rollout-TP=1 assumption across placement, SGLang boot, full-weight transport, Qwen MoE conversion, and trainer memory ownership.

```mermaid
flowchart LR
    Recipe["NEW: Qwen3-30B full-weight recipe<br/>TP8 / EP8 / TensorWeightSync"]
    Trainer["CHANGED: ARTrainer<br/>coordinates train/rollout GPU ownership"]
    Handle["CHANGED: Handle + Placement<br/>DP / TP / PP / EP rank layout"]

    subgraph Train["Train side: 8 UniRL workers"]
        Load["CHANGED: Qwen MoE load<br/>split expert keys to packed tensors"]
        FSDP["FSDP full-parameter training<br/>DP=8"]
        Export["CHANGED: Qwen MoE export<br/>packed or EP shard to SGLang names"]
        TensorSync["CHANGED: TensorWeightSync<br/>each TP rank serializes local CUDA IPC"]
    end

    subgraph Rollout["Rollout side"]
        Host["UniRL rollout tp_rank=0<br/>owns the HTTP server"]
        Shell["UniRL rollout tp_rank=1..7<br/>no-op shells"]
        SRT["SGLang scheduler processes<br/>TP0..TP7 / EP0..EP7"]
    end

    Recipe --> Trainer
    Trainer --> Handle
    Handle --> FSDP
    Handle --> Host
    Handle --> Shell

    Load --> FSDP
    FSDP -->|"optimizer update"| Export
    Export --> TensorSync
    TensorSync -->|"all-gather 8 device-local IPC payloads<br/>send in TP-rank order"| Host
    Host --> SRT
    SRT -->|"tokens + rollout log-probs"| Trainer
    Trainer -->|"loss / backward / optimizer step"| FSDP

    NCCL["CHANGED: NCCLWeightSync<br/>TP-aware world size and rank offsets"]
    Separate["Separate rollout slab<br/>SGLang TP groups"]
    Export -. "alternative transport" .-> NCCL
    NCCL -.-> Separate

    IPC["CHANGED: generic IPCWeightSync<br/>TP greater than 1 fails fast"]
    Export -. "unsupported path" .-> IPC
```

`EP8` is internal to the SGLang scheduler group; it does not create another eight UniRL EP workers. Only the UniRL `tp_rank=0` rollout role boots the SGLang server. The other seven rollout roles are no-op shells, while their colocated training/weight-sync ranks still participate in full-tensor gathering and serialize the CUDA IPC handles belonging to their own devices.

### Per-rollout memory ownership

```mermaid
sequenceDiagram
    participant T as ARTrainer
    participant F as FSDP train state
    participant W as TensorWeightSync
    participant S as SGLang TP8/EP8

    T->>S: wake_up(tags=["weights"])
    T->>W: sync updated full weights
    W->>S: TP-ordered payloads for each bucket
    T->>F: offload parameters, gradients, optimizer
    T->>S: full wake_up for KV cache and CUDA graphs
    S-->>T: generate tokens and rollout log-probs
    T->>S: sleep and release rollout memory
    T->>F: onload training state
```

### File-level change summary

| Layer | Main files | Change |
|---|---|---|
| Runnable configuration | `examples/ar/qwen3_cppo_30b_a3b_base_dapo_sglang_full_weight_sync.yaml` | Adds the 8×H20 Qwen3-30B full-weight CPPO recipe with FSDP, TP8/EP8 SGLang, bf16 512 MiB tensor buckets, and colocated memory handoff. |
| Rank and placement model | `distributed/group/handle.py`, `placement.py` | Carries TP/PP/EP through `HandleRef`, builds TP-fast rank layouts, validates node-local complete TP groups, and obtains real Ray CUDA tokens rather than deriving local ordinals from cluster-global device IDs. |
| SGLang configuration and boot | `sglang/config.py`, `engine.py`, `backends/{base,http,native}.py` | Adds TP/PP/EP config, required-`ServerArgs` fail-closed handling, tp-rank-0 server ownership, child-local CUDA visibility, bounded process-tree cleanup, and native TP>1 rejection. |
| Full-weight transport | `weight_sync/full/{tensor,nccl,ipc}.py` | Adds per-device TP tensor payload gathering and synchronized failure propagation; makes NCCL ranks TP-aware; rejects the old generic local-rank-0-only IPC path for TP>1. |
| Qwen MoE load/export | `train/backend/sharded_load.py`, `utils/peft_merge.py` | Packs split Qwen expert checkpoint keys for the training model and exports packed/EP-sharded experts under the per-expert names accepted by SGLang live reload. |
| Trainer lifecycle | `trainer/ar.py`, `trainer/async_ar.py` | Adds staged wake/sync/offload/generate/sleep/onload, bootstrap rollback, rollout-DP-aware divisibility, and TP-aware NCCL target discovery for separate slabs. |
| Sampling compatibility | `sglang/adapters/text.py` | Stops forcing `logprob_start_len=0` in every SGLang request. |

This PR does not add a new RL algorithm. It completes the framework path:

`FSDP full-weight training -> Qwen MoE conversion -> TP8 weight transport -> SGLang TP8/EP8 rollout -> colocated GPU-memory handoff`.

CUDA forward-compatibility for an older host driver plus the pinned CUDA 13 runtime remains an image/launcher prerequisite and is not implemented by this diagrammed code path.

leviking98z-rgb (approved):

PR #227, opened 2026-07-20
feat(rollout): support SGLang TP/EP LoRA sync

NancyFyong:
## Summary

LoRA weight sync for a multi-GPU SGLang rollout: make the adapter hot-load work when the rollout engine runs with TP>1, preserve the optimized TP1 path, and add the recipe that exercises it.

This branch was rebased onto current `main`. Most of what the previous description claimed has since landed on `main` through other PRs (rollout `RankInfo` TP/PP/EP metadata and the handle-layer TP kwargs, the `tp_rank>0` no-op shell engine, `pp_size>1` fail-closed config validation, native-backend TP fail-closed, `CUDA_VISIBLE_DEVICES` normalization). What remains here is exactly three changes:

- **Select LoRA tensor serialization by TP topology.** For TP1, keep SGLang's optimized `MultiprocessingSerializer`: its `ForkingPickler` payload contains shared-storage handles, the spawned server inherits the sender's multiprocessing authkey, and an end-to-end A/B measured a 2.09x warm hot-load speedup. For TP>1, serialize the CPU tensors as base64 pickle bytes: SGLang broadcasts one request to every TP scheduler, but each `resource_sharer` file descriptor is one-shot, so a shared multiprocessing payload lets one rank consume the handles and makes the others fail with `EOFError`. The plain-pickle payload contains no descriptors and is safe to broadcast.
- **Push the adapter only from the TP leader** in `LocalLoraWeightSync.sync`. Every rank still extracts — that walk is a train-mesh collective — but only `tp_rank == 0` hosts a SGLang server, and the engine already drops the push on the other ranks, so those ranks no longer log a push that never happened.
- **Add `examples/ar/qwen3_cppo_30b_a3b_base_dapo_sglang_lora.yaml`**, the single-node 8xH20 Qwen3-30B-A3B CPPO recipe that drives a rank-16 adapter through the SGLang HTTP backend at TP=8 / EP=8. `main`'s full-weight sibling recipe already refers to this file by name, so this also fixes that dangling reference.

Out of scope: full-weight tensor/NCCL TP sync, checkpoint-engine IPC sync, native-backend TP (fail-closed on `main`), and diffusion / `sglang_diffusion` TP>1.

## Related Issue

- RFC: #217 — Multi-GPU SGLang Rollouts with TP, EP, and Checkpoint-Engine Weight Sync
- #191 — the original unsplit rollout branch; this is the focused LoRA split.

## Test Plan

No test files: this repository has no test suite, so validation is the lint gates, targeted serializer checks, and the GPU A/B below.

- `pre-commit run --files <all changed files>` — full suite green, including `check-docstring-lines`, `check-recipe-targets`, `check-core-dependencies`, `ruff check`, and `ruff format`.
- `python lint/check_recipe_targets.py examples/ar/qwen3_cppo_30b_a3b_base_dapo_sglang_lora.yaml` — all recipe `_target_` paths resolve.
- Targeted CPU dispatch check with a fake SGLang serializer: TP1 called `MultiprocessingSerializer.serialize(..., output_str=True)`; TP8 did not call it and round-tripped the base64 pickle.
- Serializer microbenchmark against the pinned SGLang 0.5.12.post1 package, 27 MiB CPU bf16 payload:
  - TP1 multiprocessing: 1/1 consumers passed, 1.9 KiB payload, 16.41 ms serialize / 3.74 ms deserialize.
  - TP1 plain pickle: 1/1 passed, 36.0 MiB payload, 65.66 ms serialize / 28.54 ms deserialize.
  - TP8 multiprocessing: 1/8 passed; seven consumers raised `EOFError` and the resource-sharer server logged seven `KeyError: 1` failures.
  - TP8 plain pickle: 8/8 passed, 39.57 ms mean deserialize.
- Real SGLang HTTP hot-load A/B on one 8xH20 node, Qwen3-0.6B, rank-16 q/k/v/o adapter (224 bf16 tensors, 8.75 MiB). The server was spawned by the serializer process, matching UniRL's multiprocessing authkey topology:
  - TP1 multiprocessing: 3/3 load + one-token generation + unload passed; warm serialization + hot-load was 46.31 ms.
  - TP1 plain pickle: 3/3 passed; warm serialization + hot-load was 96.89 ms.
  - TP8 multiprocessing: 0/1; seven schedulers raised `EOFError`, the HTTP connection closed, and the engine exited after `SIGQUIT`.
  - TP8 plain pickle: 3/3 load + generation + unload passed; warm serialization + hot-load was 119.52 ms.
- The end-to-end server image had SGLang 0.5.10.post1; the 0.5.12 package could run the exact serializer microbenchmark but its server boot was blocked by that image's older `sglang-kernel`. The relevant `ForkingPickler` / `SafeUnpickler` / TP request-broadcast implementation is the same in both versions and was checked directly.
- The 30B training curve below was not re-run after the rebase. Post-review recipe edits are `cuda_graph_max_bs: 64` (new cap), moving `enable_memory_saver` / `enable_weights_cpu_backup` / `skip_server_warmup` from `engine_kwargs` to typed config fields (identical server intent), and dropping two `bundle.config` keys that nothing reads.

## Experiment Snapshot

LoRA path only; full-weight rollout validation belongs to the full-weight sync PR.

Qwen3-30B-A3B, CPPO, single 8-GPU node, SGLang HTTP backend at TP=8 / EP=8, rank-16 adapters on `q_proj` / `k_proj` / `v_proj` / `o_proj`, LR `1e-4`, 8 prompts x 16 samples/prompt, rollout concurrency 64, `max_new_tokens=4096`, `temperature=1.0`, `top_p=1.0`, `top_k=0`, eval every 10 rollouts with avg@8.

> **LoRA:** train/eval reward curve.

LoRA train/eval reward curve

## Compatibility / Risk

The serializer choice is topology-aware:

- TP1 uses SGLang's multiprocessing serializer. This requires the receiver to inherit the sender's multiprocessing authkey; UniRL satisfies that because the HTTP server is spawned by the same Worker process, and the native Engine spawns its schedulers from that process. An unrelated external HTTP client cannot use this path.
- TP>1 uses plain pickle + base64. This copies the adapter into the request (tens of MB for rank-16 on a 30B model, plus ~1.33x base64 expansion) instead of sending shared-memory handles. The measured TP8 Qwen3-0.6B update was 119.52 ms on loopback; correctness takes priority because one shared descriptor payload cannot be broadcast safely.

The TP-leader guard is a no-op for `tp_size=1` (every rank is its own leader) and for multi-DP layouts each group's leader still pushes to its own server. `pp_size>1` remains fail-closed on `main`. `verify` stays off for SGLang because the SGLang engine exposes no checksum read-back API — that is vLLM-Omni-only. Diffusion behavior is unchanged.

Recipe risk: `mem_fraction_static: 0.5` is the value the run above used; retune it when changing the model, the TP layout, or the FSDP offload settings.

## Reviewer Notes

The branch was force-pushed as a rebase, so its history was reduced from the previous 20 commits. The old history contained six CUDA / `torch-memory-saver` preload patches that a later commit removed wholesale after @celve's review; replaying them would have meant resolving conflicts in code that the same branch deletes two commits later. Author attribution is preserved.

The earlier conflicts against `main` were entirely #316 (one-line docstrings, enforced by `check-docstring-lines` with no allowlist) colliding with this branch's multi-line docstrings. Resolved by adopting `main`'s one-line form and moving the rationale into the nearest README `## Gotchas`, per the rule that this prose does not belong in a `#` comment block either. That also let one stale gotchas bullet be corrected: with a TP>1 rollout engine the push owner is each group's `tp_rank == 0`, not global rank 0.

Two independent AI reviews (GPT-5.6, Grok-4.6) found no blocking defect and agreed on the missing CUDA-graph cap and unread `bundle.config` LoRA keys, both fixed here. They initially split on whether to gate the serializer on `tp_size`; the GPU A/B above resolved that question empirically in favor of the topology-aware implementation. AI assistance was used for the rebase and this description; the diff, A/B harness, raw results, and validation were checked before publication.

## Checklist


zzhuoxin1508:
`async_ar.py` now passes two new kwargs (`tp_size`, `pp_size`) when connecting weight sync, but the NCCL weight sync's `connect()` was never updated to accept them. Since async training hard-requires NCCL weight sync, every async run now dies at startup with a `TypeError` — including existing single-GPU-per-engine configs that worked before this PR.

A comment added in this PR says connect already rejects pp_size>1, but no such code exists on this branch — looks like the nccl.py part of this change was never committed.

Separately, connect still assigns NCCL ranks assuming one GPU per rollout target; with TP engines spanning multiple GPUs, two TP groups would get overlapping ranks and the rendezvous hangs.

NancyFyong:
Thanks for catching this. Fixed in 9320c23: `NCCLWeightSync.connect()` now accepts `tp_size`/`pp_size`, fails closed for `pp_size > 1`, and assigns each rollout TP engine a disjoint NCCL rank block via `rank_offset = i * tp_size + 1`. PR checks are passing.

zzhuoxin1508:
Two smaller issues, both from changes in this PR, neither blocking:

`engine.py` adds a `tp_device_ids` init kwarg described as compatibility for "direct callers written against the first TP draft" — but that draft was never merged, so no such callers exist. Grepping the repo, the parameter appears only in this one function: nobody passes it, the mutual-exclusion check can never fire, the elif branch is unreachable, and the stored attribute is never read.

The new pickle+base64 serialization block in `set_lora` is duplicated verbatim between the HTTP and native backends — 11 identical lines including the comment explaining the ForkingPickler race — so future edits to one copy will silently miss the other. One symptom already: the native docstring above it still says "same MultiprocessingSerializer call", describing the implementation this PR replaced.

NancyFyong:
Thanks, addressed both non-blocking cleanups in 8349264: removed the unused `tp_device_ids` compatibility path, deduplicated LoRA tensor serialization into a shared SGLang backend helper, and updated the native backend docstring. The current PR checks are passing.

zzhuoxin1508 (dismissed):
LGTM

celve (dismissed):
I’m concerned this introduces too much fragile runtime patching for the LoRA TP change. `_normalize_cuda_visible_devices` and `_serialize_lora_tensors` seem reasonable, but the preload/environment helpers mutate process-global state, monkey-patch `torch-memory-saver`, and depend on SGLang internals. Could we isolate this in a dedicated launcher process or fix it upstream—and temporarily limit TP support to the HTTP backend if necessary? Without that isolation and focused tests, this will be difficult to maintain across dependency upgrades.

NancyFyong:
> I’m concerned this introduces too much fragile runtime patching for the LoRA TP change. `_normalize_cuda_visible_devices` and `_serialize_lora_tensors` seem reasonable, but the preload/environment helpers mutate process-global state, monkey-patch `torch-memory-saver`, and depend on SGLang internals. Could we isolate this in a dedicated launcher process or fix it upstream—and temporarily limit TP support to the HTTP backend if necessary? Without that isolation and focused tests, this will be difficult to maintain across dependency upgrades.

 Thanks for the careful review. I agree that the CUDA preload / torch-memory-saver / SGLang scheduler monkey-patching is too fragile for this LoRA TP PR.

  I’ll narrow the scope here: keep only the low-risk CUDA visibility validation and shared LoRA tensor serialization helper, remove the process-global preload/env monkey-
  patching and SGLang-internal scheduler hooks, and scope rollout TP support to the HTTP SGLang backend for now.

  LocalLoraWeightSync will remain unchanged and continue to load adapters through the backend seam (`set_lora`); the difference is that the TP recipe will use the HTTP
  SGLang backend instead of the native in-process backend. Native TP will fail fast with a clear error and can be revisited separately once we have an upstream-supported
  launch path or focused tests.

  I’ll update the recipe and rerun the checks.

nussejzz:
Thanks for your contribution, please fix the yaml comment and the pr description before merge

Totally, LGTM

NancyFyong:
Addressed the reviewer follow-up:

- Updated the YAML recipe comments to describe the SGLang HTTP TP=8 / EP=8 LoRA-sync recipe directly.
- Updated the PR description to remove the stale native TP validation wording and clarify that native SGLang TP is intentionally unsupported in this PR.
- Rebased/merged the branch onto latest upstream main before the final update.

Validation after the update:

- `git diff --check`
- YAML parse check for `examples/ar/qwen3_cppo_30b_a3b_base_dapo_sglang_lora.yaml`
- `python lint/check_recipe_targets.py examples/ar/qwen3_cppo_30b_a3b_base_dapo_sglang_lora.yaml`
- `SKIP=no-commit-to-branch pre-commit run check-yaml --files examples/ar/qwen3_cppo_30b_a3b_base_dapo_sglang_lora.yaml`
- `SKIP=no-commit-to-branch pre-commit run check-recipe-targets --files examples/ar/qwen3_cppo_30b_a3b_base_dapo_sglang_lora.yaml`
- Latest PR checks are all passing: lint, PR body, PR title, status label, and open-source scan.

nussejzz:
LGTM

CjhHa1 (approved):

unirl/rollout/engine/sglang/backends/native.py
```diff
@@ -466,7 +464,7 @@ def set_lora(
     ) -> None:
         """Serialize the LoRA tensor bag and hot-load it on the Engine."""
         self._require_alive("set_lora")
-        serialized = self._rt["MultiprocessingSerializer"].serialize(lora_tensors, output_str=True)
```
celve: Will this introduce performance regression on TP=1?

PR #229, opened 2026-07-20
fix(vllm_omni/lora-sync): fix verify readback manager level and packed tensors.

cboss6:
## Summary
Fix two latent bugs in the LoRA weight-sync verify read-back (BucketedIPCReceiveMixin._diffrl_loaded_lora_checksums) so that RemoteLoraWeightSync(verify=True) produces a trustworthy N/N match signal instead of false negatives. Both are read-only fixes to the check-sum read-back path; no sync/write behavior changes.

### Background
When verify=True, LoRA sync does a checksum reconciliation after every push:

the trainer hashes the LoRA tensors it pushed (the expected set), and
the rollout worker reads back the LoRA tensors actually registered in the vLLM engine and hashes them (_diffrl_loaded_lora_checksums), so the two multisets can be compared.
This verify path is off by default (verify=False), so this read-back had effectively never executed in anger — which is exactly why the two bugs below sat undetected. They only surface the moment someone turns verify=True on (e.g. to debug a train/inference mismatch), and when they do, they manifest as false "verify FAILED" with loaded 0 / a wrong count — actively misleading, because the LoRA was in fact loaded correctly.

Note: this PR is only about making the verify tool itself correct. It is model-agnostic — _diffrl_loaded_lora_checksums is shared by every model that uses RemoteLoraWeightSync verify (HI3 included) — and does not touch any model-specific weight-naming logic.

## Test Plan
- `pytest -q tests/distributed/weight_sync/test_lora_verify.py`

## Checklist

### DIFF
--- a/unirl/rollout/engine/vllm_omni/worker/ipc_receive_mixin.py
+++ b/unirl/rollout/engine/vllm_omni/worker/ipc_receive_mixin.py
@@ -476,10 +476,12 @@ class BucketedIPCReceiveMixin:
     ) -> dict:
         """Full-byte SHA-256 of the worker's loaded LoRA adapter tensors.
 
-        Walks ``lora_manager._registered_adapters[adapter_id].loras`` and
-        hashes ``lora_a`` / ``lora_b`` / ``bias`` / ``embeddings_tensor``
-        when present. ``lora.optimize()`` has already run by this point
-        — ``lora_b`` is post-scaling — so the trainer must apply the
+        Walks the inner ``LoRAModelManager._registered_adapters[adapter_id].loras``
+        and hashes ``lora_a`` / ``lora_b`` / ``bias`` / ``embeddings_tensor``
+        when present. Packed modules store their ``lora_a`` / ``lora_b`` as a
+        list of sub-tensors (one per fused projection), which are hashed
+        per-shard as ``<field>.<i>``. ``lora.optimize()`` has already run by
+        this point — ``lora_b`` is post-scaling — so the trainer must apply the
         matching ``alpha / r`` scaling before hashing for equality. The
         helper :func:`weight_sync.checksum.compute_lora_checksums_post_optimize`
         does that.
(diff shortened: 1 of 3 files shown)


unirl/rollout/engine/vllm_omni/worker/ipc_receive_mixin.py
```diff
@@ -520,6 +527,14 @@ def _diffrl_loaded_lora_checksums(
...
                 if isinstance(t, torch.Tensor):
                     per_field[field] = fingerprint_tensor(t)
+                elif isinstance(t, (list, tuple)):
+                    # Packed modules (``qkv_proj``, ``gate_up_proj``) store one
+                    # sub-tensor per fused projection; ``None`` slots mark
+                    # absent shards. Hash each present shard separately so the
+                    # readback covers the whole fused layer, not just the first.
+                    for i, sub in enumerate(t):
+                        if isinstance(sub, torch.Tensor):
+                            per_field[f"{field}.{i}"] = fingerprint_tensor(sub)
```
leviking98z-rgb, line 537: **Blocking:** these hashes are never consumed by the current verification path. `LoraWeightSyncBase._assert_loaded` collects only the exact keys `lora_a` and `lora_b` (`f["lora_a"] if "lora_a" in f`, and likewise for B). For a packed layer this patch emits only `lora_a.0`, `lora_a.1`, ... / `lora_b.0`, ..., so every shard is skipped and verify still reports `engine loaded 0 / 0`. I reproduced this with a `qkv_proj` entry containing three A/B shards. Please update `_assert_loaded` to flatten both exact and suffixed field keys (or preserve the existing output schema via virtual per-shard layers), and add a packed-readback regression test.

leviking98z-rgb (commented):
Found one blocking correctness issue in the packed-tensor path. The inline comment has the concrete failure mode: the new suffixed checksum keys are not consumed by the downstream verifier, so packed verification still reports 0/0.

leviking98z-rgb (approved):

PR #230, opened 2026-07-20
fix(vllm_omni/adapters): gate shift assertion on needs_sigmas

cboss6:
## Summary
Gate the base adapter's construction-time `shift` assertion on `needs_sigmas`, so that AR-only adapters are no longer rejected for lacking a diffusion-only config field. Diffusion adapters are unaffected.

## Test Plan
No test, reason: no need

## Checklist


leviking98z-rgb (approved):
Reviewed the AR-only sigma-policy guard and initialization ordering. The change preserves diffusion validation, skips schedule construction for needs_sigmas=False adapters, and fails invalid diffusion schedules before backend startup. Targeted local validation passed; no blocking findings.

PR #231, opened 2026-07-20
fix(train/stack): set train mode before update for AR models.

cboss6:
## Summary
HF ``from_pretrained`` returns the model in eval mode and nothing else in the AR path flips it. ``GradientCheckpointingLayer.__call__`` only checkpoints when ``self.gradient_checkpointing and self.training`` — so eval mode silently no-ops the bundle's ``use_gradient_checkpointing=True``, and the differentiable replay forward retains every decoder layer's activations → OOM on long sequences. Put the trainable module in train mode before the update (mirrors ReFLPolicy.sample_and_decode's ``self.backend.model.train()`` on the diffusion path).

## Test Plan
No test, reason: no need

## Checklist

### DIFF
--- a/unirl/train/stack/base.py
+++ b/unirl/train/stack/base.py
@@ -337,10 +337,10 @@ class TrainStack(Remote):
         ``num_updates_per_batch`` optimizer steps run over disjoint updates, and
         ``on_rollout_end`` runs once — see :meth:`_run_updates`.
         """
-        # Put the trainable module in train mode before the update (specifically for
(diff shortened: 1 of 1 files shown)


leviking98z-rgb (approved):
Reviewed the train/eval mode boundary and the auto-merge result against current main. The old-policy anchor is computed in eval mode, gradient-bearing updates re-enter train mode so Hugging Face gradient checkpointing engages, and the change merges cleanly with the current TrainStack loss-weighting/eval_track additions. Targeted merged-tree validation passed; no blocking findings.

PR #232, opened 2026-07-20
feat(opd): DiffusionOPD — on-policy distillation for diffusion models (SD3.5-M, multi-teacher)

KemingWu:
## Summary

Ports *"DiffusionOPD: A Unified Perspective of On-Policy Distillation in Diffusion Models"* (arXiv 2605.15055) onto UniRL: the student rolls out its own trajectory, a frozen teacher LoRA replays that same trajectory, and the loss is the closed-form per-step Gaussian KL between the two transition means — no reward model in the loss. Multiple teachers (PickScore / OCR / GenEval) are distilled one-teacher-per-batch (the paper's MOPD protocol).

> **Note (2026-08-02):** the implementation on this branch was replaced by the maintainers with a core-primitive rework (merge `351efe94`; original commits preserved in history). The original implementation validated the approach live on 2×8 H20 but was structurally broken after its rebase over #214 (`unirl.types.prompts` was removed upstream) and carried a counter-based teacher routing that desynchronized on checkpoint resume. See Reviewer Notes for what was kept.

The feature lands as four independently reusable pieces, each riding an existing contract:

- **Frozen teacher adapters as a backend primitive** — `backend.lora_cfg.frozen_adapters: [{name, path}]` (`unirl/train/lora.py::inject_frozen_adapter`): injected **pre-FSDP-wrap** so FSDP shards them and LoRA checkpoints stay save/load-symmetric; geometry comes from each teacher's own `adapter_config.json`; the optimizer and rollout weight-sync never see them. `adapter_active()` is the supported routing primitive (re-asserts the requires_grad split peft's `set_adapter` breaks under FSDP2).
- **Algorithm as a pure loss** (`unirl/algorithms/diffusionopd.py`): `prepare_part` picks the teacher named by the batch's `metadata["domain"]` and freezes teacher means on the pre-declared `segment.sde_means` CONCAT field; `compute_loss_and_backward` reuses `_gaussian_kl_div` + `_transition_sigma` from `algorithms/base.py`. `add_kl_coefficient=false` ⇒ ODE mean-matching (paper's headline mode), `true` ⇒ SDE KL with σ from `strategy.transition_std`. Declares `requires_advantages=False`.
- **Data-driven teacher routing** (`unirl/data/data_source.py::MultiDomainRLDataSource`): one dataset per domain, single-domain batches, every row stamped `metadata["domain"]`. Routing travels **with the data**, so resume fast-forwarding can never desynchronize prompts from teachers; eval splits the budget across all domains in one pass. The trainer projects root metadata onto the train Part; `requires_advantages=False` algorithms skip advantage derivation.
- **Per-domain reward monitoring** (`unirl/reward/local/per_domain.py::PerDomainRewardScorer`): dispatches rows to per-domain inner scorers (local or remote), surfaces `component_rewards["<domain>"]` so wandb draws one curve per teacher. Monitoring-only for OPD.

- **MOPD task-cycle gradient accumulation** (`accumulate_rollouts: M`, review follow-up): the paper takes ONE optimizer step per full teacher cycle (official config: `gradient_accumulation_steps = num_batches_per_epoch = len(teachers)`). The window is **one `train_track` call**: a bare `Part` is the plain path (every existing caller unchanged), a TUPLE of Parts is a window — zero once, prepare and backward every part at `1/M` on the never-stepped weights (backward scale only; logged losses stay raw), ZeRO-2 reduce-scatter deferred to the final part's last micro, step once. `train_track` keeps its one-call-one-step invariant, mid-window checkpoints are structurally impossible (no cross-call state, no trainer phase arithmetic), the wandb `train/step` axis counts real optimizer steps, and the per-domain `distill_loss_*` keys land side by side in each step's point. Tuple, not list: DP_SCATTER recurses tuples element-wise while lists are row-sliced as data.

Recipe: `examples/diffusion/sd3/sd3_opd_multi_teacher.yaml` (SD3.5-M + 3 teacher LoRAs + `accumulate_rollouts: 3` + local PickScore/OCR + remote classical GenEval via `unirl-reward-service/configs/geneval_service.yaml` — the mmdet 2.x stack needs py≤3.10 and stays out-of-process). The OCR (19652/1017) and GenEval (33199/553) prompt sets now ship in-repo under `datasets/` (imported from X-GenGroup/Flow-Factory, Apache-2.0; geneval `include`/`exclude` normalized from JSON-encoded strings to real JSON values), so the recipe starts out of the box. Reward is fully optional for `requires_advantages=False` algorithms — delete the recipe's `reward:` block to train with no reward model (RL algorithms and `eval_interval>0` still hard-require one at construction).

## Related Issue

N/A

## Test Plan

Lifecycle + metrics round (`24b577a7`) — GPU (8×H20) with the reward stack enabled, plus the resume A/B protocol, all on one tree:

- **Per-domain curves and phase timing, from the offline wandb run** (3 windows over 9 rollouts, `accumulate_rollouts: 3`):

  | `_step` | num_samples / groups | `reward_{pickscore,ocr,geneval}_mean` | `distill_loss_{pickscore,ocr,geneval}` | `train/step` | `perf/{generate,reward,train}_time_s` |
  |---|---|---|---|---|---|
  | 0–2 | 24 / 24 | 0.8397 / 0.0000 / 0.8821 | 0.00080 / 0.00032 / 0.00078 | 1 | 3.7 / 3.7 / 11.3 |
  | 3–5 | 24 / 24 | 0.7921 / 0.0000 / 0.8830 | 0.00084 / 0.00028 / 0.00107 | 2 | 2.7 / 0.3 / 7.9 |
  | 6–8 | 24 / 24 | 0.8408 / 0.0000 / 0.8520 | 0.00060 / 0.00029 / 0.00044 | 3 | 2.7 / 0.3 / 7.8 |

  All three reward domains and all three per-teacher loss curves carry finite values in every window point; `num_samples`/`group_count` = 24 = 3 rollouts × 8 prompts confirms the pooling; `train/step` = 1,2,3 confirms one optimizer step per window; no non-finite entry anywhere. (`reward_ocr` is 0.0 because PaddleOCR crashes on this box — a known env issue, unrelated to the PR: the scorer fail-softs to worst-case distance. The point of the check is that the key is present and finite.)
- **Resume A/B on the same tree**: `optimizer_step_count` 2→3 at checkpoints 6/9 in both runs; all four adapters × 382 tensors present; the three teacher adapters bitwise identical across checkpoints, across runs, and against their source safetensors; the post-resume window update magnitude bit-for-bit equal to the uninterrupted run's (`3.0517578125e-4`).
- Inline harness: `_tensor_stats` drops non-finite entries; `pooled_window_reward_metrics` yields finite per-domain means for a NaN-sparse window; both framework installers (`_install_train_step_reset_hook`, `install_phase_timing`) find and wrap `train_step` and fire per call; the window emits one distinct `distill_loss_<domain>` key per part with `prepare` interleaved; single-part path bit-for-bit historical.
- `compileall` / recipe-dotpath lint (2458) / boundary lint / ruff / full pre-commit — clean.

Window-refactor round (`d54211f0`, tree `a1ae23d6` after merging the branch's latest main merge):

- Inline harness (`TrainStack` with recording fakes): 3-part window produces exactly `zero → sync=False ×2 → sync=True → step` with `1/3` backward scales, window-mean loss, the stepping call's grad_norm, and per-domain metric keys side by side; the single-part path reproduces the historical `zero → sync=True → step` bit-for-bit; deferred-sync empty-final guard raises; mid-window empty parts tolerated; multi-update windows rejected.
- GPU no-reward + resume A/B smoke re-run on the refactored tree (8×H20): **ALL acceptance items pass again** — one log line per window with `grad_norm > 0` on every line, `optimizer_step_count` 2→3 at checkpoints 6/9 in both runs, teachers bitwise-frozen (0.0 diffs vs time/runs/source), and the post-resume window update magnitude again **bit-for-bit equal** to the uninterrupted run's (`3.0517578125e-4`).
- `compileall` / recipe-dotpath lint (2458) / boundary lint / ruff / full pre-commit — clean.

Review-fix round (`14aae29e`, CPU box, venv `/root/unirl-cpu-venv`):

- `python -m compileall -q unirl` / `lint/check_recipe_targets.py` (2458 dotpaths incl. the recipe) / `lint/check_experimental_boundaries.py` / `ruff check` + `ruff format --check` on every touched file — all pass.
- Hydra compose: `python -m unirl.train_diffusion --config-name=diffusion/sd3/sd3_opd_multi_teacher --cfg job --resolve` — resolves with `accumulate_rollouts: 3` and the in-repo dataset paths.
- Dataset load: both `datasets/geneval/*.jsonl` and `datasets/ocr/*.txt` load through the real `TextPromptDataset`; geneval rows surface `metadata={tag, include:[{class,count,...}]}` in the list-of-dicts shape the classical scorer consumes.
- Inline harness (real peft): frozen-adapter loader accepts a complete checkpoint and **rejects a truncated one** (missing `lora_B` → ValueError; base-model `missing_keys` correctly ignored).
- Inline harness (`TrainStack` with recording fakes): M=3 window produces exactly `zero → sync=False ×2 → sync=True → step` with `loss_scale = 1/3` per backward and raw (unscaled) logged losses; M=1 defaults reproduce the historical call sequence bit-for-bit; deferred-sync guard raises when the stepping rollout has no backward after accumulated ones; mid-window empty rollouts are tolerated; `_run_updates` rejects accumulation with `num_updates_per_batch > 1`; phase-from-absolute-`rollout_id` is resume-aligned for arbitrary `start_rollout`.

Earlier rounds, on the merged tree (`351efe94`, CPU box, venv `/root/unirl-cpu-venv`):

- `python -m compileall -q unirl unirl-reward-service` — passes.
- `python lint/check_recipe_targets.py` — 2458 recipe `_target_` dotpaths resolve (includes the new recipe).
- `python lint/check_experimental_boundaries.py` — ok.
- `ruff check` / `ruff format --check` on all touched files — clean (repo pre-commit excludes vendored trees).
- Inline harness on the rework (documented in the team board AI Context): real-peft frozen-adapter inject/switch/restore incl. narrow-teacher fall-through; ODE/SDE loss bit-matched against hand-computed KL with the real `FlowSDEStrategy` σ; `metadata["domain"]` routing incl. every error path (missing/mixed/unmapped domain); `Part.slice` micro-batch alignment of `sde_means`; backward reaches the student adapter only.
- GPU smoke: **Not run; reason:** needs a GPU box — planned as follow-up validation: FSDP2 sharding of frozen adapters + LoRA checkpoint save/load roundtrip, then an 8×H20 short run (teacher-student delta norm ≫ 0, three per-domain reward curves, resume across a domain boundary).

The original implementation's live 2-node run (2×8 H20, pre-#214 base) validated the approach end-to-end: adapter switching produced non-zero teacher-student deltas, single-domain batches routed correctly, and per-domain reward curves appeared in wandb.

## Compatibility / Risk

- Additive: new algorithm / data source / scorer; existing recipes and algorithms unchanged.
- `LoraConfig.frozen_adapters` defaults to `None` — no behavior change unless set.
- Trainer skips advantage computation only for `requires_advantages=False` algorithms (all existing ones declare `True`); root-metadata projection only fills an empty field on the train Part.
- `_activate` / `_set_adapter_requires_grad` moved `ema.py → lora.py` (pure move; `ema.py` re-imports).
- Frozen adapters are sharded and checkpointed like everything else — LoRA checkpoints containing teacher keys load back symmetrically because injection happens at build time, before any load.

## Reviewer Notes

- **Review response (2026-08-03, commits `f5f84e7a` / `fef062f7` / `14aae29e`)** — all four external-review findings addressed:
  1. *MOPD lacked cross-teacher accumulation* → `accumulate_rollouts` contract (above); recipe pins `3` and the trainer rejects `M` not a multiple of the domain count, mirroring the official config's assert.
  2. *Reward was a hard dependency despite not entering the loss* → `reward:` is now optional end-to-end (entry, DP preflight, `train_step`, `evaluate`) for `requires_advantages=False` algorithms, with construction-time errors for RL algorithms / `eval_interval>0` / `reward_fraction>0` without a reward.
  3. *Recipe referenced datasets not in the PR* → OCR + GenEval prompt sets ship in-repo (Flow-Factory import, Apache-2.0, `.gitignore` whitelisted like the existing sets; geneval metadata normalized to parsed JSON).
  4. *Frozen-adapter loader silently accepted missing weights* → `missing_keys` filtered to the injected adapter's lora banks and rejected (peft zero-inits `lora_B`, so a missing tensor silently nulls the teacher on that layer).
  5. *(second round, `250d2655` → superseded by `d54211f0`)* *Mid-window checkpoints cannot carry accumulated gradients* — first fixed fail-fast, then fixed **structurally**: the window became one `train_track` call, so a checkpoint between calls can never split it. The surviving `num_rollouts`/`save_interval`/`eval_interval` multiple-of-`M` checks are cadence checks (fire-at-window-ends), not correctness guards; recipe moves to `num_rollouts: 9999`.
- **Third review round (`a352b70e` / `24b577a7`)** — two regressions the window refactor introduced, plus one the GPU run found:
  1. *`train_step` was renamed, breaking framework hooks* — `BaseTrainer`'s transfer-queue buffer reclaim and `install_phase_timing` both install by resolving `getattr(trainer, "train_step")`, so both were silently skipping (a buffer-reclaim and `perf/*` regression for every `DiffusionTrainer` recipe, not just OPD). `train_step` is back as the per-window unit of work — the reclaim fires when it returns, the only point with no live `TensorRef` into the queue — and `accumulate_rollouts > 1` with the transfer_queue transport is now rejected at construction (a window would have to hold every rollout's buffers; unvalidated).
  2. *Window logging only carried the last rollout's rewards* — with per-domain scorers every other domain was NaN-poisoned and never ticked. `pooled_window_reward_metrics` pools `rewards`/`component_rewards` over the window (group stats re-key off the pooled ids) and `_tensor_stats` drops non-finite entries, so every domain has a finite point in every window.
  3. *All per-teacher loss curves were labeled with the last domain* — the window prepared every part before backwarding any, so `prepare_part`'s per-part state (DiffusionOPD's active teacher, which names the metric) was stale for all but the final part. Preparing each part immediately before its own backward fixes it; gradients are unchanged (no step inside a window). Caught by the reward-enabled GPU run, which emitted one `distill_loss_geneval` curve carrying the mean of all three domains.
- **Comment style aligned with #305** (`chore: clean Python comments`): this PR's multi-line `#` rationale blocks were compressed to the one-line standard (docstrings retained), and the branch merged `origin/main` past #305 so the five files both sides touch carry the cleaned baseline.
- **GPU reproduce smoke (8×H20, 201 rollouts = 67 optimizer steps)** validated the paper's behavior end-to-end: `grad_norm` nonzero exactly every 3rd rollout with `optimizer_step_count == 33` at checkpoint-100; per-domain distill losses fell 50%/25%/43% (pickscore/ocr/geneval); post-hoc A/B at identical seeds — PickScore 0.846 → 0.908, OCR text accuracy 0.447 → 0.941 (29 wins / 3 ties / 0 losses over 32 prompts).
- **GPU no-reward + resume smoke (8×H20, M=3, `~reward`, 9 rollouts, save every 3)**: run A (uninterrupted) vs run B (exit at checkpoint-6, resume to 9). All acceptance items pass — no reward role is ever constructed; `grad_norm` nonzero exactly at rollouts 3/6/9 in all three runs; `optimizer_step_count` 2→3 at checkpoints 6/9 for both runs; resume re-enters the window at pickscore→ocr→geneval; all four adapters (`default` + 3 teachers) ×382 tensors in every checkpoint; the three teacher adapters are **bitwise identical** across checkpoints, across runs, and against the source safetensors; the post-resume window update magnitude matches the uninterrupted run's bit-for-bit (`max|ck9−ck6| = 3.0517578125e-4` in both) with monotone `lora_B` growth — faithful continuation, not re-initialization. The smoke also caught (and `89640e5f` fixed) `train()` receiving `${oc.env:...}`-interpolated knobs as strings. Known limitation observed: student `lora_A` init RNG is unseeded (pre-existing, backend-wide), so two *fresh* runs start from different student inits — literal cross-run weight equality needs a seeded LoRA init (candidate follow-up, not OPD-specific).
- **Maintainer replacement, with credit.** @KemingWu's original implementation established the algorithm port, proved the frozen-teacher-adapter approach live on H20s, and identified the real PEFT pitfalls (direct `_active_adapter` routing under FSDP; `inject_adapter_in_model` flipping the student's `requires_grad`) — all carried into the rework. The replacement was structural: post-#214 the branch imported a deleted module (`unirl.types.prompts`); the dual-counter round-robin desynchronized on checkpoint resume (`trainer.train()` fast-forwards the data source; the algorithm counter restarts at 0); eval only ever covered the first domain; `gather_lora_state_dict` swept lazily-injected teachers into every checkpoint asymmetrically.
- The 561-line in-process mmdet GenEval scorer was dropped (unrunnable in the trainer's py3.12 env; the recipe uses the remote service). The 2-node taiji launcher/job snapshot was dropped from the tree; its environment findings (mmcv-full 1.7.2 source build for sm_90, `setuptools<70`, `click<8.2`, `transformers<4.45`) are worth a follow-up doc page.
- Follow-ups: GPU smoke + curve reproduction (above), EMA of student params (paper: decay 0.9 / interval 8), fp16-vs-bf16 alignment study.
- AI-assisted (maintainer side); duplicate-work check: no other open PR touches `diffusionopd.py`, `frozen_adapters`, `MultiDomainRLDataSource`, or `per_domain.py`.

## Checklist


CjhHa1 (commented):
First review pass: the P0 (blocking) and P1 (should-fix-before-merge) items are attached inline on the relevant lines. The five P2 items (eval only covers the first teacher, `_compute_std_var` diverging from the kernel it claims to mirror, `noise_level`/`eta` being decoupled, dead code and missing exports, minor config issues) will follow.

**Overall.** The algorithm itself is cleanly structured and fits the `StageAlgorithm` contract well. `sde_means` is a pre-declared `FieldKind.CONCAT` field, so micro-batch slicing is correct; casting to fp32 before the squared difference and storing anchors with `.detach().cpu()` both match the existing FlowGRPO / FlowDPPO conventions; `requires_advantages=False` and `supports_multi_update=False` are declared correctly and the train stack enforces both. `geneval` and `per_domain` are registered consistently in both `_BUILTIN_SCORERS` and `_BUILTIN_SPECS`. The 2-node job JSON and launcher leak no real credentials — everything is a placeholder.

unirl/data/multi_teacher_data_source.py
```diff
@@ -0,0 +1,269 @@
...
+
+import logging
+import os
+from typing import Any, Dict, Iterator, List, Optional
+
+import torch
+from torch.utils.data import DataLoader
+
+from unirl.types.primitives import Texts
+from unirl.types.prompts import RolloutInputs
```
CjhHa1: **P0 · Blocking: this module fails at import.**

`unirl/types/prompts.py` was deleted on `main` by #214 (the sample-native rollout refactor), and `RolloutInputs` no longer exists anywhere in the repo. Verified:

```
ModuleNotFoundError: No module named 'unirl.types.prompts'
```

The branch merged `main` twice (73f428f, 6ab7786) without conflict, because this file is purely additive and `main` deleted a different file — so git merged it cleanly and nothing surfaced. The five green CI checks don't catch it either: only ruff lint runs, which doesn't resolve imports, and no test in the repo imports this module.

Since the data source is the entry point for the whole recipe, `sd3_opd_pickscore.yaml` currently cannot start.

unirl/data/multi_teacher_data_source.py
```diff
@@ -0,0 +1,269 @@
...
+                md["teacher_domain"] = teacher_name
+                metadata_list.append(md)
+
+            primitives: Dict[str, Any] = {"text": Texts(texts=prompts)}
+            return RolloutInputs(
+                primitives=primitives,
+                sample_ids=sample_ids,
+                group_ids=list(prompt_ids),
+                metadata=metadata_list,
+            )
```
CjhHa1: **P0 · Blocking: renaming `RolloutInputs` isn't enough — the return type has to be `Sample`.**

Two constraints:

1. The trainer enforces the type (`unirl/trainer/base.py:47`):

```python
if not isinstance(inputs, Sample):
    raise TypeError(f"{caller}: expected Sample input, got {type(inputs).__name__}.")
```

2. `PerDomainRewardScorer` routes on `metadata[i]["teacher_domain"]`, and the reward request takes its metadata from `sample.root_metadata(-1)` (`unirl/reward/service.py:52`) — i.e. from the **root `Part`**. Metadata hung on a bare dataclass field never reaches the scorer, so all three domains fall into `unmapped` and get flagged as failures (see the per_domain.py comment: a failure aborts the entire rollout).

I'd reuse `_input_sample(...)` from `unirl/data/data_source.py:260`, which passes metadata into `Part.input(...)` correctly:

```python
primitives: Dict[str, Any] = {"text": Texts(texts=prompts)}
return _input_sample(primitives, sample_ids=sample_ids, metadata=metadata_list)
```

`group_ids` doesn't need to be passed explicitly — `Part.group_ids` is a property derived from `sample_ids` lineage, so passing bare `prompt_id`s here diverges from the framework's semantics.

unirl/algorithms/diffusionopd.py
```diff
@@ -0,0 +1,477 @@
...
+        elif pipeline is not None:
+            bundle = getattr(pipeline, "bundle", None)
+            if bundle is not None:
+                self._transformer = getattr(bundle, "transformer", None)
+
+        # Load teacher adapters (frozen) onto the transformer.
+        # Deferred to first prepare_segment call because at __init__ time the
+        # transformer may not yet be wrapped as a PeftModel (LoRA injection
+        # happens later in the FSDPBackend lifecycle).
+        self._teachers_loaded = False
```
CjhHa1: **P1: teacher adapters are injected after FSDP has already wrapped the model — worth double-checking.**

By the time this is deferred to the first `prepare_segment`, `base_backend._finalize_construction` has already wrapped/sharded the model and built the optimizer from `trainable_params(model)`. `inject_adapter_in_model` then adds fresh `nn.Linear` modules into already-sharded submodules. Those parameters belong to no FSDP param group, so each rank keeps a full replica of them and they sit outside FSDP's device/dtype management.

Compare the existing multi-adapter path in this repo, `unirl/train/ema.py::inject_nft` (lines 157-168): it injects both adapters *before* wrapping, precisely to avoid this. I'd move teacher injection into the backend construction path as well.

Two smaller points while here:

- Lines 40-42 state that writing `_active_adapter` directly is "the same pattern as `unirl.train.lora.adapters_disabled`" — but `adapters_disabled` manipulates `_disable_adapters`, which is a different attribute.
- Adapter switching is already solved once in `ema.py::_activate_keep_grad` (lines 296-310): it calls `set_adapter` and then restores the requires_grad split, with a comment documenting the FSDP2 pitfall. There are now two mechanisms for the same job.

unirl/algorithms/diffusionopd.py
```diff
@@ -0,0 +1,477 @@
...
+            peft_cfg = PeftLoraConfig(
+                r=adapter_cfg.get("r", 32),
+                lora_alpha=adapter_cfg.get("lora_alpha", 64),
+                target_modules=adapter_cfg.get("target_modules", []),
+                lora_dropout=adapter_cfg.get("lora_dropout", 0.0),
+                bias=adapter_cfg.get("bias", "none"),
+            )
+
+            # Inject adapter structure (adds new LoRA A/B matrices under adapter_name).
+            inject_adapter_in_model(peft_cfg, model, adapter_name=adapter_name)
```
CjhHa1: **P1: silently degrades when the teacher's `target_modules` differ from the student's.**

The teacher config is read from its own `adapter_config.json`, which need not match the eight target modules the student declares in `sd3_opd_pickscore.yaml`. Meanwhile `_use_adapter` writes `_active_adapter = [name]` on **every** `LoraLayer`, and PEFT's `LoraLayer.forward` simply `continue`s past any layer whose `self.lora_A` lacks that adapter. Layers the teacher doesn't cover therefore fall through to base weights with no error and no warning.

The result is that the teacher's transition means apply only on a subset of layers and the distillation target is weakened, while the delta norm stays greater than zero — so nothing in the logs reveals it.

I'd assert after injection that the teacher covers the same set of `LoraLayer`s as `default`, and fail loudly otherwise. That mirrors the uniqueness check already done on teacher names in `__init__`.

unirl/algorithms/diffusionopd.py
```diff
@@ -0,0 +1,477 @@
...
+        sde_indices = segment.sde_indices
+        if sde_indices is None or sde_indices.numel() == 0:
+            segment.sde_means = None
+            return
+        target_steps = sde_indices.tolist()
+
+        # Round-robin teacher selection (paper: each batch uses one teacher).
+        teacher_idx = self._rollout_counter % len(self.teachers)
+        self._rollout_counter += 1
+        tc = self.teachers[teacher_idx]
```
CjhHa1: **P1: teachers and prompts drift apart silently after a resume.**

`_rollout_counter` and `MultiTeacherRLDataSource._iter_counter` are two independent counters kept in sync only by the convention that both increment once per rollout. On resume, `unirl/trainer/diffusion.py:639-643` fast-forwards the data source alone:

```python
for _ in range(start_rollout):
    self.data_source.get_samples(self.batch_size)
```

The algorithm is freshly constructed, so its counter restarts at 0. Whenever `start_rollout % 3 != 0`, the OCR teacher ends up supervising GenEval prompts after a resume — exactly the failure mode the module docstring in `multi_teacher_data_source.py` describes as pulling the student off. The delta-norm log below won't catch it either, since it only compares teacher against student and never validates the domain.

The "teachers order MUST match" note in `sd3_opd_pickscore.yaml` is really a workaround for this implicit coupling. The cleaner fix is to have the data source stamp the teacher name into the batch metadata and let the algorithm read and assert on it, which removes the coupling entirely. Failing that, `_rollout_counter` needs to be checkpointed and fast-forwarded alongside the data source.

unirl/algorithms/diffusionopd.py
```diff
@@ -0,0 +1,477 @@
...
+                t_mean = result.prev_sample_means[:, 0].float()
+                s_mean = student_result.prev_sample_means[:, 0].float()
+                delta_norm = (t_mean - s_mean).norm().item()
+                logger.info(
+                    "DiffusionOPD[%s]: teacher-student delta norm at step %d = %.6f "
+                    "(should be >> 0 if adapter switch works)",
+                    tc["name"],
+                    target_steps[0],
+                    delta_norm,
+                )
```
CjhHa1: **P1: an unconditional debug replay sits in the hot path.**

This block runs an extra student replay on every rollout on every rank purely to log a delta norm, and the comment labels it "Debug". The cost is a full transformer forward with CFG, and the `logger.info` fires on every rank.

Worth gating behind a config flag or `logger.isEnabledFor(logging.DEBUG)`. If the adapter switch has already been validated, deleting it outright is cleaner.

unirl/reward/local/per_domain.py
```diff
@@ -0,0 +1,229 @@
...
+                errors=errors,
+                compute_time=time.time() - start,
+            )
+        except Exception as e:
+            return RewardResponse(
+                rewards=[0.0] * bs,
+                successes=[False] * bs,
+                errors=[str(e)] * bs,
+                compute_time=time.time() - start,
+            )
```
CjhHa1: **P1: a reward failure kills a run whose loss doesn't use rewards.**

This broad `except Exception` returns `successes=[False]*bs`, and `RewardService.score_and_attach` raises as soon as any single item is flagged (`unirl/reward/service.py:159-164`):

```python
failed = [(i, e) for i, (ok, e) in enumerate(zip(reward_response.successes, reward_response.errors)) if not ok]
if failed:
    raise RuntimeError(
        f"Reward computation flagged {len(failed)} of {len(reward_response.successes)} "
        f"sample(s) as failure. First few: {failed[:3]}"
    )
```

GenEval is a cross-node HTTP call with a 300s timeout, so one blip on the reward node takes down the whole job. But the PR description repeatedly notes that reward is monitoring-only for OPD and never enters the loss — losing training over a monitoring metric inverts the failure mode.

The branch on lines 158-160 that flags unmapped items as failures has the same problem, and it will fire on 100% of items while the metadata plumbing is still broken. For a monitoring-only path I'd degrade gracefully — record NaN and warn — rather than fail the rollout.

CjhHa1 (commented):
Inline notes for the P0/P1 items in the review summary above.

haonan3 (approved):
LGTM

PR #138, opened 2026-06-27
feat(rollout): LTX-2 T2V sglang_diffusion rollout (audio co-denoise aux, full+LoRA)

leviking98z-rgb:
## Summary

Adds **LTX-2 T2V** to the `sglang_diffusion` rollout engine (full-FT + LoRA), with the audiovisual co-denoising handled for rollout<->replay alignment.

LTX-2 is a unified AV transformer: the video DiT cross-attends to a co-denoised **audio latent** each step, so the trainside replay must consume the SAME per-step audio the rollout used.
- `Ltx2T2VAdapter` — LTX-2 latents are packed token sequences (rank-4 `[B, T+1, seq, dim]`), not a spatial grid; relaxes the strict 6-D gate.
- **Audio-trajectory carry** — captures sglang's co-denoised audio trajectory and routes it to `segment.aux_latents` so `LTX2DiffusionStage.replay` reproduces the exact audio (else "aux_latents missing" + curve divergence).
- `patch_ltx2_rollout_sde` (sigma-schedule pin / AV-decode output carry / SDE log-prob bridge for LTX-2's overridden stages); RL-mixin scheduler swap; forced-MATH-SDP attention workaround; aux-components-on-CPU for the colocate budget.
- Configs: `ltx2_t2v_sglang_{loramerge,full}.yaml`.

## Related Issue

Engine support matrix — #127. Depends on the foundation #135.

## Test Plan

- `py_compile` + import smoke (registers as `ltx2`) — passes.
- GRPO smoke, rollout<->replay ratio: **LoRA 0.9985 -> 0.9988 -> 0.9992** (3 rollouts), **full-FT 0.9968** — the sglang dual-AV rollout matches trainside replay.

## Compatibility / Risk

Additive; LTX-2-specific patches are gated on the LTX-2 stages. The aux-trajectory plumbing (`collect_aux_trajectory_latents`, `RawResult.aux_trajectory_latents`, segment `aux_latents`) is additive and `None` for non-LTX-2 families.

## Reviewer Notes

Independent of #136/#137/#139. Full-FT multi-rollout on a single node uses the documented colocate-offload settings in the config.

## Checklist

- [x] I reviewed the changed code and removed unrelated/generated artifacts.
- [x] I updated tests, docs, and configs where needed, or explained why not.

PR #233, opened 2026-07-20
feat(ltx2): add aligned SGLang rollout support

leviking98z-rgb:

Independently revalidated the effective merge with current `main` (`fce4f4b`): synthetic merge clean, full pre-commit clean, and both LTX-2 recipes compose. An out-of-tree inline harness passed for the cross-engine step-noise seed contract, exact Diffusers 0.37 text packing, transformer forward-signature compatibility, and lockstep video/audio trajectory collection and sparse trimming. The monkey-patch call sites were also checked against the pinned upstream SGLang `v0.5.12.post1` sources. No test file was added.

## Summary

Adds the minimum LTX-2 T2V SGLang LoRA-merge rollout path and keeps only the
alignment changes needed for it to implement the same objective as the existing
trainside recipe:

1. **Rollout distribution:** driver-pinned sigma schedule and audio/video `x_T`,
   identical per-sample/per-step SDE noise, `max_sequence_length=512`, canonical
   Gemma packing, fp32 denoising state, fp32-equivalent LTX RoPE arithmetic, and
   aligned T+1 video/audio trajectories.
2. **Prompt mask:** LTX connector padding is replaced by learned registers, so its
   post-connector mask is all-valid. Both rollout and replay now represent that as
   `None`; this also keeps SGLang on its native attention backend instead of the
   crashing masked-SDPA fallback.
3. **GRPO weighting:** `adv_use_global_std=false`, matching trainside per-prompt
   group normalization.
4. **Audio conditioning:** carry separate video/audio prompt embeddings and the
   co-denoised audio trajectory into replay.

The remaining non-algorithmic changes are runtime prerequisites: the LTX adapter,
rollout-aware scheduler/log-prob bridge, trajectory transport, diffusers 0.37/0.38
connector compatibility, and CPU parking for frozen trainside auxiliary modules in
colocate mode.

This revision intentionally excludes the full-FT recipe, the qk-norm and separate
scheduler patch files, force-MATH-SDPA, environment-gated mask/RoPE behavior,
SPS16/online-LoRA recipes, and LoRA-init/optimizer-precision experiments.

## Related Issue

Supersedes the closed draft #138 and contributes to the rollout-engine support
matrix in #127. The SPS16-only RFC #216 remains closed and is not included.

## Test Plan

- `pre-commit run --from-ref origin/main --to-ref HEAD`: passed.
- Ruff lint/format and Python compilation for all 20 changed Python files: passed.
- Hydra full-resolution composition for the trainside and SGLang LoRA-merge
  recipes: passed.
- Deterministic step-noise probe: trainside and SGLang produce identical tensors
  for the same `(seed, step_index, sample_id)`.
- Text packing probe: UniRL's diffusers-0.37 compatibility path is bit-identical to
  SGLang; the diffusers-0.38 path delegates normalization to the connector exactly
  once.
- RoPE probe: patched SGLang split/interleaved outputs are bit-identical to
  diffusers fp32 arithmetic.
- Regression probes: a pinned rollout schedule still preserves SGLang's
  `ltx2_phase=stage1` side effect; mismatched audio/video trajectory batch or
  T+1 lengths fail fast, while sparse columns are trimmed in lockstep.
- 8x H20 reduced-resolution (9-frame) SGLang smoke with no force-MATH patch
  and native FlashAttention: reward `0.6865`, ratio `1.0000+/-0.0000`.
- 8x H20 paired 9-frame trainside smoke: reward `0.6857`, ratio
  `1.0000+/-0.0000`.

## Compatibility / Risk

- The checked-in recipes no longer depend on `UNIRL_LTX_NO_TEXT_MASK` or
  `UNIRL_LTX_BF16_ROPE`.
- The LTX prompt-mask optimization only drops a tensor when every position is
  valid; a genuinely partial mask is preserved.
- Frozen VAE/Gemma/connectors stay on CPU only when
  `aux_components_on_cpu=true` (the SGLang colocate recipe); the trainside recipe
  retains its existing device placement.
- The implementation remains pinned to the repository's SGLang
  `0.5.12.post1` integration surface.

## Reviewer Notes

- One SGLang LoRA-merge recipe is retained; the unvalidated full-FT sibling is
  excluded.
- The published commits contain no agent/Claude co-author trailer.
- The PR is rebased on current `main`.

## Checklist

- [x] Rebased on current `main` and reviewed the cumulative diff file by file.
- [x] Removed unrelated/experimental compatibility patches and recipes.
- [x] Kept only the four alignment fixes plus required rollout plumbing and one recipe.
- [x] Static checks, composition probes, and an 8-GPU SGLang smoke pass.
- [x] Paired 8-GPU trainside smoke recorded above.

## EXP
截屏2026-07-26 02 53 54


KemingWu (approved):
LGTM

PR #234, opened 2026-07-21
feat(sft): support agent conversation fine-tuning

xshrz:
## Summary
- Add OpenAI-style multi-turn agent SFT with optional tools and structured tool calls.
- Render conversation history and assistant targets through the same Qwen template while supervising only the assistant suffix.
- Add deterministic agent dataset preparation and a Qwen3-8B LoRA/FSDP recipe with W&B and checkpoint support.

## Related Issue
N/A

## Test Plan
- `python -m pytest -q tests/sft/test_agent_sft.py` (`4 passed`), covering overlong structured-target filtering, `append_eos=False`, default manifest paths, and non-empty train/validation splitting.
- `python -m unirl.utils.prepare_sft_agent --out-dir datasets/sft_agent_toolcall_12k`
- `SFT_DATA=datasets/sft_agent_toolcall_12k/train.jsonl SFT_EVAL_DATA=datasets/sft_agent_toolcall_12k/val.jsonl QWEN3_PATH=Qwen/Qwen3-8B python -m unirl.train_sft --config-name=sft/validation/qwen3_agent_sft_lora --cfg job --resolve`
- Validated with Qwen3-8B on 1 node × 8 NVIDIA H20 GPUs; verified finite train/eval loss, W&B metrics, and checkpoint creation.

## Compatibility / Risk
- Legacy `prompt`/`response` manifests remain supported.
- Agent prompt and response lengths are bounded by configurable token limits.
- Adds `huggingface-hub` as a direct dependency.

## Reviewer Notes
- Built on the SFT domain introduced by #209.
- No overlapping open PR or issue was found.
- Generated datasets, checkpoints, credentials, and outputs are excluded; focused regression tests are included under `tests/sft/`.

## Checklist


xshrz:
企业微信截图_ce3d61fe-7f69-4dbc-ab76-e5ac09e94f6f

企业微信截图_223f232a-0f47-4e12-a738-c68d4c922ea6

examples/sft/validation/qwen3_agent_sft_lora.yaml
```diff
@@ -0,0 +1,105 @@
...
+    path: unirl.models.qwen3.conditions.Qwen3ARConditions
+
+stack:
+  _target_: unirl.train.stack.TrainStack
+  micro_batch_size: 1
+  max_grad_norm: 1.0
+
+track_builder:
+  _target_: unirl.train.sft.track_builder.ARSupervisedTrackBuilder
+  max_response_length: 1024
```
CjhHa1, line 99: **P1: The default dataset is severely mismatched with the 1,024-token cap.** Using the Qwen3-8B tokenizer, 4,891 of 10,000 final-answer targets and 324 of 27,156 tool-call targets exceed this limit. The current builder truncates them and forces an EOS token, which cuts some `<tool_call>` JSON objects in the middle and creates invalid supervision. Please filter overlong samples by token length during preprocessing; structured calls should at least be dropped or kept structurally complete rather than truncated and terminated with EOS.

unirl/train/sft/track_builder.py
```diff
@@ -191,23 +199,29 @@ def _tokenize_responses(self, records: Sequence[Record]) -> Tuple[List[torch.Ten
...
                 )
-            budget = self.max_response_length - (1 if self.append_eos else 0)
+            needs_eos = self.append_eos and (not ids or ids[-1] != eos_id)
+            budget = self.max_response_length - (1 if needs_eos else 0)
             if len(ids) > budget:
-                ids = ids[:budget]
+                ids = list(ids[:budget])
                 truncated += 1
-            if self.append_eos:
+                if not needs_eos and eos_id is not None:
```
CjhHa1: **P2: This violates the `append_eos=False` contract.** When EOS appending is disabled, `needs_eos` is always false, so any overlong response still reaches this branch and has its last retained token replaced with EOS. With `ids=[1,2,3]`, limit 2, and EOS 99, this produces `[1,99]` instead of `[1,2]`. Please also guard this replacement with `self.append_eos`.

examples/sft/validation/qwen3_agent_sft_lora.yaml
```diff
@@ -0,0 +1,105 @@
...
+
+data_source:
+  _target_: unirl.data.sft.SupervisedDataSource
+  manifest_path: ${oc.env:SFT_DATA}
```
CjhHa1: **P2: The documented prepare/launch flow cannot run as written.** Preparation writes `datasets/sft_agent_toolcall_12k/{train,val}.jsonl`, but the launch example sets neither `SFT_DATA` nor `SFT_EVAL_DATA`, and these interpolations have no defaults. Following the instructions fails during Hydra resolution with `Environment variable 'SFT_DATA' not found`. Please either default both fields to the generated paths or set both variables in the launch example.

unirl/utils/prepare_sft_agent.py
```diff
@@ -0,0 +1,128 @@
...
+
+    source_path = hf_hub_download(repo_id=args.dataset, filename=args.filename, repo_type="dataset")
+    rows = _read_jsonl(source_path)
+    random.Random(args.seed).shuffle(rows)
+    if args.max_trajectories > 0:
+        rows = rows[: args.max_trajectories]
+    if len(rows) < 2:
+        raise SystemExit(f"prepare_sft_agent: only {len(rows)} trajectories found.")
+
+    n_val = max(1, round(len(rows) * args.val_fraction))
```
CjhHa1: **P2: Valid CLI arguments can produce an empty training split while the command exits successfully.** For example, `--max-trajectories 2 --val-fraction 0.9` gives `n_val=2` and writes a zero-row `train.jsonl`; the failure is deferred until training loads the manifest. Please clamp `n_val` to `[1, len(rows) - 1]` or fail immediately when the training split would be empty.

CjhHa1 (commented):
The following four issues were reproduced in the default Agent SFT workflow; see the inline comments for details.

examples/sft/validation/qwen3_agent_sft_lora.yaml
xshrz: Fixed. `prepare_sft_agent` now loads the configured tokenizer (defaulting to `QWEN3_PATH`, or `Qwen/Qwen3-8B`), renders each assistant target with the same chat-template and `enable_thinking` settings used by training, accounts for the builder's EOS behavior, and drops targets exceeding `--max-response-length` (default 1,024) before writing either manifest. It also reports filtered tool-call and final-answer counts separately.

The preparation path and worker builder now share the target-tokenization helper so their length accounting cannot drift. As a defensive backstop, the builder rejects an overlong agent target instead of truncating a structured assistant turn, so a custom/unfiltered manifest cannot produce partial `<tool_call>` JSON supervision.

unirl/train/sft/track_builder.py
xshrz: Fixed — the EOS replacement branch is now guarded by `self.append_eos`. With `ids=[1, 2, 3]`, `max_response_length=2`, `eos_token_id=99`, and `append_eos=False`, the retained target is `[1, 2]`; no EOS is injected. I also updated the one-time truncation warning so it no longer claims EOS was kept when appending is disabled.

examples/sft/validation/qwen3_agent_sft_lora.yaml
xshrz: Fixed by giving both recipe interpolations defaults that match the documented preparation output:

- `SFT_DATA` defaults to `datasets/sft_agent_toolcall_12k/train.jsonl`
- `SFT_EVAL_DATA` defaults to `datasets/sft_agent_toolcall_12k/val.jsonl`

The environment variables remain available as overrides, but the documented prepare command followed by the launch command no longer requires extra setup.

unirl/utils/prepare_sft_agent.py
xshrz: Fixed in `3511e69`. Validation allocation is clamped with `min(len(rows) - 1, max(1, ...))`, so every accepted input keeps at least one trajectory in both splits. In the reproduced `--max-trajectories 2 --val-fraction 0.9` case, `n_val` is now 1 and the training manifest remains non-empty.

examples/sft/validation/qwen3_agent_sft_lora.yaml
leviking98z-rgb: Verified at current head and covered in `fe354a7`: overlong structured assistant targets are filtered during preparation, and the worker raises instead of truncating if an unfiltered overlong agent target reaches it.

unirl/train/sft/track_builder.py
leviking98z-rgb: Verified and regression-covered in `fe354a7`: with `append_eos=False`, `[1, 2, 3]` under a two-token limit remains `[1, 2]`; the last retained token is not replaced.

examples/sft/validation/qwen3_agent_sft_lora.yaml
leviking98z-rgb: Verified and regression-covered in `fe354a7`: with both environment variables unset, the recipe resolves to the train/validation manifests written by the documented preparation command.

leviking98z-rgb (approved):
Reviewed the complete current diff after syncing `main` and verified all four previously reported data-path issues. Overlong structured assistant targets are filtered and fail closed at the worker, legacy truncation respects `append_eos=False`, recipe defaults match preparation outputs, and valid split settings keep both manifests non-empty. Focused regressions now cover each case. Verified locally: 4 pytest cases passed, targeted pre-commit passed, and `git diff --check` is clean. GitHub lint, body/title validation, CodeQL actions, CodeQL Python, and the status workflow are green. No remaining code-level blocker.

zzhuoxin1508:
## 16-GPU Agent SFT Validation
I ran the Agent Trace SFT recipe on Qwen3-8B and evaluated tool-calling behavior before and after training.
### Setup
- **Model:** Qwen3-8B
- **Dataset:** `pyromind/agentic-tool-call-dataset-12k`
- **Training:** LoRA SFT, 2 nodes × 8 NVIDIA H20 GPUs, 300 steps
- **Evaluation:** 64 fixed held-out tool-call targets
- **Objective:** Generate the next assistant `<tool_call>` from the conversation history
- **Comparison:** Base Qwen3-8B vs. SFT checkpoint at step 300
### Results
- **MCP request success proxy:** 39.06% → 67.19% (**+28.13 pp**)
- **MCP error-rate proxy:** 60.94% → 32.81% (**46.2% relative reduction**)
- **Tool-name exact match:** 45.31% → 82.81%
- **Full-call exact match:** 21.88% → 42.19%
- **Validation loss:** 1.7867 → 0.8306
The MCP success proxy checks whether the generated call has a valid structure, references an available tool, and provides arguments compatible with the tool schema. This is an offline generation evaluation: no real MCP server was invoked, so transport, permission, and server-runtime failures are not covered.
### Code Version
The experiment was based on the official `upstream/main` commit:
`cf183744239345f1be24c24e0e03cdc7d167cfc0`
This was the latest `Tencent-Hunyuan/UniRL:main` commit when the experiment was prepared.
Clipboard_Screenshot_1786008573
Two local changes were used:

1. Changed `model_precision` from `bf16` to `fp32` because the original setting failed before the first training step with `FSDP expects uniform original parameter dtype but got {torch.bfloat16, torch.float32}`. The base model was loaded in BF16 while the injected LoRA parameters were initialized in FP32. Mixed-precision computation remained BF16.

PR #238, opened 2026-07-21
fix(trainer): preserve failures during async teardown

CjhHa1:
## Summary

- preserve the primary training failure when final async draining also fails
- always flush pending checkpoints and close WandB, including resume-sync and baseline-eval failures
- share the teardown guard across `AsyncARTrainer` and `AsyncDiffusionTrainer`
- add CPU-only coverage for clean exit, drain failure, primary-error preservation, and `KeyboardInterrupt`

## Dependency

Stacked on #192 because the shared fix covers its new `AsyncDiffusionTrainer`. Keep this PR draft until #192 merges; its diff against `main` will then collapse to the teardown changes.

## Test plan

- [x] `PYTHONPATH=$PWD python3 -m pytest -q tests/test_async_teardown.py` (4 passed)
- [x] `pre-commit run --files unirl/trainer/base.py unirl/trainer/async_ar.py unirl/trainer/async_diffusion.py tests/test_async_teardown.py`
- [x] async diffusion Hydra compose (`--cfg job --resolve`)


CjhHa1:
Superseded by #239. Main merged the generic async runtime in #211 while this draft was being prepared, so the replacement narrows the follow-up to AsyncDiffusionTrainer and stacks cleanly on the latest main + #192.

PR #239, opened 2026-07-21
fix(trainer): preserve async diffusion teardown cause

CjhHa1:
## Summary

- preserve the primary training exception when final async diffusion draining also fails
- always run pending-checkpoint and WandB finalization after a secondary drain failure
- match the teardown behavior already merged for `AsyncARTrainer` in #211

## Dependency

Stacked on #192 because it fixes the new `AsyncDiffusionTrainer`. Keep this PR draft until #192 merges; then merge current `main` into this branch so the visible diff collapses to the teardown guard.

## Test plan

- [x] `pre-commit run --files unirl/trainer/async_diffusion.py`
- [x] async diffusion Hydra compose (`--cfg job --resolve`)
- [x] verified clean-exit drain failures still propagate; secondary drain failures are logged while the primary exception remains active


CjhHa1:
Superseded by #192, including equivalent teardown hardening in af190716, and subsequently centralized by #315. No unique changes remain.

PR #243, opened 2026-07-22
perf(sft): optimize supervised data preparation

CjhHa1:
## Summary
- rebase the PR onto current `main` (`27be70b3`) and preserve current agent, audio, image, and video SFT behavior
- name the two data inputs symmetrically as `train_manifest_path` and `eval_manifest_path`
- make resume fail closed when the train manifest content, seed, or shuffle mode changes
- require an explicit validation split across all affected SFT recipes
- batch response tokenization, parallelize image loading, and overlap Qwen-VL CPU preprocessing with training
- add opt-in, fingerprinted caches for frozen eval-mode text conditions and deterministic VAE latents
- validate cache targets at construction, version the disk-cache schema, and reject unsafe cache requests
- reuse the existing `PhaseTimer` for train and eval phase breakdowns
- keep prefetch cursor updates O(batch), serialize eval against in-flight tokenizer work, and treat concurrent cache eviction as a miss

## Results
- SD3 frozen text cache: `1.20x` step speedup
- SD3 frozen VAE cache: `1.50x` step speedup
- BAGEL frozen VAE cache: `10.86x` second-epoch encode speedup with zero latent difference
- batched tokenizer: `3.41x` tokenizer-phase speedup
- parallel cold image loading: `3.51x` image-load speedup
- micro loss-weight batching and fused dispatch were measured and intentionally excluded

## Test Plan
- [x] Full repository pre-commit suite on rebased head
- [x] All 13 SFT recipes compose with `train_manifest_path`; all nine eval-policy-affected recipes resolve without an implicit train fallback
- [x] Focused CPU checks for lazy manifest-aware exact resume and epoch-crossing peek/commit equivalence without per-step reshuffles
- [x] Focused checks for batched tokenization, parallel image loading, eval/prefetch serialization, cache target validation, cache hits/invalidation, and concurrent cache eviction
- [x] Shared `PhaseTimer` integration on the first train step and eval phase metric emission
- [x] Final diff against `main` contains no test-tree changes
- [x] Original Qwen3 and Qwen-VL LoRA SFT smoke runs
- [x] Original SD3 text-condition and VAE-cache A/B runs
- [x] Original BAGEL VAE-cache A/B and one-step full SFT smoke
- [x] Original checkpoint-sidecar resume through the next optimizer step

## Compatibility / Risk
- `SupervisedDataSource.manifest_path` is renamed to `train_manifest_path`; every in-repository caller is migrated.
- Encoder caches remain opt-in and require an explicit cache fingerprint; cache requests for missing, trainable, or training-mode encoders fail during construction.
- Cache directories are explicit trusted training artifacts; cache files are loaded through PyTorch serialization.
- Evaluation no longer silently reuses the training manifest; it is disabled unless an eval manifest is explicit.
- Qwen-VL prefetch has one trainer-level opt-in and is enabled in the two Qwen-VL example recipes; builders without the capability are rejected.
- The measured GPU A/B and smoke results above are author-provided historical evidence; the current rebase pass is CPU/static validation.

### DIFF
--- a/examples/sft/bagel_agent_sft.yaml
+++ b/examples/sft/bagel_agent_sft.yaml
@@ -16,7 +16,7 @@
 # Env:
 #   BAGEL_PATH      HF id or local ckpt (default ByteDance-Seed/BAGEL-7B-MoT)
 #   SFT_DATA        train manifest (searchgen prepare_sft train.jsonl)
-#   SFT_EVAL_DATA   eval manifest (defaults to SFT_DATA — supply a real split!)
+#   SFT_EVAL_DATA   optional eval manifest (validation is disabled when unset)
 #
 # Download/cook guide: datasets/searchgen/README.md (manifests shared with qwen_vl_agent_sft).
 #
@@ -135,6 +135,6 @@ track_builder:
 
 data_source:
   _target_: unirl.data.sft.SupervisedDataSource
-  manifest_path: ${oc.env:SFT_DATA}
-  eval_manifest_path: ${oc.env:SFT_EVAL_DATA,${oc.env:SFT_DATA}}
+  train_manifest_path: ${oc.env:SFT_DATA}
+  eval_manifest_path: ${oc.env:SFT_EVAL_DATA,null}
   seed: 42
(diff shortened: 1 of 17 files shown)


nussejzz (approved):

nussejzz:
LGTM @CjhHa1

PR #237, opened 2026-07-21
fix(ci): skip fork review label writes

CjhHa1:
## Summary
- skip status-label mutation for fork-triggered `pull_request_review` runs, whose `GITHUB_TOKEN` is read-only
- preserve label synchronization for same-repository reviews and all existing `pull_request_target` events
- rely on GitHub's native `reviewDecision` to gate merges when the cosmetic fork label update is skipped

## Test Plan
- [x] run repository pre-commit checks on `.github/workflows/pr-status-labels.yml`
- [x] validate YAML and whitespace
- [ ] submit or dismiss a review on a fork PR after merge and confirm the label job is skipped rather than failed

PR #245, opened 2026-07-24
fix(ci): sync review labels for fork PRs

celve:
## Summary

- route pull request review events through an unprivileged collector so fork PRs no longer attempt label writes with a read-only `GITHUB_TOKEN`
- reconcile the live GitHub `reviewDecision` in a separate `workflow_run` job with narrowly scoped label-write permission
- preserve the existing `pull_request_target` lifecycle behavior and add a manual bulk reconciliation for already-stale open PR labels

Fork review runs currently fail with HTTP 403 after `pull_request_target` has added `need review`, leaving approved PRs mislabeled. The two-stage handoff passes only a validated PR number and never checks out or executes pull-request-controlled content.

## Related Issue

N/A (tracked as LIN-591).

## Test Plan

- `uvx --from pre-commit pre-commit run --files .github/workflows/pr-status-labels.yml .github/workflows/pr-review-event.yml .github/workflows/pr-review-status-labels.yml`
- `git diff --check upstream/main`
- parsed all three workflow files with PyYAML and ran `bash -n` against every embedded `run:` block
- ran a read-only reconciliation dry run against all open upstream PRs and confirmed the expected stale-label set
- End-to-end Actions run not performed; `pull_request_review` and `workflow_run` changes must exist on the default branch before they can be exercised.

## Compatibility / Risk

No application API, configuration, checkpoint, or data-format changes.

The privileged workflow has only `actions: read`, `pull-requests: read`, and `issues: write`; it consumes one validated numeric artifact and does not check out fork code. Manual dispatch intentionally reconciles all open PRs.

## Reviewer Notes

This supersedes #237. That PR suppresses failing fork writes but deliberately leaves review labels stale; this PR completes the privilege-separated label update instead.

Duplicate-work check found no other overlapping open PRs. Implemented with Codex assistance; the full staged diff and permission boundary were reviewed before publication.

## Checklist


leviking98z-rgb (dismissed):

.github/workflows/pr-review-status-labels.yml
```diff
@@ -0,0 +1,105 @@
...
+  workflow_dispatch:
+
+permissions:
+  actions: read
+  pull-requests: read
+  issues: write
+
+concurrency:
+  group: pr-review-status-labels
+  cancel-in-progress: false
```
leviking98z-rgb, line 18: This constant concurrency group can silently drop reconciliations for unrelated PRs. `cancel-in-progress: false` protects only the running job; GitHub retains at most one pending run in the default concurrency queue and replaces it when another run arrives. For example, while a bulk dispatch is running, approvals for PR A and then PR B leave only B pending, so A can remain mislabeled indefinitely. Please enable `queue: max`, remove the global serialization, or read and validate the artifact in a preparatory job and use a per-PR concurrency key for the mutation job. See: https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency

.github/workflows/pr-review-status-labels.yml
```diff
@@ -0,0 +1,105 @@
...
+
+            draft=$(jq -r '.isDraft' <<<"$pr")
+            decision=$(jq -r '.reviewDecision // ""' <<<"$pr")
+            if [ "$draft" = "true" ]; then
+              target="wip"
+            else
+              case "$decision" in
+                APPROVED) target="approved" ;;
+                CHANGES_REQUESTED) target="changes requested" ;;
+                *) target="need review" ;;
```
leviking98z-rgb, line 68: This mapping conflicts with the retained `synchronize` transition in `pr-status-labels.yml` (lines 33-36). That workflow deliberately changes `changes requested` to `need review` after a push, while this code maps the still-live `CHANGES_REQUESTED` decision back to `changes requested` (the repository ruleset currently has `dismiss_stale_reviews_on_push: false`). A manual reconciliation or a COMMENTED review after the push therefore reverses the lifecycle transition, so the final label depends on which workflow runs last. Please choose one source of truth and apply it consistently in both workflows, ideally under the same per-PR serialization.

.github/workflows/pr-review-status-labels.yml
```diff
@@ -0,0 +1,105 @@
...
+
+            for label in "wip" "need review" "changes requested" "approved"; do
+              [ "$label" = "$target" ] && continue
+              if jq -e --arg name "$label" '.labels | any(.name == $name)' <<<"$pr" >/dev/null; then
+                encoded=$(jq -rn --arg label "$label" '$label | @uri')
+                gh api --method DELETE \
+                  "repos/$GITHUB_REPOSITORY/issues/$number/labels/$encoded" \
+                  >/dev/null
+              fi
+            done
```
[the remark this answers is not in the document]
leviking98z-rgb: Confirmed. A concrete failing interleaving is: this job snapshots `changes requested` and computes the target as `approved`; `pr-status-labels.yml` deletes that label first; this DELETE then receives 404; and `set -e` exits before `approved` is added. Please treat an already-absent label as a successful deletion (or selectively ignore 404), and re-read/retry if atomic reconciliation is required. Note that adding `|| true` alone does not resolve the separate cross-workflow state-machine race.

leviking98z-rgb (approved):

PR #246, opened 2026-07-24
fix: unblock bagel trainside load and surface worker OOMs instead of hanging

celve:
## Summary

Two fixes found while running `examples/unified_model/bagel_trainside_unigrpo.yaml` on a single 8-GPU node.

1. **`fix(bagel/vendor)` — flash-attn fallback stub.** When flash-attn is not installed, the vendored bagel fallback puts a bare `types.ModuleType("flash_attn")` in `sys.modules`. A bare module has `__spec__ = None`, so transformers' `importlib.util.find_spec("flash_attn")` availability probe raises `ValueError("flash_attn.__spec__ is None")` and crashes `BagelBundle` load — before the SDPA `flash_attn_varlen_func` fallback (the whole point of the stub) can be used. The stub now gets a real `ModuleSpec`, so on a box without flash-attn the intended SDPA path loads cleanly. (transformers still reports flash-attn *unavailable* via `importlib.metadata.version`, so it correctly stays on SDPA.)

2. **`fix(trainer)` — surface worker OOMs instead of a hung teardown.** A worker CUDA OOM (or any worker exception) re-raised on the driver by `ray.get` was swallowed: the training loop fell straight into `finally: _finish_wandb()`, whose `_wait_for_checkpoints()` does a *blocking* `ray.get` on the very workers the failure had just wedged in an NCCL collective — so teardown hung forever and the real traceback never printed. The only thing that eventually surfaced was a generic 30-minute `ProcessGroupNCCL` collective-timeout, pointing at comms rather than the real memory cause. Two changes:
   - the training loop logs the failure at the catch point (`Training loop aborted at rollout N`) and re-raises, so the cause is visible immediately;
   - `_finish_wandb` still *attempts* the best-effort checkpoint/weight-sync flush while an exception unwinds — a healthy crash (Ctrl-C, a driver-side error) still drains the in-flight async DCP save so the last checkpoint is not left half-written — but **bounds** it with ray's native `ray.get(refs, timeout=…)`. A wedged worker can't service the flush, so on timeout the `GetTimeoutError` is caught and logged and the primary exception is preserved and re-raised.

The PE trainer override is updated to accept and forward the same optional timeout, preventing the base teardown hook from raising `TypeError` on PE success or exception paths.

The bound is an optional `_ray_get_timeout` on the generic dispatch (`handle_fn`), popped before dispatch so it never reaches the worker method; it is applied only on the exception-teardown path (`UNIRL_TEARDOWN_FLUSH_TIMEOUT_S`, default 120s). Every other call passes no timeout and is byte-identical.

## Related Issue

LIN-590 (Linear).

## Test Plan

- Five focused regressions passed before their PR-added files were removed, covering reserved-kwarg stripping/timeout forwarding, fallback `ModuleSpec`, primary-exception preservation, and PE timeout compatibility. The final PR diff contains no test files.
- Targeted changed-file pre-commit suite and `git diff --check` passed.

Hardware/setup: 1×8 H20 (95 GiB/GPU), torch 2.11+cu129, single-node Ray, `celve/unirl @ LIN-590/main` (`66b546da`).

- **flash-attn stub fix** (venv without flash-attn): `importlib.util.find_spec("flash_attn")` returns a truthy spec and no longer raises; the SDPA `flash_attn_varlen_func` is injected, and `bagel_trainside_unigrpo` loads `BagelBundle` and runs a full rollout (previously crashed at bundle load).
- **trainer teardown fix** — reproduced a rank-0 CUDA OOM with `bash examples/run_experiment_single_node.sh unified_model/bagel_trainside_unigrpo num_devices=6 +devices_per_node=6 batch_size=6 sampling.ar.samples_per_prompt=4 sampling.ar.max_new_tokens=512 num_rollouts=2000 save_interval=100`, run with `UNIRL_TEARDOWN_FLUSH_TIMEOUT_S=45` so the bound is easy to observe:
  - **Before:** ~30 min silent hang (`gpu=[0,100,100,100,100,100,0,0]`), then only `[Rank N] Watchdog caught collective operation timeout: WorkNCCL(... OpType=_REDUCE_SCATTER_BASE ...)`.
  - **After (actual log):** `[unified_model][ERROR] Training loop aborted at rollout 0` + full `torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 1.74 GiB. GPU 0 ...` (FSDP `foreach_reduce`) at `20:39:06`; the flush hits the bound exactly `45s` later at `20:39:51` → `[base][ERROR] Failed to flush checkpoint/weight-sync state during trainer teardown` caused by `ray.exceptions.GetTimeoutError: Get timed out`; the OOM re-raises as the fatal exception, `TRAIN_EXIT=1`, GPUs freed. No `Watchdog caught collective operation timeout` (exits at the bound, well before any NCCL timeout).
- `python -m py_compile` on the five changed production files passes.

## Compatibility / Risk

Low. No config, checkpoint, data-format, API, or resource-requirement changes.
- The trainer change only affects the exception-teardown path: it bounds a best-effort flush that runs **only when the process is already crashing**. On a healthy crash the flush still completes within the bound, so the last checkpoint is still saved; successful runs pass no timeout and are unchanged.
- The dispatch change is a new *optional* kwarg defaulting to `None` (unbounded); the PE override now preserves that contract for both of its backends.
- The flash-attn change only affects the no-flash-attn fallback path; with real flash-attn installed, `import_module("flash_attn")` succeeds and the stub branch is never taken.

## Reviewer Notes

- The trainer fix improves **observability** of a worker OOM; it does not change the memory footprint. `bagel_trainside_unigrpo` at `num_devices=6` still OOMs on GPU 0 (it sits at the ~95 GiB ceiling); actually fixing that (e.g. `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True` or `num_devices=8`) is a separate, config-level change intentionally not in this PR.
- **Why bound rather than skip:** an earlier revision skipped the flush entirely whenever any exception unwound, which would also drop the in-flight `dcp.async_save` on a *healthy* crash (Ctrl-C / driver error) and leave the last checkpoint partial. Bounding instead still drains when the workers are responsive and only gives up when one is wedged.
- **Relationship to existing teardown work:** #211 (merged) and #239 (draft, `AsyncDiffusionTrainer`) establish the `active_exception` "don't mask the primary exception during teardown" pattern and treat `BaseTrainer._finish_wandb` as the reference; those handle cleanup that *raises*. This PR extends the same `_finish_wandb` to the case where cleanup *hangs* — a blocking `ray.get` on a worker wedged in an NCCL collective — by bounding it.
- **Duplicate-work check (upstream):** no open PR fixes the flash-attn `__spec__` stub. #218 (draft, bagel vLLM-Omni) edits the same `vendor/__init__.py` but only its provenance docstring, not the executable stub — no functional overlap, minimal conflict risk.
- Follow-up worth considering (not in this PR): shorten the NCCL collective timeout and enable the flight recorder so *unforeseen* single-rank deaths also abort fast with a per-rank dump.
- A temporary `[LIN590]` diagnostic used during investigation is not part of this change.

## Checklist

- [x] I validated the guard paths, docs, and configs without retaining test files in the PR diff.


zzhuoxin1508:
Two issues:

1. **base.py: the skip fires on *any* in-flight exception, not just wedged workers.** `sys.exc_info()` is also truthy for KeyboardInterrupt or a driver-side error with perfectly healthy workers, and `_wait_for_checkpoints()` is the only place that drains an in-flight `dcp.async_save` — so with `checkpoint_async` enabled, a Ctrl-C after an interval save can now silently leave the last checkpoint partially written, where the old code drained it in seconds.

2. **vendor/__init__.py: the comment states the opposite of what the code does.** With a real `ModuleSpec`, `find_spec("flash_attn")` now returns a truthy spec ("installed"), not "absent" — transformers only reports absent because its probe also checks `importlib.metadata.version`. Any bare `find_spec(...) is not None` probe will now misreport flash-attn as available.

celve:
> Two issues:
> 
> 1. **base.py: the skip fires on _any_ in-flight exception, not just wedged workers.** `sys.exc_info()` is also truthy for KeyboardInterrupt or a driver-side error with perfectly healthy workers, and `_wait_for_checkpoints()` is the only place that drains an in-flight `dcp.async_save` — so with `checkpoint_async` enabled, a Ctrl-C after an interval save can now silently leave the last checkpoint partially written, where the old code drained it in seconds.
> 2. **vendor/init.py: the comment states the opposite of what the code does.** With a real `ModuleSpec`, `find_spec("flash_attn")` now returns a truthy spec ("installed"), not "absent" — transformers only reports absent because its probe also checks `importlib.metadata.version`. Any bare `find_spec(...) is not None` probe will now misreport flash-attn as available.

Thanks — both good catches, addressed:

1) base.py skip: agreed, skipping on every exception would drop the in-flight async save on a healthy Ctrl-C. Switched skip → bounded flush — still drains, but bounds the ray.get with a timeout (exception-path only, default 120s), so it only gives up when a worker is actually wedged; primary exception preserved.

2) vendor comment: fixed to note find_spec now returns a truthy spec (transformers still stays on SDPA via importlib.metadata.version).

Verified on 1×8 H20: the rank-0 OOM prints its full traceback and the job exits at the bound (~1 min) instead of the 30-min NCCL hang.

leviking98z-rgb (commented):
Reviewed the complete current diff after syncing `main`. The original teardown design now strips `_ray_get_timeout` before worker dispatch, applies it only to `ray.get`, preserves the primary exception when a flush times out, and leaves the success path unbounded. During review I found and fixed a compatibility blocker in `PETrainer`: its `_wait_for_checkpoints` override did not accept the new timeout keyword, which would have raised `TypeError` even on successful PE teardown; both trained sides now preserve the base contract. The flash-attn fallback also has a valid `ModuleSpec`. Verified: 5 targeted pytest cases passed, targeted pre-commit and py_compile passed, `git diff --check` is clean, and current lint/body/title/CodeCC checks are green. I found no remaining code-level blocker.

leviking98z-rgb (approved):

PR #249, opened 2026-07-24
fix(rollout/fastvideo): verify worker sigma schedule

### LINKED ISSUE #248: [Bug] FastVideo rollouts do not verify the worker-used sigma schedule
## Problem

The FastVideo rollout engine pins `RolloutReq.sigmas`, converts them to the
static-shift pre-image expected by FastVideo, and sends that schedule to the
worker. After generation, however, it ignores FastVideo's authoritative
`trajectory_timesteps` and constructs the latent segment using the original
requested sigmas.

If the worker changes, truncates, or otherwise transforms the schedule, UniRL
continues replaying the trajectory as though the requested schedule was used.
This can silently desynchronize rollout transitions from trainer-side log-prob
replay.

The other dedicated diffusion engines already verify their worker-echoed
schedules with `verify_engine_used_sigmas`.

## Evidence

The supported dependency is
`Zcchill/FastVideo@7fe1d7db9a0b8aebb46679e7924f597431f23665`.

In that revision:

- `TimestepPreparationStage` stores `scheduler.timesteps` in `batch.timesteps`.
- The WAN flow scheduler represents those values as
  `sigma * num_train_timesteps`.
- `DenoisingStage` records every loop timestep in
  `RLData.trajectory_timesteps`.
- The returned field has shape `[T]`; the terminal zero exists only in
  `scheduler.sigmas`.
- Collection is active whenever `RLData.enabled` is true, including replay
  mode where `collect_log_probs` is false.
- `MultiprocExecutor` returns `RLData` to UniRL.

UniRL currently consumes the returned trajectory but not its associated
timestep field.

## Expected behavior

FastVideo rollouts should fail before trainer replay when:

- the worker does not return `trajectory_timesteps`;
- the returned schedule has the wrong length;
- the returned schedule differs from `RolloutReq.sigmas`.

The adapter should account for FastVideo's representation by appending the
omitted terminal zero and letting the shared verifier normalize the
`num_train_timesteps` scale.

Invalid static-shift inversion inputs should also fail explicitly instead of
producing NaN or infinite wire values.

## Acceptance criteria

- Validate every FastVideo sample's returned schedule.
- Cover replay and native-logprob modes through the common `RLData.enabled`
  path.
- Reuse `verify_engine_used_sigmas`.
- Reject non-finite or non-positive shifts and invalid requested schedules.
- Preserve existing request, response, segment, and checkpoint formats.
- Run a WAN2.1 GPU smoke against the supported FastVideo revision when suitable
  hardware is available.

## Compatibility

The existing recipe already requires the supported FastVideo revision. Older
forks that do not return `trajectory_timesteps` should fail closed instead of
silently serving unverifiable trajectories.

morluto:
## Summary

FastVideo previously trusted the schedule pinned on `RolloutReq.sigmas` after
generation, even though the worker returns the timesteps its scheduler actually
used. A worker-side schedule change could therefore label a trajectory with the
requested sigmas and silently desynchronize trainer replay.

This change validates every returned sample with the repository's existing
`verify_engine_used_sigmas` helper. The FastVideo adapter:

- appends the terminal zero omitted from FastVideo's `[T]` echo;
- decodes both integer UniPC ticks and floating Euler timesteps with WAN21's
  model-owned `WAN21DiffusionStep.TIMESTEP_SCALE`;
- mirrors UniPC's integer truncation on the expected schedule before comparison;
- reads `scheduler.config.num_train_timesteps` whenever a generator is
  constructed or reconstructed through the executor's worker RPC; and
- requires every worker to report the same positive integer scale and requires
  that value to equal the WAN21 model contract.

`1000` is therefore defined once by the model path that conditions the WAN
transformer, rather than hard-coded inside the FastVideo adapter or inferred
from the sequence currently being verified. Worker metadata is an independent
consistency check, not a second source of truth.

The existing request-side SDE schedule generation and static-shift pre-image
logic are unchanged. The top-level `trajectory_timesteps` fallback remains
available when `RLData` does not carry the field.

## Related Issue

Fixes #248

## Test Plan

- `python -m py_compile unirl/models/wan21/diffusion.py unirl/rollout/engine/fastvideo/engine.py` — passed.
- `ruff format --check unirl/models/wan21/diffusion.py unirl/rollout/engine/fastvideo/engine.py` — passed.
- `ruff check unirl/models/wan21/diffusion.py unirl/rollout/engine/fastvideo/engine.py` — passed.
- `git diff --check` and full diff inspection — passed.
- Focused checks against FastVideo
  `7fe1d7db9a0b8aebb46679e7924f597431f23665` — passed:
  - real UniPC integer ticks `[1000, 833]` and real Euler floating timesteps
    round-trip to the expected normalized schedule;
  - a one-tick drift is rejected;
  - worker scales `500` and `4000`, worker disagreement, and an empty RPC result
    are rejected; and
  - the real scheduler metadata reader returns `1000`.
- End-to-end rollout/training smoke:

  ```bash
    --config-name diffusion/wan21/wan21_t2v_dancegrpo_fastvideo \
    num_devices=1 ++devices_per_node=1 num_rollouts=1 batch_size=1 \
    sampling.samples_per_prompt=2 sampling.num_inference_steps=2 \
    sampling.height=256 sampling.width=256 stack.num_updates_per_batch=1 \
    +enable_fsdp_offload=true +model_config.meta_init_transformer=true \
    backend.fsdp_cfg.activation_checkpointing=false \
    reward.backend._target_=fastvideo_smoke_reward.RampVideoRewardScorer \
    reward.backend.base_device=cpu \
    reward.backend.config._target_=fastvideo_smoke_reward.RampVideoRewardSpec \
    logging.report_to_wandb=false
  ```

  One NVIDIA H20 with `Wan-AI/Wan2.1-T2V-1.3B-Diffusers`, the recipe's
  `datasets/pickscore/train.txt` prompt source, and the FastVideo revision above.
  The worker reported
  `num_train_timesteps=1000 matches WAN21 model contract across 1 workers`,
  then completed one rollout and one optimizer update without a traceback
  (`ratio=1.0000±0.0000`).

No standalone unit-test file is included; the final change is the model
contract plus FastVideo wire adapter and worker metadata query, all exercised by
the focused checks and H20 smoke.

## Compatibility / Risk

No public configuration, checkpoint, request, response, or segment format
changes. WAN21's step kernel gains a public class constant for its existing
training-time timestep convention.

The existing WAN2.1 recipe already requires the FastVideo revision that returns
`RLData.trajectory_timesteps`. This change also relies on that pinned executor's
`collective_rpc` API and scheduler module. Missing metadata, invalid values,
worker disagreement, or a worker/model scale mismatch fail closed instead of
silently producing unverifiable rollouts.

Runtime cost is one small control-plane RPC per generator construction or wake,
plus one CPU schedule comparison per generated sample.

## Reviewer Notes

The representation boundary is:

- UniRL: normalized `[T+1]`, including terminal zero.
- WAN21 model conditioning: normalized sigma multiplied by the model-owned
  `TIMESTEP_SCALE` (`1000`).
- FastVideo UniPC: integer
  `trunc(sigma * scheduler.config.num_train_timesteps)` `[T]`, terminal omitted.
- FastVideo Euler: floating
  `sigma * scheduler.config.num_train_timesteps` `[T]`, terminal omitted.
- Worker scheduler scale: queried independently and required to match the model
  contract before generation.

SGLang's shared verifier behavior is unchanged. Its dynamic normalization
handles an echoed schedule's representation units; this FastVideo check also
enforces the stronger WAN21 worker/model contract because UniPC quantizes to
integer ticks.

The adapter remains private to FastVideo and delegates the final comparison to
the existing shared verifier. No new shared SDE API or duplicated verifier is
added.

No overlapping FastVideo or `trajectory_timesteps` issue or PR was found.
Issue #127 is the broader Video RL roadmap.

AI assistance was used for dependency investigation, implementation, review,
and draft preparation.

## Checklist


leviking98z-rgb:
Closing this PR in favor of a cleanly scoped replacement. Further investigation showed that accepting truncated integer UniPC timestep echoes masks a real rollout/replay precision mismatch. The replacement will make the UniRL sigma contract and a stateful UniPC step strategy authoritative, wire FastVideo through an engine adapter, preserve exact fail-closed schedule round-trip validation, and remove the integer-truncation and shift-preimage compatibility paths. This PR will not be merged as-is.

PR #251, opened 2026-07-25
fix(fastvideo): use UniRL canonical UniPC strategy

### LINKED ISSUE #248: [Bug] FastVideo rollouts do not verify the worker-used sigma schedule

leviking98z-rgb:
## Summary

Fix FastVideo rollout/replay drift by making UniRL own the deterministic UniPC solver state while keeping sigma construction in the existing `FlowMatchSchedulePolicy`.

The FastVideo adapter now:

- sends the already-shifted canonical sigma schedule to the worker verbatim, keeps `sigma * WAN21DiffusionStep.TIMESTEP_SCALE` as floating-point model timesteps, and verifies every worker-echoed schedule with the shared verifier;
- rejects a FastVideo scheduler scale that differs from WAN21's model-owned `1000` contract and rejects the legacy `DIFFUSIONRL_FASTVIDEO_DANCEGRPO_TIMESTEP_LONG` cast that would silently truncate model conditioning after schedule verification;
- routes configured exploration indices through FastVideo's existing Dance/Flow SDE helper and all remaining indices through `unirl.sde.UniPCStrategy`;
- resets UniPC multistep history across SDE jumps, slices deterministic placeholder log-prob columns before constructing `LatentSegment`, and fails closed for unsupported scheduler/config combinations.

This avoids the old integer-timestep loss (`833` versus `833.333...`) without moving model execution, SDE transition density, reward, or FlowGRPO into the scheduler layer. The UniPC implementation is reusable by other adapters, but this PR wires it only into FastVideo; trainside, SGLang, and vLLM-Omni continue to use Euler on deterministic non-SDE steps.

## Related Issue

Fixes #248.

## Test Plan

- The 16-case CPU regression suite passed before its PR-added files were removed, covering the golden trajectory, history reset, worker/runtime dispatch, external sigma handling, floating timestep enforcement, and WAN scale mismatch.
- Full changed-production-file hook suite passed, including YAML, Python AST, ruff check/format, private-key detection, and recipe target resolution.
- `python -m py_compile unirl/models/wan21/diffusion.py unirl/rollout/engine/fastvideo/_unipc.py` (passed)
- `git diff --check origin/main` (passed)
- Temporary CPU parity/reset harness before publication:
  - 120 configurations across solver orders 1/2/3, `bh1`/`bh2`, lower-order-final on/off, corrector-disable variants, and 1/2/3/5/14 steps;
  - 36 history-reset segments beginning at nonzero indices;
  - all outputs matched FastVideo `FlowUniPCMultistepScheduler` exactly (`max_abs_diff=0`).
  Both the focused regression files and the broader parity harness were removed after validation; the final PR diff contains no test files.
- Two end-to-end WAN2.1 FastVideo jobs on 1 x H20 using `Wan-AI/Wan2.1-T2V-1.3B-Diffusers`, `Zcchill/FastVideo@7fe1d7db9a0b8aebb46679e7924f597431f23665`, `datasets/pickscore/train.txt`, a deterministic lightweight reward, 2 samples, 4 denoising steps, 256 x 256 output, and one optimizer update:
  - replay old-logp mode: full rollout, replay, and update completed; `ratio=1.0000+/-0.0000`;
  - native old-logp mode with `rollout.config.native_logprob=true algorithm.old_logp_source=rollout`: full rollout, native old-logp consumption, replay, and update completed; `ratio=1.0000+/-0.0000`.
  The final guard-only commit does not alter this validated `num_train_timesteps=1000`, floating-timestep path.

## Compatibility / Risk

- No checkpoint, dataset, request/response, or segment format changes. `WAN21DiffusionStep.TIMESTEP_SCALE` is an additive public model contract for rollout adapters.
- FastVideo deterministic non-SDE steps intentionally change from its RL branch's hard-coded Euler fallback to the configured UniPC trajectory; generated samples can therefore change from the trainside recipe and from other rollout engines.
- The adapter requires `FASTVIDEO_WAN_SCHEDULER=unipc`, floating model timesteps, `num_train_timesteps=1000`, `prediction_type=flow_prediction`, `predict_x0=true`, `final_sigmas_type=zero`, and the supported FastVideo worker symbols. Unsupported or older forks fail closed instead of silently using a different trajectory.
- The runtime adapter depends on the pinned FastVideo integration surface. Dance and Flow are the supported SDE kernels for this path; other SDE types are rejected explicitly.

## Reviewer Notes

- This supersedes closed PR #249. That PR validated FastVideo's integer scheduler ticks but could not recover precision already lost by truncation. This PR removes that loss at its source while retaining the worker/model timestep-scale invariant.
- `UniPCStrategy` is engine-independent math, not a claim that all current rollout engines use UniPC. FastVideo still supplies its solver-order/type configuration; making that configuration and other engine adapters canonical is follow-up work.
- Duplicate-work check: searched open PRs and issues for `UniPC`, `FastVideo scheduler`, and related terms. No overlapping open change exists besides this PR; #248 is the tracked bug.
- `UniPCStrategy` is adapted under Apache-2.0 from FastVideo/Wan's `FlowUniPCMultistepScheduler`, itself based on Hugging Face Diffusers v0.31.0. SPDX, copyright, provenance, and modification notices are retained in the source.
- Focused CPU regressions and the larger 120-configuration parity harness were run for the canonical solver and FastVideo adapter, then kept out of the final PR diff.
- The complete nine-file diff against `origin/main` was reviewed, and no unrelated/generated artifacts are included.
- Suggested review order: `unirl/sde/unipc.py`, `unirl/rollout/engine/fastvideo/_unipc.py`, then the small model-contract and engine dispatch diffs.

## Checklist

- [x] I validated the implementation, docs, and configs; the final PR diff contains no test files and the prior regression results are documented above.


leviking98z-rgb:
Maintainer self-review at head `82c8721`: I reviewed the complete nine-file diff and cross-checked the runtime hooks against the pinned FastVideo integration (`7fe1d7d`). The canonical sigma/timestep contract, SDE-vs-UniPC dispatch, history reset, deterministic placeholder slicing, worker entrypoint patch, and fail-closed configuration guards are covered; I found no remaining code-level blocker. Verified: 16 targeted CPU tests passed, the changed-file pre-commit suite and py_compile passed, `git diff --check` is clean, and all current lint/body/title/CodeCC checks are green. Because this PR is authored by this account, GitHub still requires an independent approval.

haonan3 (approved):
LGTM

PR #252, opened 2026-07-25
fix(distributed): serialize worker-local NCCL transfers

CjhHa1:

Per maintainer policy, the separately restored in-tree test file has been removed again; no PR-added test file remains.

Independent current-head validation:

- the original out-of-tree ordering/inheritance regressions passed;
- real Ray + NCCL + `DevicePool` + `colocate_store` ring-localize runs on **2/4/6/8 H20 GPUs**, five iterations per topology: all passed with exact tensor values and source IDs checked;
- targeted pre-commit, `git diff --check`, and the seven-PR integration suite passed.

## Summary
- process worker-local NCCL device pairs deterministically instead of lazy-initializing every pair concurrently
- post each receiver before its matching sender and preserve within-pair tensor batching/rebinding

## Root cause
Ray workers execute blocking transport RPCs serially. Launching all unbatched P2P pairs at once let different ranks enter lazy two-rank NCCL communicator initialization in incompatible orders on the shared TCPStore, producing a rendezvous timeout and a stuck `Worker.transport_op`.

## Validation
- out-of-tree local ordering regression — 2 passed
- targeted pre-commit (including Ruff) and `git diff --check` passed
- real H20 `colocate_store` ring-localize matrix, 5 iterations each: 2/4/6/8 GPUs passed
- BAGEL unified full-FT, 8×H20, batch 8, 4 samples/prompt: 2/2 rollouts completed; image ratio `1.0000±0.0000`
- exact 6-GPU BAGEL run progressed beyond localize into backward; it then exposed a separate capacity OOM (reduce-scatter needed 1.74 GiB with 1.27 GiB free), whose secondary NCCL watchdog timeout is not a transport deadlock

## Test Plan
- [x] out-of-tree local ordering and backend inheritance regression (`2 passed`)
- [x] 2/4/6/8 GPU forced foreign-TensorRef ring transfers
- [x] 8-GPU BAGEL full-FT end-to-end rollout/training


leviking98z-rgb (commented):
Reviewed the full production diff and restored the in-tree regression coverage. The controller now processes device pairs in deterministic order, posts each receiver before its matching sender, waits per pair, and preserves batching, sliced shapes/dtypes, result-key association, and destination rebinding. I found no remaining code-level blocker. Verified: 2 targeted pytest cases passed, targeted pre-commit passed, git diff --check is clean, and the latest lint/body/title/CodeCC checks are green.

leviking98z-rgb (approved):

Issue #86, opened 2026-06-16
[Tracking] Add PPO value critic and GAE support


yhl48:
I would like to work on the roadmap item "Policy-gradient / PPO: value critic + GAE" from #25 

Proposed scope:
- inspect existing GRPO/DRPO advantage computation
- add value function/critic support
- implement GAE advantage computation
- add PPO-style value loss and relevant config knobs

Could you please confirm whether this should target AR first, diffusion first, or both? 

Please also let me know if there is anything I missed above, or if there is anything you would like me to take note of before implementation. Thanks! 

@CjhHa1

CjhHa1:
Hi, thanks for your proposal! You can start by  implementing algo on AR models. Have Fun Contributing! 😃

leviking98z-rgb:
Thanks for your contribution! We have merged your PRs mentioned in this RFC. 

Please complete a PPO training example at your convenience, and include the training recipe and curve.

yhl48:
@leviking98z-rgb thanks! I don't have access to any GPUs at the moment, not sure if I will be able to provide a curve, how do contributors usually get around this?

PR #254, opened 2026-07-25
feat(types): add GAE advantage computation

yhl48:
## Summary

Part 1/3 for #86. Adds pure-tensor Generalized Advantage Estimation (GAE) under `unirl/types/advantages.py`, alongside the existing GRPO advantage path in `RolloutTrack.compute_advantages`. This keeps advantage computation in `types/` (pre-train signal) rather than `algorithms/` (loss classes). No trainer or model wiring yet — that lands in follow-up PRs.

AI-assisted implementation; checked open PRs — no overlapping GAE work found.

## Related Issue

Part of #86

## Test Plan

- `pytest tests/types/test_advantages_gae.py -q` on Python 3.12 — 12 passed, including 1D/2D nonzero-padding mask regressions.
- `SKIP=no-commit-to-branch pre-commit run --files unirl/types/advantages.py tests/types/test_advantages_gae.py` — all hooks passed.
- `git diff --check origin/main...HEAD` — clean.

## Compatibility / Risk

N/A — new module and tests only; no config, checkpoint, or API changes to existing training paths.

## Reviewer Notes

Follow-ups (separate PRs): AR value head + replay values (part 2), `PPO` algorithm with value loss + recipe (part 3). GAE uses episodic bootstrap `V_{T+1}=0` and optional validity masks for packed sequences.

## Checklist

### DIFF
--- a/tests/types/test_advantages_gae.py
+++ b/tests/types/test_advantages_gae.py
@@ -1,0 +1,109 @@
+"""CPU unit tests for GAE advantage computation."""
+
+from __future__ import annotations
+
+import math
+
+import pytest
+import torch
+
+from unirl.types.advantages import compute_gae_advantages
+
+
+def test_gae_hand_computed_lambda_one() -> None:
+    """Sparse terminal reward; λ=1; hand-computed backward pass."""
+    rewards = torch.tensor([0.0, 0.0, 1.0])
+    values = torch.tensor([0.2, 0.5, 0.8])
+    advantages, returns = compute_gae_advantages(
+        rewards, values, gamma=1.0, gae_lambda=1.0
+    )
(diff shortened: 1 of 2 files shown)


yhl48:
@CjhHa1 @haonan3

leviking98z-rgb (commented):
Reviewed the full diff after fixing the GAE boundary cases. The implementation now prevents bootstrapping from padded values in both 1D and 2D paths, avoids a host sync in the 1D recurrence, and the finite-horizon gamma=1 test now uses a correct value baseline. I found no remaining code-level blocker. Verified: 12 targeted pytest cases passed (including nonzero-padding regressions), targeted pre-commit passed, git diff --check is clean, and all current head checks are green.

leviking98z-rgb (approved):

PR #255, opened 2026-07-25
refactor(types): rename RolloutReq.stage_config to task_config

CjhHa1:

Final review changed this from a breaking rename to a compatibility migration. This update supersedes the older hard-error and “mechanical rename” statements below.

- `task_config` is the only serialized `RolloutReq` field.
- Deprecated `stage_config` remains a readable/writable property alias; passing it to the constructor maps to `task_config` and emits a `DeprecationWarning`.
- Supplying conflicting old and new constructor values fails clearly.
- `train_diffusion` prefers the canonical recipe key, accepts the old key with a `FutureWarning`, and rejects ambiguous configs containing both.
- The old alias is not serialized by `Batch`.

Current-head validation:

- No PR-added test file remains; canonical and legacy paths pass a non-persisted current-head compatibility smoke.
- Synthetic integration with #186 and #242: both legacy recipes compose and their routing metadata is preserved.
- Clean merge with latest main and all 21 pairwise combinations among the seven reviewed candidate PRs.
- Targeted pre-commit, `git diff --check`, and the seven-PR integration suite passed.

Compatibility risk is now low: existing callers and recipes continue to work during the deprecation window, while new code can migrate to `task_config` immediately.

## Summary

`RolloutReq.stage_config` carries per-request **task routing** metadata (`task`, `bot_task`, `sys_type`, `chat`, `ar`) and names no pipeline stage — but the repo now spells three unrelated concepts with a `stage` prefix, and this was the odd one out:

| Name | Occurrences | Meaning |
|---|---|---|
| `stage_attr` / `stage_attrs` | ~230 | which model stage an algorithm / engine binds to (`DiffusionStage`, `ARStage`) |
| `stage_configs/` | 24 | vllm-omni's **inference engine** sub-stage YAMLs |
| `stage_config` | 43 | per-request task routing — *this* |

`stage_config` vs `stage_configs` differ by one character and mean entirely different things. Renamed to `task_config`, which matches what the field actually drives (`BagelPipeline._resolve_task`, `HunyuanImage3Pipeline.generate`).

The rename is mechanical (43 call sites / 19 files), plus three things worth a look:

1. **The old key remains as a deprecated compatibility path.** `train_diffusion` maps legacy `stage_config` to `task_config` with a `FutureWarning`, while rejecting only ambiguous configs that provide both. `RolloutReq` exposes a non-serialized alias so in-flight callers do not silently lose routing metadata.
2. **`RolloutReq`'s docstring now states what the field is not**, so the next reader does not have to rediscover the `stage_attr` / `stage_configs/` distinction.
3. **`unirl/README.md` drops the training-loop sense of "stage"** — rollout / reward / algorithm / train / weight-sync are loop *components*. No Python code ever used "stage" that way (the trainers already say `rollout → reward → advantage → optimizer step`), so this was the only place the word was overloaded in prose. "Stage" now means exactly one thing repo-wide: a model pipeline's trainable unit.

Two drive-by fixes in the same blast radius: the `stage_ar` local in `sglang/utils/sampling.py` (it holds `task_config["ar"]`) becomes `task_ar`, and a stale `stage_params["task"]` docstring reference in `hunyuan_image3/pipeline.py` — a field that never existed — becomes `task_config["task"]`.

## Related Issue

N/A

## Test Plan

- The PR-added regression file was removed per maintainer policy; current behavior was validated with a non-persisted inline smoke check.
- `SKIP=no-commit-to-branch pre-commit run --all-files --show-diff-on-failure` — all hooks passed.
- `git diff --check origin/main...HEAD` — clean.
- A repository-wide review confirmed remaining `stage_config` uses are deliberate compatibility points or in-flight recipes.
- Full training smoke was not rerun; compatibility-sensitive routing passed the current-head inline smoke and synthetic integration checks.

## Compatibility / Risk

- **Recipe key migration (user-facing).** New recipes should use top-level `task_config`. Legacy `stage_config` remains accepted with a `FutureWarning`; configs specifying both are rejected to avoid ambiguity.
- **Not an external protocol key.** The field crosses the Ray boundary as part of `RolloutReq`, but SGLang and vllm-omni receive its *values* mapped into their own payload fields — `"stage_config"` never appeared on any HTTP wire — so there is no engine or reward-service compatibility surface.
- Not persisted in checkpoints. No data format, resource, or migration impact.
- Will conflict textually with any open PR that touches the same `stage_config` call sites; conflicts resolve by taking the new name.

## Reviewer Notes

Worth confirming the deprecation-window behavior: the canonical key is documented, legacy-only configs warn and continue, and ambiguous dual-key configs fail clearly.

The bulk of the diff is a mechanical identifier rename; review focus is `unirl/types/rollout_req.py`, `unirl/train_diffusion.py`, `unirl/README.md`, and the compatibility call site in `unirl/models/hunyuan_image3/modes/it2i.py`.

## Checklist


leviking98z-rgb (commented):
Reviewed the full diff against main after the task_config regression test was added. The rename is mechanically complete, the intentional compatibility break is documented and fail-fast, and I found no remaining code-level blocker. Verified: targeted pytest 2 passed, targeted pre-commit passed, diff check clean, and current head checks are green.

leviking98z-rgb (approved):

PR #258, opened 2026-07-26
fix(distributed): fail fast on rank errors

CjhHa1:
## Summary
- collect Ray actor results in completion order while preserving rank-ordered outputs
- surface the first rank-local exception immediately and best-effort cancel pending sibling tasks
- poison the DevicePool after a distributed failure so checkpoint/teardown handles cannot enter another blocking RPC

## Root cause
`ray.get(refs)` waits for every rank result before raising any contained exception. If rank 0 OOMs before an FSDP collective while peers have already entered NCCL, the original OOM remains hidden until peer watchdogs time out roughly 30 minutes later. Teardown can then issue another distributed call and wait again.

Regular Ray actor tasks blocked in native NCCL cannot be force-cancelled safely. This change therefore makes the driver fail fast, marks the shared pool terminal, and prevents further Handle dispatch; actor cleanup remains owned by Ray when the driver exits.

## Validation
External tests are intentionally kept out of tree.

- mock success/error/order/poison regressions plus real local Ray actor integration: 5 passed
- Ruff check/format, compile, `git diff --check`, and repository hooks passed
- exact BAGEL 6-GPU full-FT OOM reproduction:
  - before: original OOM surfaced after ~50 minutes, following the 30-minute NCCL reduce-scatter watchdog
  - after: `RayTaskError(OutOfMemoryError)` surfaced in 1201s (~20 minutes total startup/generate/backward), immediately when rank 0 failed
  - DevicePool was poisoned at `UnifiedModelTrainStack_0.train_track`
  - checkpoint teardown was rejected immediately instead of dispatching another RPC
  - no NCCL watchdog wait; GPUs returned to the occupy workload

The e2e validation checkout also carried #252 locally to bypass the earlier, orthogonal TensorRef P2P lazy-communicator deadlock. This PR remains independently based on `main`.

## Test Plan
- [x] completion-order result collection preserves rank order
- [x] original first exception is re-raised
- [x] pending actor tasks receive best-effort cancellation
- [x] poisoned pool blocks subsequent Handle calls
- [x] real Ray actor first-error latency
- [x] exact 6-GPU OOM failure latency and teardown behavior

### DIFF
--- a/unirl/distributed/group/device_pool.py
+++ b/unirl/distributed/group/device_pool.py
@@ -72,10 +72,29 @@ class DevicePool:
         self._worker_id_to_slot: Dict[str, int] = {}
         self._device_to_workers: Dict[int, List] = {}  # device_id → [slot0, slot1, ...]
         self._worker_by_id: Dict[str, Any] = {}  # worker_id → handle
+        self._poison_reason: Optional[str] = None
 
     @property
     def num_gpus(self) -> int:
         return self.num_devices
+
+    def mark_poisoned(self, *, rank: int, role_name: str, method_name: str, error: BaseException) -> None:
+        """Poison the pool with the first rank-local failure context."""
+        if self._poison_reason is not None:
+            return
+        error_text = str(error)
+        error_summary = error_text.splitlines()[0] if error_text else "<no message>"
+        self._poison_reason = (
+            f"rank={rank}, role={role_name!r}, method={method_name!r}, error={type(error).__name__}: {error_summary}"
+        )
+        logger.error("DevicePool poisoned: %s", self._poison_reason)
+
+    def assert_usable(self) -> None:
+        """Reject distributed RPCs after any rank-local actor task failure."""
+        if self._poison_reason is not None:
+            raise RuntimeError(
+                f"DevicePool is poisoned; refusing to issue another distributed RPC ({self._poison_reason})"
+            )
 
     @property
     def transport_cls(self) -> type:
@@ -292,6 +311,8 @@ class DevicePool:
         slot_id: Optional[int] = None,
     ) -> Handle:
         """Create a Handle for role_cls on this pool."""
+        self.assert_usable()
+
         from unirl.distributed.group.placement import current_placement
 
         if device_ids is None:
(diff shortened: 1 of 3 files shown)


leviking98z-rgb (changes requested):
The failure mode is real and the completion-order gather / pool-poisoning direction is valuable, but this branch needs to be ported to the current async architecture before it can merge.

Two blocking changes are required:

1. `unirl/rollout/async_runtime.py` was removed by #287. Async rollout now uses `Handle.launch_nowait()` → `PendingHandleCall` → `RolloutPool`. In current main, `PendingHandleCall.ready()` returns true only after every rank ref is ready, and `RolloutPool` calls `result()` only then. If one rank OOMs while peers are stuck in NCCL, `ready()` remains false forever, so the new completion-order gather in `result()` is never reached. Please rebase and make the pending-call polling path surface any ready rank-local exception immediately while still preserving rank-ordered successful results.

2. Pool usability needs to be enforced at the current shared launch boundary. After #287, blocking calls, `launch_nowait()`, and related paths share `_launch_call()`, while slot calls have their own launch path. Guarding only the old `handle_fn` / deleted dispatcher leaves new non-blocking dispatch able to bypass poisoning. Please place the check in the central current entry points so no distributed RPC can be issued after the first observed rank failure.

The synchronous completion-order collection, whole-gather timeout preservation, best-effort cancellation, and teardown poisoning behavior otherwise look directionally correct. Please rebase onto current main and rerun the focused synchronous + async rank-failure tests against `PendingHandleCall`/`RolloutPool`; this should remain open rather than be closed as superseded.

CjhHa1:
Rebased onto current `main` and ported the fail-fast path onto the post-#287 async architecture.

1. `PendingHandleCall.ready()` now inspects currently-ready rank refs and raises the first rank-local exception immediately (poisoning the pool and best-effort cancelling siblings). `RolloutPool` already treats a `ready()` exception as a pool failure, so a rank OOM no longer waits for peers stuck in NCCL. `wait()` / `result()` / `_resolve_call` gather through `get_actor_results`, so the blocking path keeps completion-order collection, whole-gather timeout, and rank-ordered successes.

2. `assert_usable()` sits on the shared launch boundaries: `_launch_call()` (blocking `handle_fn` + `launch_nowait()`) and `Slot.launch()`. `create_remote` / `Handle.__init__` / `initialize` are also guarded so no distributed RPC is issued after the first observed rank failure.

Focused mock + `PendingHandleCall` / `RolloutPool` + local Ray first-error tests: 10 passed. External tests remain out of tree.

leviking98z-rgb (approved):

PR #260, opened 2026-07-26
feat(qwen3-omni): enable audio-in-video for qwen3-omni thinker AR (text + video + audio in video)

cboss6:
## Summary

This PR extends the Qwen3-Omni Thinker RL path in UniRL from video-only to **audio-in-video** reinforcement learning:

- trains `Qwen3-Omni-30B-A3B-Instruct` on Daily-Omni audio/video multiple-choice reasoning with GSPO and LoRA;
- fuses the mp4 audio track into the video stream through TMRoPE (`use_audio_in_video=true`) on both the rollout and the training-replay side;
- serves Thinker-only rollout through vLLM-Omni with TP=4 and time-shares the four GPUs with the FSDP actor via sleep/wake;
- synchronizes LoRA weights to the anchored rollout engine before every rollout and verifies the readback;
- aligns the audio/video token layout, processor features and policy log-probs between the HF replay path and the vLLM-Omni rollout.

The main recipe is:

`examples/ar/qwen3_omni_audio_video_gspo_lora_vllm_omni_1x4.yaml`

An 8-GPU variant with the same rollout topology is:

`examples/ar/qwen3_omni_audio_video_gspo_lora_vllm_omni_1x8.yaml`

The rollout topology is defined in:

`unirl/rollout/engine/vllm_omni/stage_configs/qwen3_omni_thinker_only_rl_1x4.yaml`

## Related PR
[feat(qwen3-omni): add qwen3-omni thinker AR using vllm-omni as rollout backend (text + video)
](https://github.com/Tencent-Hunyuan/UniRL/pull/235)
[fix(vllm-omni): stop engine subprocesses outliving the driver](https://github.com/Tencent-Hunyuan/UniRL/pull/265)

### Process topology

`unirl/train_ar.py` initializes Ray with an explicit `runtime_env` (so `LD_LIBRARY_PATH`, HF offline flags and `UNIRL_*` debug variables reach every actor and every spawned vLLM worker), then constructs `ARTrainer`. The data source stays on the driver; every model-owning component becomes a Ray role.

```mermaid
flowchart LR
  subgraph DRV["Driver process"]
    DS["MultimodalRLDataSource<br/>train.jsonl / val.jsonl"]
    TR["ARTrainer"]
  end

  subgraph TRAIN["Shared Ray Workers · GPU 0-3 · FSDP DP=4"]
    BD["Qwen3OmniBundle<br/>Thinker + AutoProcessor"]
    PL["Qwen3OmniPipeline<br/>chat_template · ar"]
    BE["FSDPBackend<br/>LoRA r32 a64 · bf16"]
    AL["GSPO"]
    ST["TrainStack"]
    RW["RewardService<br/>MCExactMatch"]
    WS["RemoteLoraWeightSync"]
  end

  subgraph ANCH["Anchored Ray Worker · GPU 1"]
    EN["VLLMOmniRolloutEngine"]
    AD["Qwen3OmniThinkerAdapter<br/>AutoProcessor · token compression"]
  end

  subgraph OMNI["vLLM-Omni subprocess · TP=4 on GPU 0-3"]
    SG["Stage 0 · thinker AR<br/>OmniARScheduler · worker_type=ar"]
    WX["Qwen3OmniARWeightSyncExtension<br/>x4 TP ranks"]
  end

  DS -->|RolloutInputs| TR
  TR -->|RolloutReq| EN
  EN --> AD
  AD -->|"prompt_token_ids + multi_modal_data"| SG
  SG --> WX
  EN -->|"RolloutResp.tracks.ar"| TR
  TR -->|"score_and_attach"| RW
  TR -->|"train_track"| ST
  ST --> AL
  AL -->|replay| PL
  ST --> BE
  BE --> BD
  WS -->|"extract LoRA (DTensor all-gather)"| BE
  WS -->|"set_lora_from_tensors_copy"| WX
```

Key placement facts:

- `rollout_anchor_device: 1` puts the engine in a **single** Ray actor. TP>1 is not expressed through UniRL's device pool yet, so the engine actor pops `CUDA_VISIBLE_DEVICES` and lets the stage YAML pin `devices: "0,1,2,3"` itself.
- The anchor must not be device 0, because `RemoteLoraWeightSync` pushes from rank 0 and a colocated engine would produce a synchronous self-call.
- TP must divide the audio tower's **20** attention heads, so the valid choices are 1, 2, 4, 5, 10, 20. TP=4 is the only one that fills a 4-GPU node.

### One training step

`weight_sync_interval: 1`, so LoRA is pushed before every rollout except the first. `enable_fsdp_offload: true` makes the trainer the single owner of CPU/GPU placement (FSDP's own `CPUOffloadPolicy` is disabled so the two mechanisms cannot fight over the frozen encoders).

```mermaid
sequenceDiagram
  participant T as ARTrainer
  participant F as FSDPBackend
  participant S as RemoteLoraWeightSync
  participant E as vLLM-Omni engine
  participant R as RewardService
  participant K as TrainStack + GSPO

  T->>F: onload()
  T->>S: extract() (all ranks all-gather LoRA)
  T->>F: offload()
  T->>E: wake_up()
  T->>S: push() -> set_lora_from_tensors_copy
  T->>E: generate(RolloutReq, 4 prompts x 8 samples)
  E-->>T: TextSegment(tokens, rollout_log_probs) + decoded + conditions
  T->>E: sleep()
  T->>R: score_and_attach()
  R-->>T: rewards
  T->>T: compute_advantages(scope=group)
  T->>F: onload()
  T->>K: train_track()
  K->>K: replay -> frozen pi_old, GSPO loss, optimizer_step
  T->>F: offload()
```

Between steps the steady state is **FSDP on CPU, rollout asleep**. Evaluation reuses the same session helper but keeps it open across all eval batches: one `extract → offload → wake → push`, then every eval batch, then a single `sleep`, so a multi-batch eval no longer pays a wake/sleep cycle per batch.

### Audio-in-video data path

This is the part that is new relative to the video-only recipe. The same mp4 must produce the *same* interleaved token layout on both sides, otherwise the GSPO importance ratio compares two different sequences.

```mermaid
flowchart TB
  MP4["mp4 clip (media_refs: video/prompt)"]

  subgraph RO["Rollout side · Qwen3OmniThinkerAdapter"]
    RD["pyav decode: frames @ fps=2.0 + mono waveform"]
    RP["AutoProcessor(use_audio_in_video=true, truncation=true,<br/>fps, size=video_max_pixels)"]
    RE["expanded prompt ids + pixel_values_videos + input_features"]
    RC["token compression -> raw prompt ids"]
    RV["vLLM-Omni processor re-expansion<br/>mm_processor_kwargs: use_audio_in_video, second_per_grid_ts"]
  end

  subgraph HF["Train side · Qwen3OmniChatTemplateStage"]
    HD["pyav decode + extract_audio_from_video_pyav"]
    HP["AutoProcessor(use_audio_in_video=true, truncation=true, ...)"]
    HC["Qwen3OmniARConditions"]
    HR["replay: get_rope_index -> TMRoPE 3D position ids"]
  end

  MP4 --> RD --> RP --> RE --> RC --> RV
  MP4 --> HD --> HP --> HC --> HR
  RE -.->|"cached encodings become replay conditions"| HC
  RV -.->|"position-by-position identical prompt ids"| HR
```

Three defects had to be fixed to make the two branches agree:

| Symptom | Root cause | Fix |
| --- | --- | --- |
| vLLM audio features one frame longer than HF (3001 vs 3000), then `Encoder cache miss` | vLLM-Omni moves the top-level `truncation` flag into `audio_kwargs` but still reads the now-missing top-level flag when rebuilding `audio_feature_lengths` | pass `truncation=True` explicitly on both processors and clamp the rebuilt lengths/mask to `n_samples // hop_length` |
| Same token count, 754 positions mismatched: HF interleaves 220 video / 13 audio, vLLM 220 video / 26 audio | vLLM-Omni falls back to a hardcoded `2.0 s` per temporal grid when `second_per_grid_ts` is absent | send `second_per_grid_ts = [temporal_patch_size / video_fps]` per audio-in-video request |
| `k3` stuck at `1.47 ~ 1.60` on a greedy single-sample probe | replay divided logits by `sampling_temperature=1e-6`, which is not greedy log-prob semantics | non-positive temperature now means unscaled `T=1` log-probs, matching vLLM greedy output |

After the fixes, on a fixed single sample: 7532 processor-expanded tokens → 101 compressed tokens → 7532 re-expanded tokens with **0 mismatches**, identical audio/video encoder input shapes, element-wise identical text embeddings, and a fused-embedding cosine similarity of `0.999585` (HF SDPA vs vLLM TP FlashAttention keeps a small floating-point gap). `k3` drops to `2.85e-3`. 

## Installation

This recipe uses UniRL's CUDA 12.9 vLLM environment:

```bash
uv venv --python 3.12 --seed .venv
source .venv/bin/activate

export VLLM_USE_PRECOMPILED=1
uv pip install -e ".[vllm,train,infer]"
```

The currently pinned stack includes:

- Python 3.12 or 3.13
- PyTorch 2.11.0 with CUDA 12.9
- vLLM 0.22.0
- vLLM-Omni 0.22.0
- Transformers 4.57.6

## Prepare the dataset

The recipe consumes Daily-Omni audio/video multiple-choice data in UniRL JSONL format.

```bash
python scripts/convert_daily_omni_dataset_format_to_unirl.py \
    --train-input /path/to/daily_omni_av_train.jsonl \
    --val-input /path/to/daily_omni_av_val.jsonl \
    --out-dir /path/to/daily_omni_av
```

The converter:

- extracts the user question text from the source chat messages;
- stores **only** the video path in `media_refs` (`{"modality": "video", "role": "prompt", "uri": ...}`) — Qwen3-Omni pulls the audio track out of that same file when `use_audio_in_video=true`, so no separate audio reference is needed;
- records the ground-truth letter in `metadata.answer`, plus `video_id` and `qa_type`;
- assigns stable `prompt_id`s of the form `daily_omni_av:<split>:<index>:<video_id>`;
- skips rows whose video file is missing unless `--keep-missing` is passed.

Two properties of the prompt and eval set matter for the reward and for signal quality:

- the prompt asks for `<think>...</think>` reasoning followed by an explicit `The answer is [X]` phrase. `<think>` is SFT-native for this checkpoint, so chain-of-thought tokenization stays cheap and `MCExactMatchRewardScorer` with `require_answer_phrase: true` picks the answer up without regex guessing;
- the validation split is a **mixed-variance subset**: prompts where the base model at `fps=2` scores between 1/8 and 7/8 correct. The raw validation set is dominated by all-correct (~44%) and all-wrong (~8%) prompts, which inflate `eval/acc` while carrying no group-relative gradient.

Export the resulting paths before launch:

```bash
export DATA_PATH=/path/to/daily_omni_av/train.jsonl
export EVAL_DATA_PATH=/path/to/daily_omni_av/val.jsonl
```

## Prepare the model

Use a local copy of the Instruct checkpoint:

```bash
export QWEN3_OMNI_PATH=/path/to/Qwen3-Omni-30B-A3B-Instruct
```

Only `Qwen3OmniMoeThinkerForConditionalGeneration` is loaded for training. The vision and audio towers stay resident for multimodal preprocessing but are frozen; the Talker and Code2Wav stages are never instantiated.

The default preprocessing configuration is:

- video sampling rate: 2 FPS;
- maximum video pixels per frame: 262,144;
- maximum prompt length: 16,384 tokens;
- **video audio input enabled** (`use_audio_in_video: true`);
- BF16 model and training precision.

Both the training pipeline and the rollout engine config carry these values, because the anchored rollout actor is not a sibling of the training workers and must reconstruct the same processor settings independently.

## Run training

```bash
DATA_PATH=/path/to/daily_omni_av/train.jsonl \
EVAL_DATA_PATH=/path/to/daily_omni_av/val.jsonl \
PYTHONPATH=$(pwd) \
python -m unirl.train_ar \
    --config-name=ar/qwen3_omni_audio_video_gspo_lora_vllm_omni_1x4
```

The default training geometry is:

- one node with 4 training devices, FSDP DP=4;
- batch size 4, i.e. 1 prompt per rank;
- 8 samples per prompt, producing 32 responses per update;
- maximum response length 2,048 tokens;
- one optimizer update per rollout batch, micro-batch size 1;
- LoRA weight synchronization before every rollout;
- FSDP state offloaded to CPU while the colocated rollout engine is awake;
- 500 rollouts, constant LR schedule.

Evaluation runs a baseline pass before training and then every `eval_interval` rollouts (5 for the 4-GPU recipe, 20 for the 8-GPU one; `eval_interval: 0` disables it). `eval_num_prompts: -1` walks the full validation file in batches of `eval_batch_size` (default 8) at `eval_samples_per_prompt: 1` and `eval_temperature: 1.0`, and logs `eval/acc` and `eval/reward`.

The vLLM-Omni stage uses:

- Thinker-only AR execution (`worker_type: ar`, `OmniARScheduler`);
- TP=4 on devices 0–3 with the `mp` executor backend;
- `Qwen3OmniMoeThinkerForConditionalGeneration`, BF16 safetensor loading;
- chunked prefill, up to 64 sequences and 32,768 batched tokens;
- 35% rollout GPU memory utilization, leaving headroom for the training allocator;
- `limit_mm_per_prompt: {audio: 1, video: 1, image: 0}` — audio fusion is requested **per request**, because enabling `use_audio_in_video` globally breaks vLLM's video-only dummy profiling during startup;
- `mm_processor_cache_gb: 0`, since every RL prompt is seen once and the cache only costs memory;
- LoRA rank support up to 64;
- `logprobs_mode: processed_logprobs` so rollout log-probs are directly comparable to replay;
- sleep mode for actor/rollout colocation.

## What is trained

Only the Qwen3-Omni Thinker is optimized.

LoRA configuration:

- rank: 32;
- alpha: 64;
- dropout: 0;
- target modules: `q_proj`, `k_proj`, `v_proj`, and `o_proj`;
- `exclude_modules: ".*visual.*|.*audio_tower.*"` — PEFT injects adapters purely by module name, so the frozen encoders have to be excluded explicitly;
- vision and audio tower parameters frozen.

The actor is wrapped with FSDP at `Qwen3OmniMoeThinkerTextDecoderLayer` boundaries and uses:

- BF16 mixed precision;
- full sharding with resharding after forward and forward prefetch;
- gradient checkpointing;
- Adam with learning rate `1e-5`;
- maximum gradient norm 1.0.

GSPO uses group-level advantage normalization with clipping ranges `1e-3` and `1.33e-3`. `old_logp_source: replay` freezes a train-side pre-update replay as π_old, so the importance ratio is computed between two HF forward passes and is not polluted by rollout/engine numerics. The vLLM rollout log-probs remain available separately and feed the `k3` train/rollout consistency metric — that metric is what caught every alignment bug listed above.

Rewards are computed by `MCExactMatchRewardScorer` with `require_answer_phrase: true` and `graded_format_reward: false`: the answer letter is only accepted from `<answer>X</answer>` or an explicit `answer is X` / `option: X` phrase, and scoring is binary. `truncated_reward: keep` leaves length-capped generations at their earned reward instead of zeroing them.

LoRA parameters are copied to the vLLM-Omni Thinker under the `thinker.` prefix using the byte-copy transport (TP-safe), and the readback checksums are verified on every TP rank after each push. If a re-push after wake fails, the engine stays marked offloaded and the next `generate()` raises, so a rollout can never silently fall back to the base model.

## Preliminary results
image

## Related Issue

N/A

## Compatibility / Risk

- The recipe requires the vLLM-Omni patches applied at engine boot (`unirl/rollout/engine/vllm_omni/patches/qwen3_omni.py`): AutoModel registration, HF processor/tokenizer shims, LoRA request passthrough, per-request AR seeds, and the audio truncation length fix. The last one compensates for an upstream vLLM-Omni bug; if upstream fixes it, the patch becomes a no-op guard rather than a conflict.
- TP is constrained to divisors of the audio tower's 20 attention heads. This is not asserted in UniRL config validation — a bad value fails inside vLLM at startup.
- `rollout_anchor_device` must be nonzero whenever `RemoteLoraWeightSync` is used; the trainer raises early otherwise.
- Orphaned `VLLM::Worker_TP*` processes can survive an aborted run and hold ~6 GB each; `ray stop --force` does not reap them because Ray already considers the task finished. Check for them before relaunching on the same GPUs.
- Existing video-only and text-only Qwen3-Omni recipes keep working: `use_audio_in_video` defaults to false and the audio-specific processor kwargs are only emitted when a waveform is actually present.

## Reviewer Notes

- The TP>1 anchored rollout path is still the temporary construction migrated from the unified-model trainer; replacing it with first-class TP/DP/PP support in the device pool is follow-up work, and it is the main reason the anchor-device and self-deadlock constraints exist.
- `second_per_grid_ts` is currently derived as `temporal_patch_size / video_fps`, which is correct for uniform sampling at a fixed FPS. If variable-rate or pre-sampled frame inputs are added later, this needs to become a real per-video timestamp list.
- Worth reviewing first: the audio truncation patch, the per-request `mm_processor_kwargs` construction in the adapter, and the eval-session lifecycle in `ARTrainer.evaluate`.

## Checklist


leviking98z-rgb:
Update after the branch-history cleanup (2026-07-28): the PR is now one core commit based on current `main`; the existing PR body is intentionally preserved unchanged.

Retained in the minimal diff:
- Qwen3-Omni audio-in-video preprocessing, HF replay fusion, and per-sample mixed audio/silent MRoPE;
- vLLM-Omni placeholder compression, audio truncation clamp, MRoPE parity, and effective-FPS timing after frame capping;
- the 64-frame recipes, official Daily-Omni converter, strict answer-phrase reward, and whole-conversation duplicate preprocessing reuse.

Removed from the rewritten history:
- generic process/fate-sharing/shutdown changes and all overlap with PR #265;
- evaluation-session/lifecycle refactors, diagnostic logging, generic media validation, legacy compatibility branches, and final-answer-selection optimization;
- the redundant `TextSegment.rollout_log_probs` declaration from the old history. No test files were added.

Validation on the rewritten head:
- `SKIP=no-commit-to-branch pre-commit run --all-files --show-diff-on-failure` — passed;
- both 1x4 and 1x8 Hydra compositions — passed;
- targeted Transformers 5.6.2 / vLLM-Omni 0.20 regressions — stock vLLM MRoPE mismatched 8 columns, patched result mismatched 0;
- real processor smoke — 120 frames capped to 64, effective FPS `1.0588235`, `second_per_grid_ts=1.8888889`, and four fork siblings reused one complete encoding;
- real Qwen3-Omni-30B H20 mixed replay — audio + silent rows used separate MRoPE modes; replay and backward passed with finite gradients;
- real TP=4 audio-in-video rollout — four requests generated successfully and returned audio/video replay conditions;
- two-round 4×H20 recipe smoke — `rollout → reward → replay/train → LoRA sync → second rollout` completed and shut down cleanly; both rounds reported reward `1.0` and `k3=2.30e-06`.

Note: the preserved cboss6 body has an empty `## Test Plan` section. The repository's PR Body workflow only reads the body (not comments), so that one check cannot become green without editing the body.

leviking98z-rgb (approved):

PR #263, opened 2026-07-27
refactor(types): retire stage_config/task_config onto Part.control

CjhHa1:
## Summary
- Rebase onto current `main` and drop the stale `RolloutReq` / HI3 `it2i` hunks (those files were rewritten or deleted by #214 / #186).
- Remove the `_resolve_task_config` remaps.
- Recipe overlay is `control:` → root `Part.control`. Trainer constructors and `prepare_input_sample` use the same name. No leftover `stage_config` / `task_config` alias or reject path.

## Test plan
- [ ] Recipes under `examples/` that pin a task use `control:` (bagel edit, HI3 t2i/it2i). vllm-omni `stage_configs/` assets are unrelated.
- [ ] `control.task: it2i` still reaches `parts[0].control["task"]` on the diffusion and unified entrypoints.


haonan3:
Rebase + rescope is needed:

- **`unirl/types/rollout_req.py`**: file was deleted entirely by #214 (RolloutReq → Sample/Part; `stage_config` semantics now live in `Part.control`). This hunk is moot and is the source of the conflict — drop it.
- **`unirl/models/hunyuan_image3/modes/it2i.py`**: rewritten by #214/#186; no `stage_config`/`task_config` reference remains. Drop this hunk too.
- **`unirl/train_diffusion.py`**: still valid — keep as is.
- **Please also cover `unirl/train_unified_model.py`**: an identical `_resolve_task_config` shim landed there via #186 after this PR was cut. Apply the same removal + fail-fast there so the deprecation is fully retired.

Notes: recipes under `examples/` are already clean (no `stage_config:` keys), so the fail-fast is a pure safety net. The RolloutReq-related test-plan items are obsolete now — the remaining checks are just the two entrypoints erroring on a leftover `stage_config:` key.

celve (approved):

PR #269, opened 2026-07-28
perf(reward/editreward): batch EditReward scoring in one forward

celve:
## Summary

The EditReward scorer's `score()` looped over items and ran the 7B VLM **once per
item** (a batch of 1), even though `EditRewardInferencer.reward()` already accepts
parallel lists and runs a single padded batch. On a real rollout (hundreds of
image pairs) that meant hundreds of sequential single-item forwards.

This rewrites `score()` to collect all valid items and issue **one batched
`reward()` call**, splitting the `(N, num_heads)` logits back per item via a new
`_shape_reward` helper that mirrors `_score_single`'s output shaping exactly. The
underlying model was trained with batching (`per_device_train_batch_size=2`), so
batched inference returns the same per-item scores — this changes throughput, not
values. A batch-level exception falls back to the per-item path, so one bad image
cannot NaN the whole batch.

## Related Issue

N/A

## Test Plan

Exercised end-to-end in a FLUX.2-Klein-4B image-edit GRPO run scored by this
service (`TIGER-Lab/EditReward-MiMo-VL-7B-SFT-2508`, 1 dedicated H20 for the
scorer), 60 rollouts × 2 arms:

- **Before:** ~2 s per image pair; a rollout's worth of scoring (~900 pairs)
  serialized into hundreds of forwards, tripping the client's 300 s timeout and
  crashing the run on the retry path.
- **After:** ~0.5 s per image pair effective; both 60-rollout runs completed
  cleanly (`TRAIN_EXITED rc=0`) with zero reward-service timeouts.
- Reward values are unchanged vs the per-item path (batched transformer inference
  with the processor's attention mask is numerically equivalent, and the model was
  trained batched).

Static: `python -m compileall unirl-reward-service/reward_service/scorers/editreward.py`
passes.

## Compatibility / Risk

- **No behavior change to reward values** — batched vs per-item scoring produces
  the same per-item outputs. Purely a throughput change.
- Robustness preserved: a batch-level failure falls back to per-item scoring, so a
  single malformed item degrades to one NaN instead of failing the batch.
- No API, config, checkpoint, or data-format change.

## Reviewer Notes

- **Duplicate-work check.** No open PR touches `scorers/editreward.py`; this is a
  self-contained perf fix in the reward service, independent of the ti2i rollout
  work in #242.
- **AI assistance.** Developed with AI assistance; every line was reviewed and the
  end-to-end validation above was run and inspected.

PR #244, opened 2026-07-23
feat(diffusion): add FlashGRPO algorithm and SGLang rollout

NancyFyong:
## Summary

Adds an opt-in Flash-GRPO path for diffusion/video RL and wires it into SGLangDiffusion rollout:

- adds `FlashSDEStrategy`, matching the upstream Flash-GRPO transition coefficient `std_dev_t = sigma_min + (sigma_max - sigma_min) * sigma`
- adds `FlashGRPO`, reusing FlowGRPO replay/anchor/KL plumbing and applying Flash-GRPO temporal gradient rectification
- adds a WAN2.1 T2V trainside Flash-GRPO recipe using PickScore prompts and `video_pickscore` reward
- extends SGLangDiffusion SDE dispatch so `FlashSDEStrategy` maps to native `rollout_sde_type="flash"`
- extends the existing SGLang SDE patch with a Flash transition kernel and adds a WAN2.1 T2V Flash-GRPO SGLang recipe

The Flash-GRPO recipes intentionally mirror the existing WAN baselines for non-Flash hyperparameters: `wan21_t2v_flashgrpo.yaml` follows `wan21_t2v.yaml`, and `wan21_t2v_flashgrpo_sglang.yaml` follows `wan21_t2v_sglang.yaml`. FlashGRPO-specific mappings are: Flash SDE strategy, FlashGRPO loss, `clip_range=1e-3`, `eta=1.0`, one SDE step from the first half of the configured schedule, matching temporal-rectification candidate indices, and the upstream effective prompt group size.

The upstream WAN script uses `WanPipeline` with text prompts, so this is a T2V integration. Its `num_image_per_prompt=4` drives a distributed repeated-prompt sampler, while the WAN generation call hardcodes `num_videos_per_prompt=2`; UniRL `samples_per_prompt` is the actual per-prompt rollout fan-out/group size, so the effective mapping is `4 * 2 = 8`.

The implementation is based on `Shredded-Pork/Flash-GRPO` commit `bd6051f68e1ab444e5ec7c6ffe0a1f7eaf559a0d`.

## Related Issue

N/A

## Test Plan

Static/config validation:

- `python -m ruff check unirl/sde/kernels.py unirl/algorithms/flashgrpo.py unirl/algorithms/__init__.py`
- `python -m py_compile unirl/sde/kernels.py unirl/algorithms/flashgrpo.py unirl/algorithms/__init__.py`
[...]

PR #270, opened 2026-07-28
feat(reward-service): add videohpsv3 scorer for T2V rollouts

NancyFyong:
## Summary

Adds a `videohpsv3` reward scorer to `unirl-reward-service`, so WAN2.1 T2V
rollouts can be scored with HPSv3 the way Flash-GRPO does: decode the clip,
score every frame against the prompt, and average the best `top_ratio` fraction
of the per-frame scores into one reward.

Flash-GRPO splits this across the wire — its client extracts frames,
JPEG-compresses them, and POSTs each video's frames to a dedicated HPSv3 server
that only ever sees "a batch of images". Here both halves live in one scorer,
matching this repo's convention that a reward name maps to a single
self-contained scorer. `HPSv3Scorer` (added in #142/#143) is reused by
composition, so it keeps owning checkpoint resolution, `max_batch_size` chunking
and the inter-chunk `empty_cache()`; this module adds only decode and aggregation.

Three commits:

1. `feat(reward-service): accept an HF repo id in hpsv3 weights_path`
2. `feat(reward-service): add videohpsv3 scorer for T2V rollouts`
3. `build(reward-service): pin trl, tensorboard and a FlashAttention-2 wheel for hpsv3`

Ordered so the repo-id support lands before the config that depends on it, which
keeps the range bisect-safe.

Points worth a reviewer's attention:

- **`frame_stride` defaults to 1, deliberately.** Raising it changes the
  estimator rather than making it cheaper. A 64-clip step of 81-frame videos
  scores 5184 frames at stride 1 but only 704 at stride 8, and `top_ratio_mean`
  then keeps 3 frames instead of 24 — so a policy can maximise reward by making
  three frames pretty and letting the other 78 rot. We saw exactly that failure
  mode on a long stride-8 run. The cost of stride 1 is affordable: the reward
  call measured **8.2s of a 1029s training step (0.8%)**, so generation dominates.
- **`weights_path` now accepts an HF repo id.** It previously handled only a
  directory holding `HPSv3.safetensors` or the file itself, and passed anything
  else through as a file path — so a repo id reached `torch.load("MizzenAI/HPSv3")`
  and raised, and every deployment had to pre-download 16.5 GB and hardcode a
  local path. Resolution moved into a `_resolve_checkpoint_path` staticmethod
  (mirroring `HPSv2Scorer`) that falls back to `hf_hub_download`, which is the
  same interpretation `hpsv3/inference.py` gives its own default. Hub errors are
  re-raised as `FileNotFoundError` naming both interpretations, because a typo'd
  path otherwise surfaces as an `HFValidationError` about repo id syntax.
- **Per-item failures return NaN rather than raising**, so one undecodable clip
  does not drop a whole 64-item bucket. Missing or mistyped video input still
  raises, because that is a wiring bug and must not be masked.
- **`cv2` is imported lazily**: `registry._try_import` swallows `ImportError` into
  a warning, so a top-level import would silently leave the reward unregistered.
- `extract_frames` and the shared HPSv3 batching helper move into
  `scorers/_common.py`, so `videoalign` (#97) and `video_hpsv3` stop keeping
  separate frame-decode paths. `videoalign` behaviour is unchanged.

## Related Issue

N/A

## Test Plan

Branch is rebased onto `upstream/main` at 5d034c3; the range is `upstream/main..HEAD`.

```
SKIP=no-commit-to-branch pre-commit run --from-ref upstream/main --to-ref HEAD
    # recipe _target_ paths resolve ... Passed (rest skipped: nested project)

ruff check reward_service/scorers/{video_hpsv3,_common,registry,videoalign,hpsv3_scorer}.py
    # All checks passed!

python -c "from reward_service.config import load_config; \
           print(load_config('configs/videohpsv3_service.yaml').rewards[0].params)"
    # {'weights_path': 'MizzenAI/HPSv3', 'config_path': None,
    #  'max_batch_size': 4, 'frame_stride': 1, 'top_ratio': 0.3}
```

`ruff check .` over the whole package reports 33 errors, all pre-existing on
`main` and untouched here; every file this PR adds or modifies is clean.

**Unit tests exist but are not included in this diff.** 50 cases across
`tests/scorers/test_video_hpsv3.py` (frame sampling, top-ratio aggregation, NaN
paths), `test_common.py` (the shared decode/batching helper) and
`test_hpsv3_scorer.py` (`_resolve_checkpoint_path`: directory, file, repo id, bad
value). I ran them against *this exact tree* — dropped into `unirl-reward-service/tests/`
on this branch, `pytest tests/ -q` → **50 passed** — then removed them, since
`unirl-reward-service` carries no test tree upstream today and CI runs no pytest.
Say the word and I will push them onto this branch.

Real-hardware validation: an 8-GPU WAN2.1-1.3B T2V Flash-GRPO job driving this
service at `frame_stride: 1`, 8 replicas, FlashAttention-2 confirmed active on
the HPSv3 backbone, currently past rollout 8/1000 with no scorer errors. Frame
accounting was verified empirically — stride 1 decodes all 81 frames and
`top_ratio_mean` keeps 24.

## Compatibility / Risk

- **Additive for scorers.** No existing reward's score changes.
- **`hpsv3_scorer.py` changes one existing behaviour**: a `weights_path` that is
  neither a directory nor an existing file used to be handed to the inferencer as
  a file path (and fail there); it is now interpreted as a repo id and downloaded.
  Valid local paths behave exactly as before.
- **The new config downloads ~16.5 GB** from `MizzenAI/HPSv3` on first actor init.
  Nodes without egress should point `weights_path` at a local directory.
- **The FlashAttention-2 wheel URL is a hard platform pin**: `cp312` / `torch2.12` /
  `cu130`, Hopper `sm_90`. It cannot be a normal version spec — PyPI's `flash-attn`
  builds from source, and HPSv3's `create_model_and_processor` picks between FA2
  and SDPA purely on whether `import flash_attn` succeeds, so the presence of the
  package *is* the switch. FA3 does not substitute: its import name is
  `flash_attn_interface` and HPSv3 hardcodes the FA2 string. Reviewers on other
  platforms will need a different wheel. **This is the line I would most like a
  second opinion on** — it is correct for our nodes and fragile everywhere else.
- `trl` and `tensorboard` are new entries in `envs/hpsv3.txt` and are not
  optional: `hpsv3/__init__` → `inference.py:9` → `hpsv3/train.py:21` imports
  `trl` at module level, and `train.py:8` pulls in `qwen2vl_trainer`
  (`trl.RewardTrainer`, `torch.utils.tensorboard.SummaryWriter`). So they are
  import-time requirements even for inference-only scoring. Verified against the
  installed package: `trl 0.21.0 / tensorboard 2.21.0 / transformers 4.57.6`.
- No checkpoint, data-format or public API changes.

## Reviewer Notes

- **Draft on purpose.** The consumer recipe is not here: it targets
  `unirl.algorithms.flashgrpo.FlashGRPO`, which lands in #244, and the
  `recipe _target_ paths resolve` hook rejects it until that merges. This PR is
  self-contained and reviewable now (the scorer has zero `unirl` imports), but I
  would rather land it after or alongside #244 so the recipe can ship with it.
- `configs/videohpsv3_service.yaml` sets `num_replicas: 8` with a comment
  explaining that those GPUs are the same physical devices the trainer holds —
  deliberate over-subscription that fits HPSv3's ~17 GB of bf16 weights alongside
  a ~48 GB training footprint on a 97 GB H20. That is our deployment shape, not a
  universal default; happy to ship `1` if you prefer the checked-in config
  conservative.
- `score_timeout_s: 1800` looks enormous for one request, and it is not one
  request: `RewardService.score_and_attach` is DP_SCATTER-dispatched, every
  training rank POSTs its own shard, and `_await_ref` starts its clock at
  dispatch — so the deadline has to cover the whole rollout step.
- **Duplicate-work check**: no open PR touches `unirl-reward-service` video
  rewards. #269 (EditReward batching) is the only recent change in that package
  and touches different files; #142/#143/#146 added the image HPSv3 scorer and
  #97 added VideoAlign, both merged and both built on here rather than duplicated.
- **AI assistance**: authored with Claude Code. The full diff was reviewed and
  every command in the Test Plan was run locally before publication.
- Follow-up, not done here: `reward_service.config` does not `expandvars` on
  params, so a `weights_path` like `$HF_HOME/...` will not expand. The repo-id
  support removes most of the need, so I left it alone.

## Checklist

- [x] I updated docs (`docs/ARCHITECTURE.md`) and configs. Tests exist and pass
      (50) but are deliberately not in this diff — see Test Plan.

### DIFF
--- a/unirl-reward-service/configs/service.example.yaml
+++ b/unirl-reward-service/configs/service.example.yaml
@@ -103,6 +103,28 @@ rewards:
       # KV footprint. Raise only after confirming GPU headroom.
       max_batch_size: 4
 
+  # Video HPSv3: scores every frame with HPSv3 and averages the top 30%,
+  # reproducing Flash-GRPO's videohpsv3 reward for WAN2.1 T2V. Needs a video
+  # (video_b64 / video_path) on the request; disabled by default like the other
+  # T2V rewards. Shares envs/hpsv3.txt with the image scorer above, so Ray
+  # reuses one cached venv. Costs one Qwen2-VL-7B forward per frame (~5.2k per
+  # 64-clip rollout at frame_stride 1), so raise server.score_timeout_s well
+  # above its 120 s default before enabling — on timeout the Ray task keeps
+  # running and its GPU time is discarded.
+  # - name: videohpsv3
+  #   scorer: videohpsv3
+  #   runtime_env: envs/hpsv3.txt
+  #   num_replicas: 1
+  #   num_gpus: 1
+  #   num_cpus: 4
+  #   max_concurrency: 1
+  #   params:
+  #     weights_path: /path/to/RewardModel/HPSv3
+  #     config_path: null
+  #     max_batch_size: 4       # frames per forward; 8 OOMs a 95GB H20
+  #     frame_stride: 1         # 1 = exact upstream semantics; 8 = 1/8 the cost
+  #     top_ratio: 0.3          # upstream Flash-GRPO value
+
   # ─────────────── vLLM-based scorers ───────────────
   #
   # All fields below (dtype / enforce_eager / swap_space / quantization /
--- a/unirl-reward-service/configs/videohpsv3_service.yaml
+++ b/unirl-reward-service/configs/videohpsv3_service.yaml
@@ -1,0 +1,41 @@
+# VideoHPSv3-only service configuration (WAN2.1 T2V / Flash-GRPO).
+#
+# Start:
+#   ray start --head --port=6379 --num-gpus=1
+#   python -m reward_service --config configs/videohpsv3_service.yaml
+# The training node connects via REWARD_SERVICE_URL=http://<this_node_ip>:8080.
+#
+# Consumer recipe: examples/diffusion/wan21/wan21_t2v_flashgrpo_sglang_videohpsv3.yaml
+
+server:
+  host: 0.0.0.0
+  port: 8080
+  # Covers a whole rollout step, not one request: score_and_attach is
+  # DP_SCATTER-dispatched and _await_ref clocks from dispatch, so queue wait
+  # counts too. On timeout the Ray task keeps running and its GPU time is lost.
+  score_timeout_s: 1800.0
+
+rewards:
+  - name: videohpsv3
+    scorer: videohpsv3
+    # Same deps as the image hpsv3 scorer, so Ray reuses one cached venv.
+    runtime_env: envs/hpsv3.txt
+    # One replica per GPU. This service runs its own Ray cluster, so these are
+    # the same physical GPUs the trainer holds — HPSv3's ~17 GB of bf16 weights
+    # fit alongside the ~48 GB training footprint on a 97 GB H20.
+    num_replicas: 8
+    num_gpus: 1
+    num_cpus: 4
+    max_concurrency: 1
+    params:
+      # HF repo id, downloaded once into the hub cache. A local directory
+      # holding HPSv3.safetensors also works, for nodes without egress.
+      weights_path: MizzenAI/HPSv3
+      config_path: null
+      max_batch_size: 4       # frames per forward; 8 OOMs a 95GB H20
+      # 1 = upstream semantics: all 81 frames scored, best 24 averaged. Raising
+      # it is a reward-hacking hole, not an optimization — at stride 8 only 3
+      # frames survive top_ratio and 78 go unconstrained. Measured cost of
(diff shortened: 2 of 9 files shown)


haonan3:
Before merging, please run an actual training job (for this reward, you need find proper training task & data) to validate the change. After integrating the VideoHPSv3 reward, confirm that the reward curve behaves normally. A screenshot of the curve or the corresponding training logs is sufficient.

Please convert the PR to draft for now and mark it as ready for review again once the validation results are available.

PR #259, opened 2026-07-26
feat(algorithms): add AR PPO with GAE and value loss

yhl48:
## Summary

Part 3/3 of #86: add PPO with token-level GAE and a clipped value objective for AR training.

**Stacked on #256. Merge #256 first, then rebase this branch onto `main` so the PR reduces to the single PPO/trainer commit.**

- Adds `PPO` / `PPOConfig` with clipped policy loss and clipped critic loss.
- Freezes old critic values with the same planned micro-batch geometry used by the training forward, then computes GAE only after those anchors are reassembled.
- Applies `loss_mask` consistently to both actor and critic optimization and metrics.
- Adds `advantage_mode: grpo | gae` to both synchronous and asynchronous AR trainers.
- Adds a current-schema Qwen3-4B-Base + DAPO + colocated SGLang PPO recipe.
- Excludes train-only `value_head.*` tensors from SGLang weight sync.

## Related Issue

Part of #86. Depends on #256.

## Test Plan

No PR-specific test files are included.

- `SKIP=no-commit-to-branch pre-commit run --all-files --show-diff-on-failure` — all hooks passed.
- `python -m compileall -q unirl` — passed.
- Hydra composition of `ar/qwen3_ppo_4b_base_dapo_sglang` with placeholder dataset/model paths — passed.
- Temporary, uncommitted correctness harness — passed:
  - exact multi-trajectory GAE values and terminal-reward placement;
  - masked tokens have zero actor and critic gradient contribution;
  - critic anchors follow planned micro geometry (`3 + 2` tokens), rather than a mismatched full-batch forward;
  - synchronous and asynchronous `advantage_mode` wiring;
  - meta-init + sharded-load Qwen3 replay followed by the actual `PPO.prepare_segment -> prepare_part -> compute_loss_and_backward` path.
- GPU smoke on an H20 with a 9.8M-parameter tiny Qwen3 checkpoint — passed:
  - eager/meta checkpoint state maximum difference: `0.0`;
  - first-update `ratio_mean`: `1.0`;
  - finite combined PPO loss;
  - finite, nonzero value-head gradient (`L2=2.7303159`).
- Full Qwen3-4B-Base comparison on 8 x H20, BF16, SGLang `0.5.12.post1` — passed:
  - 8 SGLang engines loaded the real checkpoint and completed full-tensor sync on every rollout;
[...]

PR #256, opened 2026-07-25
feat(models): add AR value head and GAE wiring

yhl48:
## Summary

Part 2/3 of #86: add the critic/value and GAE data path required by AR PPO. This branch has been rebuilt on the current `main`.

- Adds an optional scalar FP32 `ValueHead` for Qwen3.
- Keeps value-head initialization deterministically zero for both eager construction and meta-init + sharded checkpoint loading.
- Lets packed and padded Qwen3 replay return aligned per-token values through `ReplayResult.values`.
- Adds packed `values`, `returns`, and `token_advantages` fields to `TextSegment`.
- Adds terminal-reward scattering and per-trajectory GAE without cross-trajectory leakage.
- Preserves packed-segment metadata during GAE preparation and treats `loss_mask` as an optimization mask, not as an artificial trajectory boundary.

The PPO optimizer and trainer wiring remain in #259.

## Related Issue

Part of #86. #259 is stacked on this PR and should be reviewed/merged after this one.

## Test Plan

No PR-specific test files are included.

- `SKIP=no-commit-to-branch pre-commit run --all-files --show-diff-on-failure` — all hooks passed.
- `python -m compileall -q unirl` — passed.
- Temporary, uncommitted correctness harness — passed:
  - eager and meta-init value heads start at exactly zero;
  - meta-init + real sharded checkpoint load matches eager load (`max_state_diff=0.0`);
  - packed replay values and log-probabilities stay token-aligned;
  - multi-trajectory GAE does not leak across sequence boundaries;
  - reward propagation continues across masked tokens while masked tokens remain excluded from optimization;
  - value-head backward gradient is finite and nonzero.
- GPU smoke on an H20 with a 9.8M-parameter tiny Qwen3 checkpoint — passed.

## Compatibility / Risk

- `use_value_head` defaults to `false`; existing Qwen3 GRPO behavior is unchanged.
- `return_values=false` preserves the legacy replay return type.
- Base Hugging Face checkpoints do not contain critic weights; the new head is initialized from scratch at zero.
- The value head is train-side only and must be excluded from rollout-engine weight sync; #259's recipe does this.

## Reviewer Notes

Please review packed/padded replay alignment and the meta-init post-load zeroing first. AI assistance was used; the final diff was reviewed and validated against the current repository state.

## Checklist

- [x] I reviewed the changed code and removed unrelated/generated artifacts.
- [x] I validated the change without adding PR-specific test files.

PR #282, opened 2026-07-30
refactor(trainer): extract the sync-loop template into BaseTrainer

haonan3:
## Summary

Extracts the synchronous training loop into a `BaseTrainer` template method. Nine trainers hand-copy the same outer skeleton today (~700–900 duplicated lines repo-wide, measured); the diffusion and AR `train()` bodies differ in exactly four lines. `BaseTrainer.train()` now owns the loop — load checkpoint → data fast-forward → wandb → baseline eval → per-rollout build / sync-gate / `train_step` / log / periodic eval / save → `finally` finish-wandb + `on_train_end` — with four hooks for the observed divergences: `_wandb_extra`, `evaluate_baseline`, `evaluate_periodic`, `on_train_end` (AR routes `_shutdown_runtime` through it, preserving its nested-`finally` ordering).

`DiffusionTrainer` and `ARTrainer` drop their `train()` bodies (line-equivalent to the template, including the resume force-sync semantics and per-module progress-logger names). `PETrainer` / `UnifiedModelTrainer` / `SFTTrainer` keep their own loops and migrate in follow-ups; async/agentic loops are intentionally out of scope (producer/consumer shapes, not this template). Independent of #210 (based on main).

## Test Plan

Static (CPU): `DiffusionTrainer.train is BaseTrainer.train` and `ARTrainer.train is BaseTrainer.train` assert-verified; `ARTrainer.on_train_end` override confirmed; compose-check `diffusion/sd3/sd3_trainside` and `ar/qwen3_grpo_4b_base_dapo_sglang` rc=0; `pre-commit run --all-files` green.

Pending GPU parity before merge:
- [ ] diffusion smoke (sd3_trainside, ~10 rollouts) — identical logs/cadence vs main, incl. a resume-from-checkpoint step (force-sync path);
- [ ] AR smoke (qwen3 grpo, ~5 rollouts) with `eval_interval>0` — baseline + periodic eval fire at the same steps; teardown runs on both clean exit and Ctrl-C.

## Compatibility / Risk

No config or entrypoint changes; `train(**kwargs)` signature is unchanged for both migrated trainers. Behavioral risk is confined to the four hook seams — hence the explicit parity checklist above. Not-yet-migrated trainers are untouched.

AI-assisted; reviewed and directed by the maintainer.


haonan3:
Closing by triage, not by verdict on the idea. The duplication is real and has **grown** since this was cut — ar / diffusion / pe / unified_model (and experimental refl) all hand-roll the same load→wandb→loop→save→finish skeleton. But the trainer surface is in an active evolution burst (AR PPO/GAE #259, value head #256, hi3 it2i #186), a template landed now adds hook-friction to every in-flight change, and this draft only migrated 2 of the 5 candidates with no GPU parity smoke.

**Re-derive when the AR-PPO wave settles**, scoped to all five sync trainers in one pass — the branch (`refactor/base-trainer-sync-template`) stays as reference.

PR #288, opened 2026-07-31
fix(qwen_image): drop txt_seq_lens for diffusers>=0.39

CjhHa1:
## Summary

Qwen-Image is broken on a clean install. diffusers 0.39.0 removed the
`txt_seq_lens` argument from `QwenImageTransformer2DModel.forward`, and
`pyproject.toml` floors diffusers at `>=0.38.0` with no upper bound — so a fresh
environment resolves to 0.39.0 and every Qwen-Image forward dies with:

```
TypeError: QwenImageTransformer2DModel.forward() got an unexpected keyword argument 'txt_seq_lens'
```

This hits `qwen_image` and `qwen_image_edit_plus`, on both the rollout and the
training path, before any model-specific logic runs.

The fix is to stop passing the argument at the four call sites. It is already a
no-op on the versions we support: 0.37 and 0.38 accept it, emit a deprecation
warning, and then ignore it — the length used for the RoPE text slice comes from
`compute_text_seq_len_from_mask(encoder_hidden_states, encoder_hidden_states_mask)`,
and that is what reaches `pos_embed(..., max_txt_seq_len=...)`. 0.39 removed the
parameter and takes the same mask-derived path. So deleting it changes nothing on
0.37/0.38 and unbreaks 0.39.

No version bound is added: `sglang[diffusion]` hard-pins `diffusers==0.37.0`
(already reconciled at the floor in #91), so an upper pin would break the sglang
path — and it is not needed.

The per-call trim to the batch's true max text length stays, but its comment was
wrong and is corrected. It claimed to guard a width mismatch in
`apply_rotary_emb_qwen`; that was real for the `txt_seq_lens`-honouring path in
<=0.36 (LIN-382 probe-f), but from 0.37 on the RoPE text slice is taken at
`encoder_hidden_states.shape[1]` — the trimmed tensor's own width — so a mismatch
is no longer possible. The trim now only avoids work on all-pad columns, and the
comment says that instead.

## Related Issue

N/A — found while validating #156 on a fresh environment. #156 has since landed
and rewrote both `predict_noise` bodies, so this branch was merged onto current
`main` and the fix reapplied on top of the new code. The merge commit takes
upstream verbatim for the two files; the fix itself is the commit after it.

## Test Plan

Source-level, against the diffusers sources for v0.37.0, v0.38.0 and v0.39.0:
confirmed that on 0.37/0.38 `txt_seq_lens` never reaches the RoPE builder — the
only thing `forward` does with it is `deprecate(...)`, after which `text_seq_len`
is derived from the mask — and that 0.39 no longer declares the parameter at all.
This is what makes the deletion a no-op rather than a behaviour change.

Repo checks: the pre-push `pre-commit` suite passes — ruff check, ruff format,
recipe `_target_` resolution, experimental-tier boundaries, plus the file
hygiene hooks.

GPU, on the current base: 1x H20, real `Qwen-Image-Edit` transformer weights,
torch 2.10.0+cu128, with diffusers 0.37.0 vs 0.39.0 as overlay venvs over one
conda env so diffusers is the only variable. Driven through #156's replay
harness (`--mode correctness --model qwen_image`, B=2, S=3, 384px), which
exercises `diffuse` + batched/serial `replay` with variable-length text masks,
and with `--cfg` the CFG negative call site too.

| code | diffusers | result |
|---|---|---|
| `upstream/main` | 0.37.0 | passes |
| `upstream/main` | 0.39.0 | **`TypeError: ... unexpected keyword argument 'txt_seq_lens'`** |
| this branch | 0.37.0 | passes |
| this branch | 0.39.0 | passes |

Across the three passing arms every digest matches to the printed precision, in
both the no-CFG and CFG variants — mean \|log-prob\| (`1.01866674` / `1.01866949`),
batched-vs-serial parity, the ratio anchor, and `grad_norm` (`0.46177297`), so
backward is covered as well as forward. Full tables in the PR comments. The
failing arm dies at the `model.transformer(...)` call in `predict_noise`, which
is the signature this PR changes.

All four changed call sites are covered: `qwen_image_edit_plus` was run the same
way, where the pre-fix code likewise dies on 0.39 at its own `predict_noise` and
both fixed arms pass.

The strongest evidence for "no number moved" is a same-process check rather than
the tables above, because a cross-process digest cannot resolve it on Edit-Plus
(repeating one identical arm moves `logp_mean` by 4.2e-3, four times the
base-vs-fix gap). Wrapping the transformer's forward, every call the real code
makes is run twice on identical inputs — once without `txt_seq_lens`, once with it
rebuilt from the mask as `main` did — and compared bitwise. Over both stages and
both CFG settings on diffusers 0.37, **48 calls, 0 differences, max abs diff
exactly 0**, with variable-length text (`txt_seq_lens=[64, 56]`) so the trim is
live. Since that keyword is the only difference between the two code versions,
they cannot disagree.

Not covered: the full training loop (VAE, text encoder, reward model, FSDP). The
argument only ever reached the transformer call, so the above brackets the change;
happy to add a `qwen_image_trainside` rollout under 0.39 if a reviewer wants it.

## Compatibility / Risk

Low, and lower than the previous revision of this PR — this is now a pure
deletion with no added indirection. No config, checkpoint, or data-format
changes; no public API change. On 0.37/0.38 the only observable difference is
that diffusers stops emitting its `txt_seq_lens` deprecation warning. On 0.39 the
call works at all, where before it raised.

## Reviewer Notes

- Simplified per @haonan3's review: the signature probe
  (`_transformer_takes_txt_seq_lens` / `_text_len_kwargs`, plus the `inspect` and
  `lru_cache` imports it needed) is gone. What remains is the deletion.
- The per-sample length tensors (`true_lens` / `neg_true`) existed only to feed
  `.tolist()`, so each site collapses to the single `max()` it actually needs.
- Four call sites across two files (`qwen_image` and `qwen_image_edit_plus`, each
  with its CFG negative branch).
- No test file added: the repo has no test suite to extend.
- Worth deciding separately whether other model families pass removed or renamed
  diffusers kwargs. I only audited the Qwen-Image ones, which are the ones with an
  unbounded floor and a removed argument.
- AI-assisted.

## Checklist


haonan3:
The bug is real and the fix works, but I think it can be much simpler.

I checked the actual diffusers source: since 0.37, `txt_seq_lens` is deprecated and ignored — the model always derives the text length from the mask, exactly like 0.39. The `max(None)` crash mentioned in the description only existed in diffusers ≤0.36, which we don't support. That's also why your test digests match bit-for-bit: on 0.37 the argument was already doing nothing.

So we can simply delete `txt_seq_lens` at the 4 call sites — no signature probe, and no version change (sglang still pins diffusers 0.37, and deletion works fine there too). Could you simplify to that, fix the related comments, and rerun your harness once to confirm the digests are unchanged?

CjhHa1:
You're right — simplified to a plain deletion, thanks.

I checked your reading against the sources for v0.37.0, v0.38.0 and v0.39.0 and it
holds: on 0.37/0.38 `forward` accepts `txt_seq_lens`, calls `deprecate(...)`, and
then never uses it — `text_seq_len` comes from
`compute_text_seq_len_from_mask(...)` and that is what reaches
`pos_embed(..., max_txt_seq_len=...)`. 0.39 simply drops the parameter and takes
the same path. So the `max(None)` claim in my description was wrong: that was
<=0.36, which we don't support. Description updated.

What's in now:

- `txt_seq_lens` deleted at the 4 call sites — no signature probe, no
  `inspect`/`lru_cache`, no version bound.
- `true_lens` / `neg_true` existed only to feed `.tolist()`, so each site
  collapses to the single `max()` it needs.
- Fixed the stale comment on the trim. It claimed to guard the
  `apply_rotary_emb_qwen` width mismatch, but from 0.37 the RoPE text slice is
  taken at the trimmed tensor's own width, so that can't happen — the trim now
  just avoids work on all-pad columns.

One thing to flag: #156 landed in the meantime and rewrote both `predict_noise`
bodies, so I merged current `main` and reapplied the fix on top of the new code.
The merge commit takes upstream verbatim for the two files, and the fix is the
commit after it, so it stays reviewable on its own.

The harness rerun is still outstanding. The old digests were taken on the
probe revision and the pre-#156 code path, so quoting them here would be
misleading — I'll rerun on the new base and post the numbers.

CjhHa1:
Reran the harness on the current base — digests are unchanged, and the failure it
fixes reproduces. Numbers below.

Setup: 1x H20, real `Qwen-Image-Edit` transformer weights (39 GB, 9 shards),
torch 2.10.0+cu128, transformers 4.57.6. The two diffusers versions are overlay
venvs over one conda env, so **diffusers is the only variable** between arms —
same torch, same weights, same seed. Harness is #156's replay harness
(`--mode correctness --model qwen_image`, B=2, S=3, 384px), which drives
`diffuse` + batched/serial `replay` through `QwenImageDiffusionStage` with
**variable-length text masks** (so the per-call trim is exercised), and with
`--cfg` also drives the CFG negative call site.

Four arms, where "base" is `upstream/main` (still passes `txt_seq_lens`) and
"fix" is this branch:

| code | diffusers | result |
|---|---|---|
| base | 0.37.0 | passes |
| base | 0.39.0 | **`TypeError: QwenImageTransformer2DModel.forward() got an unexpected keyword argument 'txt_seq_lens'`** |
| fix | 0.37.0 | passes |
| fix | 0.39.0 | passes |

The failing arm dies exactly where the report said, at the `model.transformer(...)`
call in `predict_noise`, before any of the harness's claims are evaluated.

Digests, no CFG:

| metric | base+0.37 | fix+0.37 | fix+0.39 |
|---|---|---|---|
| `logp_scale` (mean \|log-prob\|) | 1.01866674 | 1.01866674 | 1.01866674 |
| `parity_max_abs` | 2.38418579e-07 | 2.38418579e-07 | 2.38418579e-07 |
| `parity_max_rel` | 2.13006629e-07 | 2.13006629e-07 | 2.13006629e-07 |
| `parity_mean_rel` | 1.60053929e-07 | 1.60053929e-07 | 1.60053929e-07 |
| `ratio_dev` | 0 | 0 | 0 |
| `cross_step_ratio` | 597287.974 | 597287.974 | 597287.974 |
| `grad_norm` | 0.46177297 | 0.46177297 | 0.46177297 |
| `ratio_dev_vs_rollout_batched` | 2.38418579e-07 | 2.38418579e-07 | 2.38418579e-07 |

With CFG (`guidance_scale > 1`, second call site):

| metric | base+0.37 | fix+0.37 | fix+0.39 |
|---|---|---|---|
| `logp_scale` | 1.01866949 | 1.01866949 | 1.01866949 |
| `parity_max_abs` | 5.96046448e-07 | 5.96046448e-07 | 5.96046448e-07 |
| `parity_max_rel` | 5.62031516e-07 | 5.62031516e-07 | 5.62031516e-07 |
| `parity_mean_rel` | 1.69982371e-07 | 1.69982371e-07 | 1.69982371e-07 |
| `ratio_dev` | 0 | 0 | 0 |
| `cross_step_ratio` | 562067.111 | 562067.111 | 562067.111 |
| `ratio_dev_vs_rollout_batched` | 6.55651093e-07 | 6.55651093e-07 | 6.55651093e-07 |

Every value matches across all three passing arms to the printed precision,
including `grad_norm` (a backward-pass digest, so this is not just a forward
comparison). #156's three claims hold identically in every arm
(batched==serial, ratio anchor exactly 1, gradient flows).

Two caveats, so the table isn't read as more than it is:

- In the CFG variant the harness skips its gradient claim with OOM at this batch
  size — in all three arms alike, so the comparison stays apples-to-apples, but
  there is no `grad_norm` for that row.
- This exercises the diffusion stage on real weights, which is what the diff
  touches, but it is not the full training loop: no VAE, text encoder, reward
  model, or FSDP. If you'd rather see a `qwen_image_trainside` rollout under
  0.39 before merging, say so and I'll run it — the reason I stopped here is
  that the argument only ever reached the transformer call.

CjhHa1:
Follow-up: the other two call sites (`qwen_image_edit_plus`) are now covered too,
and the no-op claim is verified more directly than the earlier digest table
managed. Same pod, same staged weights, same two overlay venvs.

**The bug reproduces at the Edit-Plus call sites, and the fix clears it.**
Pre-fix Edit-Plus code on diffusers 0.39 dies at its own `predict_noise`:

```
File "unirl/models/qwen_image_edit_plus/diffusion.py", line 125, in predict_noise
    noise_pred_packed = model.transformer(
TypeError: QwenImageTransformer2DModel.forward() got an unexpected keyword argument 'txt_seq_lens'
```

with and without CFG; both fixed arms pass. So all four changed call sites are
now empirically covered, not just the two in the base stage.

**On the "does it change any number" question, I have to correct how I'd framed
the evidence.** A cross-process digest can't answer it for Edit-Plus. Repeating
one *identical* arm three times moves `logp_mean` by 4.2e-3, while the
base-vs-fix gap was 1.0e-3 — the spread within an arm is four times the candidate
signal, so a `DIFFERS` verdict there means nothing. (Within a single process it's
exact: two replays agree to 0. The variation appears across processes, and only
for Edit-Plus, whose source-image concat doubles the token count — 1152 vs 576 —
which is consistent with a different, atomics-accumulating GEMM being selected
per process. The base Qwen-Image shapes were bit-stable across processes, which
is why that table came out identical.)

So I tested the claim inside one process instead, which removes the noise
entirely. Wrapping the transformer's forward, every call the real code makes is
run **twice on the very same inputs** — once without `txt_seq_lens` (this PR) and
once with it, rebuilt from the mask exactly as `main` did — and the outputs are
compared bitwise. Same weights, same process, same kernels; the keyword is the
only thing that varies. And since its presence is the *only* difference between
the two code versions, bitwise equality here means the versions cannot disagree.

diffusers 0.37.0, real `Qwen-Image-Edit` weights, variable-length text
(`txt_seq_lens=[64, 56]`, so the trim is live):

| stage | CFG | calls compared | calls that differed | max abs diff |
|---|---|---|---|---|
| `qwen_image` | off | 8 | 0 | 0 |
| `qwen_image` | on | 16 | 0 | 0 |
| `qwen_image_edit_plus` | off | 8 | 0 | 0 |
| `qwen_image_edit_plus` | on | 16 | 0 | 0 |

48 calls, zero differences, max abs diff exactly `0` — not "small", identically
zero. That is the source-level reading (0.37/0.38 `deprecate(...)` the argument
and then derive the length from the mask regardless) confirmed on real weights at
tensor level.

---

Unrelated finding while wiring this up, flagging it rather than folding it in:
**batched-step replay is broken for Edit-Plus on `main`.**
`QwenImageDiffusionStage._tile_conditions` hardcodes
`QwenImageConditions(text=..., negative_text=...)`, and
`QwenImageEditPlusDiffusionStage` doesn't override it, so the tiled conditions
lose `image_latent` and `predict_noise` raises
`AttributeError: 'QwenImageConditions' object has no attribute 'image_latent'`.
Reproduces on `upstream/main` (998f2f32) independently of this PR, so it predates
it — #156 extended the batched path to `qwen_image` and the Edit-Plus subclass
inherited it without tiling support. It's latent rather than live:
`batch_replay_steps` defaults to False and no Edit-Plus recipe sets it (only
`examples/pe/pe_trainside_pickscore.yaml` turns it on). It also fails loudly
rather than silently dropping the source image, which is the better failure mode.
Happy to file it separately — it wants a one-line `_tile_conditions` override,
which does not belong in a diffusers-compat fix.

CjhHa1:
Filed the Edit-Plus batched-replay defect separately as #319 (one `_tile_conditions` override, +23/-2, off `main`). It is independent of this PR in both directions — #319 does not touch `txt_seq_lens`, and this PR does not touch the tiler — so they can be reviewed and merged in any order.

zzhuoxin1508 (approved):

PR #289, opened 2026-07-31
feat(async): dynamically dispatch prompt rollouts for AR and DiT

zzhuoxin1508:
## Summary

- Replace batch-level async AR/DiT rollout dispatch with dynamic per-prompt scheduling to reduce tail latency from slow prompts.
- Keep rollout workers saturated with per-worker lanes and immediate refill, overlapping the next generation wave with reward scoring and training.
- Preserve correctness across weight publications by bounding admission and staleness, tracking output versions, and retaining hard eval/save/final boundaries.

## Changes from `main`

- Map each DP rollout worker to an independent engine slot. Prompts enter a shared queue, and `RolloutPool` assigns the next prompt to a worker with available capacity instead of statically partitioning requests.
- Use `per_worker_inflight` to control how many prompts each worker can process concurrently. When a request completes, the scheduler immediately fills the available slot.
- Start training after the first `batch_size` complete prompt groups are collected, without waiting for every prompt in the original input batch.
- Refill rollout capacity after collection and before reward scoring, allowing the next generation wave to overlap with current reward computation.
- Bound rollout admission by `max_inflight`, `buffer_max_staleness`, eval/save boundaries, and the final training boundary.
- Decouple `weight_sync_interval` from the rollout staleness budget so valid buffered work can survive weight publication. Reject staleness budgets smaller than `weight_sync_interval - 1`.
- Read and submit complete dataset batches; partial refills are rejected.
- Track `output_version` on rollout results and reject training batches that mix different rollout weight versions.
- Derive worker actor concurrency from `per_worker_inflight` when omitted, preserving capacity for control calls.
- Remove the async AR rollout-DP divisibility requirement while keeping train-DP divisibility unchanged.
- Keep synchronous trainers, checkpoint formats, and dataset formats unchanged.

## Test Plan

- [ ] Complete the updated clean 100-step BAGEL benchmark (72/100 at the latest measurement).
- [x] Run a 10-step Qwen3 async AR A/B test in the same 16-GPU environment.
- [x] Verify both async runs complete without OOM and report `stale_dropped=0`.
- [x] Run Python compile, Ruff format/lint, Hydra configuration, whole-batch admission, and partial-refill rejection checks.

### Diffusion benchmark (updated)

Configuration:

```yaml
num_devices: 8
batch_size: 16
train_fraction: 0.5
max_inflight: 1
per_worker_inflight: 1
weight_sync_interval: 1
buffer_max_staleness: 2
sampling:
  samples_per_prompt: 16
  num_inference_steps: 14
  height: 512
  width: 512
```

- Workload: 16 prompts × 16 samples = 256 images per step, using 4 train GPUs and 4 rollout GPUs.
- PR #289 fixed run: **135.15 seconds/step** over 71 measured intervals (72/100 steps complete at update time).
- The run remains healthy with no TensorStore errors; model weights are loaded from local `/root` storage.
- [W&B run](https://wandb.ai/zhouzhuoxin1508-tencent/bagel-editreward/runs/rxeo3v84)
- The previous 186.48/175.94 seconds-per-step measurements came from an outdated setup and are excluded from this benchmark.

### AR benchmark

Configuration:

```yaml
max_inflight: 2
weight_sync_interval: 1
buffer_max_staleness: 2
```

- `fully_async`: 112.1 seconds/step on average, with an average reward of 0.1527.
- PR #289: about 90.0 seconds/step on average, with an average reward of 0.1395.
- Average step time decreased by about 19.7%, corresponding to about 24.6% higher throughput.
- Both runs completed without OOM, and `stale_dropped=0` in both runs.


zzhuoxin1508:
bagel test result
Clipboard_Screenshot_1786349669

zzhuoxin1508:
qwen3 test result
max_inflight : 2
weight_sync_interval: 1
buffer_max_staleness: 2
Clipboard_Screenshot_1786436142

CjhHa1 (approved):
LGTM

PR #284, opened 2026-07-31
chore(scripts): drop ad-hoc verify harnesses, move dataset converters to datasets/

haonan3:
## Summary

Repo policy since #99/#267 is **no unenforced test tree** — yet `scripts/` was quietly regrowing one: #158 left seven `ep_verify/` torchrun harnesses and #210 a KL-batching regression script. This PR:

- **Deletes** `scripts/ep_verify/` (7 files) and `scripts/verify_refl_kl_batching.py`. They were verification evidence for their PRs; they remain recoverable from git history (#158 / #210) if anyone needs to re-run them.
- **Moves** the two dataset converters next to the data they produce, following the existing `datasets/<name>/` convention (cf. `datasets/pickscore/prpocess.py`):
  - `scripts/convert_video_r1_260k_to_unirl.py` → `datasets/video_r1_260k/`
  - `scripts/convert_daily_omni_dataset_format_to_unirl.py` → `datasets/daily_omni_av/`

  The qwen3-omni recipes' default `data_path` already points into exactly these directories, so converter output now lands beside the converter.
- Fixes a copy-paste bug in the daily-omni converter docstring (its example invoked the video-r1 script) and updates self-referencing paths.

`scripts/` now holds only `check_recipe_targets.py` — the pre-commit/CI guard, untouched here (#283 is reworking it separately).

## Related Issue

N/A

## Test Plan

- `python scripts/check_recipe_targets.py` → `check-recipe-targets: 2310 unirl _target_ paths resolve.`
- `python -m py_compile` on both moved converters → OK
- `ruff check` + `ruff format --check` on the moved files → clean
- `git grep "scripts/convert_\|scripts/ep_verify\|scripts/verify_refl"` → no references anywhere (the removed/moved files were referenced nowhere outside `scripts/`; no CI workflow touches them)

## Compatibility / Risk

- No shipped-code changes: neither `scripts/` nor `datasets/` is packaged (wheel includes `unirl.*` only).
- ⚠️ **Merge-order interaction**: #279 and #280 branched before #210 merged and still carry `scripts/verify_refl_kl_batching.py` as an **addition** — merging them after this PR without a rebase would resurrect the file. Please drop it from those branches when rebasing.
- #283 only touches `scripts/check_recipe_targets.py`; no overlap.

## Reviewer Notes

[...]

PR #281, opened 2026-07-30
feat(experimental): port SD3 image ReFL and remove the legacy core ReFL path

haonan3:
## Summary

Ends the two-ReFL transition: ports SD3 image ReFL onto the `experimental.refl` contract and deletes the legacy core path. **Stacked on #210** — the diff collapses to the last commit once #210 merges.

- **Port** (`experimental/refl/models/sd3.py` + `examples/sd3_pickscore_refl.yaml`): the legacy DRaFT-K flow re-expressed through the package contract — `diffuse_with_grad` (mid/final grad window, per-sample `[B]` KL), `decode_with_grad` reusing the mainline grad decode (pixels in `[0,1]` for the core image scorers). Hyperparameters preserved from `refl_sd3.yaml`; the reward pairs the **core** PickScore scorer — no package-local reward code. `pipeline_target` is now the only thing a config swaps between WAN video ReFL and SD3 image ReFL, demonstrating the boundary is not WAN-shaped.
- **Deliberate non-support**: CFG under BPTT (`guidance_scale` must be 1.0, as the legacy path always ran) — fails loudly.
- **Removed**: `unirl/train_refl.py`, `unirl/trainer/refl.py`, `unirl/train/refl/` (ReFLPolicy), `unirl/models/draft.py` (its only consumer), `examples/diffusion/refl_sd3.yaml`; trainer README updated. `eval_suites` stays (used by DiffusionTrainer/PETrainer). Rationale: ReFL bypasses the core `StageAlgorithm` abstraction (no replay, no advantages) and per the tier design lives in `experimental/`; net core reduction ≈ −600 lines.

## Test Plan

Static (CPU, at this head): compose-check ×3 (`sd3_pickscore_refl`, both WAN configs) rc=0; `scripts/check_recipe_targets.py` — 2311 paths resolve (no dangling refl dotpaths); repo-wide grep confirms zero references to the removed modules; `pre-commit run --all-files` green.

Pending GPU before merge:
- [x] `sd3_pickscore_refl` smoke — DONE, see the smoke comment (10/10, clean exit, PickScore-range rewards, grads flow, ~1.3s/step) (8 GPUs, ~10 rollouts): reward rises as the legacy path did; validates the port end to end.

## Compatibility / Risk

- Launch surface change: `python -m unirl.train_refl` is gone; the replacement is `python -m experimental.refl.run --config-name=sd3_pickscore_refl`. No deprecation shim by design (experimental-tier policy: fail loudly).
[...]

PR #280, opened 2026-07-30
docs(algorithms): placement policy — core canon vs experimental variants (RFC)

haonan3:
## Summary

RFC: writes the **algorithm placement policy** into `unirl/algorithms/README.md`. **Stacked on #210** (references the experimental tier); the diff collapses to the last commit once #210 merges.

Three rules: (1) core keeps the canon — reference family bases (GRPO, FlowGRPO, SFT) and team-flagship published algorithms (CPPO, DRPO, FlowDPPO); (2) model-specific one-offs and unpublished variants start in the `experimental/` package that needs them, selected via `_target_` exactly like core algorithms (trainers unaffected); (3) movement in either direction is a deliberate PR with config dotpaths migrated in the same change (`check_recipe_targets` gates stale paths).

**No algorithms move in this PR** — it establishes the policy so in-flight and future algorithm PRs choose their home at land time. Comment here if you want the lines drawn differently.

## Test Plan

Docs-only. `pre-commit run --all-files` green.

## Compatibility / Risk

None (documentation). Adoption applies to future PRs; existing algorithms are untouched until a dedicated placement PR moves them.

AI-assisted; reviewed and directed by the maintainer.

PR #290, opened 2026-07-31
fix: post-merge review sweep — lint blind spots, bagel prefill guard, label workflows, refl nits

haonan3:
## Summary

Fix batch from an adversarial post-merge review of the five recently merged PRs (#210, #277, #279, #284, #286). Every item below was reproduced/verified before fixing; the review evidence lives in the per-item notes.

**Boundary lint had systematic blind spots (#279), and they already mattered on the tree:**
- The cross-package rule missed `from experimental import <sibling>`, bare `import experimental`, relative imports climbing past the package root, and treated top-level `experimental/*.py` (incl. `__init__.py`) as a lint-free shared space — the most natural spellings of exactly the sideways borrowing the README forbids. All caught now, in any spelling.
- The "core dependency" set was `[project.dependencies]` only, so every extras pin — torch, torchvision, sglang, vllm — was re-declarable undetected; the in-tree `videoalign/requirements.txt` was already re-declaring `torchvision` with no signal. The set now unions base deps, every `[project.optional-dependencies]` extra, and `[tool.uv] override-dependencies`; the scan covers `requirements*.txt` and fails closed on pip option/include lines and URL/VCS lines it cannot vet. The videoalign re-pin is dropped (comment documents that the engine extra provides it) and the face extras file documents the same for its top-level `torchvision` import.
- New tier-direction rule in `check_recipe_targets.py`: recipes outside `experimental/` may not `_target_` into `experimental.*`, and a package's recipes may not target a sibling package — `_target_` is this repo's primary wiring mechanism and previously carried zero directional signal.
- `experimental/README.md` now states which rules are script-enforced (1, 2, 4) vs review-enforced (3, 5, 6), and that private packages are lint-gated locally only.

**Bagel (#277 follow-up):** `BagelPipeline._build_contexts` runs the same vendored inference-signature prefills the #277 fix declared eval()-load-bearing, but with no guard — SFT/trainside callers reaching it in train() get the exact `packed_query_sequence` TypeError back. Same guard-and-restore applied.

**#284 follow-ups:** the two moved converters landed inside gitignored `datasets/*` (tracked but invisible to ruff/rg/editors; new sibling files needed `add -f`) — the dirs are whitelisted with their jsonl outputs kept local-only. Two recipe comments pointing at nonexistent `scripts/*.sh` launchers now point at `examples/run_experiment_single_node.sh`. `experimental/private_*` also covers stray private files now (was dir-only).

**Label workflows (#286 follow-ups):** merged/closed PRs shed their status label (they kept `need review` forever — see #210/#279 right now); `synchronize` self-heals a PR that lost its label to a missed event/transient failure; the review-labels consumer drops its shared concurrency group (GitHub keeps one pending run per group, so reviews landing on two PRs within seconds silently dropped the earlier sync).

**refl nits (#210 follow-ups):** `seed: null` no longer TypeErrors at the worker (`int(params.seed)` had no None branch; `generate_latents` already supports unseeded); the unreachable grad-input mismatch guard in `worker.py` is dropped (its dict-index comprehension KeyErrors first, and `resolve()` consumes the same keys earlier still).

## Related Issue

N/A

## Test Plan

- Negative/positive battery for both guards (34 checks, all pass): every blind spot above rebuilt as a synthetic violation in a lab tree — from-import/bare/relative/`__init__`-hub sideways imports flagged; extras pin, uv-override pin, case-variant name, `-r` include, VCS line, and `requirements-gpu.txt` coverage flagged; self-imports (absolute/relative/from-bare), core imports, additive names, and `name @ url` direct refs stay clean; core→experimental and sibling `_target_`s flagged while own-package recipes pass with no unresolved noise.
- Real tree: `python lint/check_experimental_boundaries.py` → ok; `python lint/check_recipe_targets.py` → `2310 recipe _target_ paths resolve.` (unchanged count).
- `git check-ignore` battery: both converters no longer ignored; their `*.jsonl` outputs, other `datasets/*` dirs, `experimental/private_notes.py`, and `experimental/private_a/f.py` all ignored.
- `python -m py_compile` on all six edited `.py` files; `bash -n` on both edited workflow scripts (extracted from YAML).
- `SKIP=no-commit-to-branch pre-commit run --all-files` → all 16 hooks pass (both hardened guards included).
- Not run: GPU smoke for the bagel guard path (CPU box) — the change is mode-neutral by construction (guard-and-restore around an already-`no_grad` block, same pattern as the merged #277 fix); and the `master_dtype: fp32` verification gap flagged for #210 is deliberately NOT addressed here (owner action on the pinned-torch env, see Reviewer Notes).

## Compatibility / Risk

- The boundary/targets guards get strictly stricter; the current tree passes clean. Future violations that previously slipped through will now fail pre-commit/CI — that is the point.
- Label workflows: behavior changes are label-lifecycle-only (`closed` cleanup, self-heal on `synchronize`, no shared concurrency). The `pull_request_target` consumers still check out nothing and interpolate no PR-controlled strings; the PR-number regex guard in the review consumer is untouched.
- Bagel: guard-and-restore around prefills; `freeze`/rollout paths unaffected (colocate short-circuits before this call).
- refl: `seed: null` now takes the supported unseeded path instead of crashing.

## Reviewer Notes

- AI-assisted (Claude Code): five parallel review agents (one per merged PR) + orchestrator hand-verification of every HIGH/MED finding before fixing; the submitter reviewed the full diff. Duplicate-work check: no open PR touches these files (#280 touches only an algorithms README, #281 touches refl but none of these lines — the wan21/wan22 seed lines may need a trivial rebase there).
- Deliberately not in this PR: (1) #210's `master_dtype: fp32` default was never GPU-verified at the merged head (the recorded smoke ran torch 2.7.1 with fp32 dropped; the restore commit postdates it and the yaml itself says it needs the pinned torch ≥2.11) — needs an owner smoke on the pinned env + README table update; (2) narrowing the label workflows' `issues: write` — likely redundant per the 403 evidence, but should be verified on a real labeled run after this merges rather than narrowed blind; (3) a backfill dispatch (`gh workflow run pr-review-status-labels.yml`) to reconcile the stale labels #286's test plan promised — one manual command, listed for the maintainer.

## Checklist


haonan3:
`bca7e1e8` addresses an external (Codex) review round — all three findings verified real and fixed:

1. **Same-PR label race (P2)**: removing the global concurrency group traded cross-PR pending-slot drops for same-PR parallel runs racing label writes (and `set -e` dying on the 404 a parallel DELETE leaves). Now: per-PR group `pr-review-status-labels-${{ workflow_run.pull_requests[0].number || run_id }}` — same-repo PRs get back the serialize-latest semantics, fork PRs (empty `pull_requests`) fall back to per-run with a tolerant DELETE (`|| true`) absorbing the benign 404.
2. **Tier-root recipe bypass (P2)**: `experimental/root_escape.yaml` targeting `experimental.refl.*` slipped through the `len(parts) > 2` sibling guard. A recipe at the `experimental/` root now errors — a recipe belongs inside one package.
3. **No-space PEP 508 direct refs (P3)**: `name@git+https://...` was mis-flagged unparseable (only `" @ "` was recognized). `@` joined the follower set — the name still gets extracted and checked against the core set (`transformers@git+...` correctly reports as a re-declared core dep), while bare `git+`/URL lines still fail closed. Kept the stdlib-only parser rather than importing `packaging`.

All three reproduced as lab cases in the negative battery before fixing (39 checks, all pass); full `pre-commit run --all-files` green.

PR #296, opened 2026-08-02
fix(vllm-omni): resolve HI3 DiT LoRA module names and fused-qkv layout

heguangxin:
> **Status: re-measured (2026-08-19).** The original diagnostics were on base `3a7330e` (~200 lines of LoRA-delivery drift since, in `patches/runtime.py` and `weight_sync/lora/*`). The `active_layers` result has now been re-measured as a fix-OFF vs fix-ON A/B on a live **TP8 vLLM-Omni HI3 it2i** training — `0/64 → 64/64` (details in Test Plan). **Caveat:** the re-measure ran on the extracted HI3-on-vLLM-Omni line (base `e1a814ff`, the two fix patches identical line-for-line to this head), **not a bare `main` HEAD** — that line ("the rest") is not on `main` yet, so a bare HEAD cannot bring this training up. `noise_pred` MAE was not re-measured this round. The namespace miss itself is confirmed still present on `8d77a54`.

## Summary

Two patches in `vllm_omni/patches/runtime.py`, both wrapping `DiffusionLoRAManager._get_lora_weights`:

* `patch_dit_hi3_lora_module_alias()` — after the stock name returns `None`, retries `transformer.layers.*` as `model.layers.*`.
* `patch_dit_hi3_fused_qkv_lora_layout()` — de-interleaves GQA-packed fused `qkv_proj` LoRA-B before it reaches vLLM's merged qkv linear. Every path that cannot establish the layout declines _and warns_, deduplicated by reason so a 64-layer model yields at most three lines; returning the interleaved weights quietly is the corruption this patch exists to prevent.

The repack refuses to install unless the alias is already installed: reversed, the alias would wrap the repack instead of feeding it, silently leaving qkv interleaved. Both patches are needed — the alias alone fixes `o_proj` and leaves `qkv_proj` wrong.

**The bug.** On HunyuanImage-3 the DiT LoRA never reached the vLLM-Omni rollout engine, and failed silently: training kept updating an adapter the sampler never consumed, so the engine sampled the frozen base model. Reward stays flat, nothing raises. The read-back that exists to catch this does not run either — `sync.verify` defaults to `False` (`weight_sync/lora/base.py:80`) and no HI3 recipe enables it. Turning it on does not help yet: at TP>1 the read-back returns only rank 0's reply and fails the same verification's rank-count check, which a companion PR fixes.

**Why engine-side and not config.** `RemoteLoraWeightSync` applies `sync.param_prefix` once in `LoraWeightSyncBase._extract()`, then pushes that one keyed copy into both HI3 engines. One value, two consumers that want opposite things: AR holds the decoder at `model.layers.*` while the trainer's `trainable_module()` yields `layers.*`, so `param_prefix: "model."` bridges that correctly; DiT is registered as `transformer.layers.*` and its stock lookup strips `transformer.` and asks for the bare `layers.*`. Neither value is wrong and no single value satisfies both, so the correction belongs in the only component that knows it is the DiT engine. The other `param_prefix: "model."` models (`qwen3_5`, `qwen3_omni`, `qwen_vl`) each have one AR consumer; HI3 is the only model pushing one adapter into two differently-named engines.

**Why #140 does not cover the second half.** `patch_ar_merged_lora_fused_tensor` hooks `MergedQKVParallelLinearWithLoRA.set_lora`, a vLLM layer entry point. The DiT stage resolves adapters through `DiffusionLoRAManager._get_lora_weights` and never reaches that `set_lora`, so fused `qkv_proj` stays GQA-interleaved and every delta lands on the wrong output rows. The diagnostic base already contained #140 and DiT qkv was still wrong. Rather than add a second implementation, this hoists #140's `_deinterleave_gqa` to a module-level `_deinterleave_fused_qkv_lora_b` shared by both hooks. The AR path keeps the same behaviour — the helper moved verbatim, and the shim only drops a now-redundant local `import torch`.

## Related Issue

N/A.

## Test Plan

* `SKIP=no-commit-to-branch pre-commit run --files unirl/rollout/engine/vllm_omni/patches/runtime.py` — all hooks pass.
* `python -m py_compile` over the changed file — clean.
* **Re-measurement (2026-08-19), fix-OFF vs fix-ON A/B**, the only variable being whether the two patches install (env gate). Live TP8 vLLM-Omni HI3 it2i (EditReward) training; a per-slot read-back over the 64 hi3 DiT `transformer.layers.*.{qkv_proj,o_proj}` resolutions:
  * **fix OFF** (patches skipped): `resolved=0, missed=64` → **`active_layers 0/64`** — reproduces the bug, every hi3 DiT LoRA lookup misses.
  * **fix ON** (patches installed): `resolved=64, missed=0` → **`active_layers 64/64`** — fix works.
  * With fix ON, `ratio = exp(HF_replay_logp − vLLM_rollout_logp)` stayed `0.9996–0.9999` across 20 rollouts (the two forwards agree once LoRA is actually synced). At rollout 1 both arms show `ratio≈1` because the adapter is still zero-init, which is exactly why `active_layers` (an install-time signal) is the discriminating metric, not `ratio`.
  * Base `e1a814ff` (extracted HI3-on-vLLM-Omni line; fix identical line-for-line to this head), **not** bare `main` HEAD — see Status caveat. `noise_pred` MAE not re-measured this round.
* Earlier diagnostics (base `3a7330e`, with a pre-fix control). Per-rank figures required added read-back instrumentation, since `collective_rpc` returns only rank 0:
   * `active_layers` `0/64 → 64/64`, `active_slices` `0/128 → 128/128`, every TP rank.
   * HF-vs-vLLM `noise_pred` MAE `~0.38 → 0.06–0.09`, i.e. down to the base-precision gap between the two forwards.

## Compatibility / Risk

No config, recipe, checkpoint or public API change. No-op for every non-HI3 modality: the alias fires only after the stock lookup returns `None` and only for `transformer.layers.*`; the repack fires only on `.qkv_proj` with a 3-way `output_sizes` matching `lora_b`. Repo-wide the only LoRA recipes targeting `qkv_proj` are the three HI3 ones (`hi3_it2i`, `hi3_vllmomni`, `hi3_vllmomni_veomni_ep`); the other diffusion backbones target separate `q`/`k`/`v` projections. The AR shim is touched only by that hoist; its behaviour is unchanged.

## Reviewer Notes

* **Re-measurement done** (`active_layers 0/64 → 64/64`, 2026-08-19). It ran on the extracted HI3-on-vLLM-Omni line (base `e1a814ff`), **not** a bare `main` HEAD, because that training line is not on `main` yet; if a bare-HEAD run is a hard requirement, the HI3 vLLM-Omni training line must land first. The `active_layers` read-back used for it is a throwaway diagnostic, not part of this PR.
* The per-rank read-back this PR's evidence relies on is itself fixed by a companion PR; before that lands, `sync.verify: true` at TP>1 fails on the rank count rather than on anything here.
* Extracted from an in-progress HI3 it2i-on-vLLM-Omni line; the rest (direct DiT-only it2i adapters, batched rollout, TP-split replicas) follows separately.

## Checklist

* I reviewed the changed code and removed unrelated/generated artifacts.
* I updated tests, docs, and configs where needed, or explained why not.


CjhHa1 (approved):
LGTM

PR #294, opened 2026-08-02
fix(rollout): guard agentic sync and deterministic diffusion eval

haonan3:
## Summary
- make `AsyncAgenticRolloutEngine.sync_weights()` the single Agentic push/version transaction, reject sync while a drive is live, and log each successful sync on the driver
- route async and partial Agentic resume/boundary sync sites through that guarded transaction
- make deterministic diffusion eval (`eval_eta <= 0`) a pure-ODE request by clearing the SDE gate and scheduler

The earlier batch-engine version-ledger and resident-policy eval changes have been removed. Batch AR/diffusion policy-version semantics are handled by #310; eval there keeps the current-policy contract (`P=U` at the hard boundary).

## Related Issue

Follow-up to #287. Complementary to #310; this PR now contains only the independent Agentic guard/logging and diffusion eval fix.

## Test Plan

- [x] `ruff check` and `ruff format --check` on all changed Python files
- [x] `python3 -m py_compile` on all changed Python files
- [x] CPU smoke: Agentic sync is rejected during a live drive without pushing, then succeeds and advances the ledger when idle
- [x] Pre-push hooks: YAML/TOML/AST/conflict/secret/Ruff/recipe-target/boundary checks

## Compatibility / Risk

- Agentic out-of-tree callers of `bump_weight_version()` must use `sync_weights(weight_sync)` so push and version advancement cannot diverge.
- Deterministic diffusion eval changes only contradictory `eta=0` + SDE-gate requests; stochastic eval is unchanged.
- No async batch eval or deployment-cadence behavior changes remain in this PR.

PR #297, opened 2026-08-02
refactor(distributed): make handle_fn and launch_nowait share the call halves

haonan3:
## Summary

Extract `_launch_half` (dispatch → localize → execute) and `_collect_half` (ray.get → rebind → collect) on `Handle`; `handle_fn`, `launch_nowait`, and `PendingHandleCall.result` now sequence the same two methods instead of carrying keep-in-parity copies — the "keep the halves in parity" comment convention (#287's own listed follow-up) is deleted. No behavior change: the grad-context preamble/postamble stays in `handle_fn`; `launch_nowait` stays `grad_mode=False`/`call_id=None`.

Split out of #295 per review: this is a distributed-RPC-layer change with its own reviewers and regression surface, independent of the rollout refactor that consumes the seam.

## Related Issue

N/A (follow-up to #287).

## Test Plan

- `ruff format --check` + `ruff check` over `unirl/distributed/group/handle.py` — clean; `python3 -m compileall` — clean.
- Pure extraction verified by line-diff: both halves are byte-identical to the previous inline bodies; call sites are thin sequencing.
- GPU: Not run; reason: CPU-only dev box. Any sync + async smoke exercises both entries (`handle_fn` on every @distributed call; `launch_nowait` on the async trainers); the #295 validation trio covers it.

## Compatibility / Risk

No public surface change; `_execute_all`/`_rebind_tree`/`_method_configs` untouched. Riskiest spot: none identified — the halves are verbatim relocations.

## Reviewer Notes

AI-assisted (Claude Code). Suggested land order: this → #294 → #295 (all touch adjacent layers; each rebases mechanically).

## Checklist

### DIFF
--- a/unirl/distributed/group/handle.py
+++ b/unirl/distributed/group/handle.py
@@ -350,12 +350,8 @@ class PendingHandleCall:
         if self._consumed:
             return self._value
         handle = self._handle
-        results = ray.get(self._refs)
-        results = [
-            handle._rebind_tree(r, handle.workers[i], worker_local=self._worker_local) for i, r in enumerate(results)
-        ]
         _, _, collect_fn, _ = handle._method_configs[self._method_name]
-        self._value = collect_fn(handle, results)
+        self._value = handle._collect_half(collect_fn, self._refs, worker_local=self._worker_local)
         self._consumed = True
         return self._value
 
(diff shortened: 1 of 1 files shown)


haonan3:
Folding back into #295 per team decision — the handle-halves commit stays cleanly separated as that PR's first commit; reviewers can review it per-commit there.

PR #302, opened 2026-08-03
feat(trainer): add opt-in diffusion role residency

nussejzz:
## Summary

- Add independent diffusion lifecycle policies for train offload during generation, train offload during reward scoring, and rollout sleep after generation.
- Preserve existing defaults while allowing measured Train/Rollout/Reward residency on colocated GPUs.
- Make rollout wake/generate/eval cleanup exception-safe, including physical vLLM-Omni rollback when LoRA restoration fails.
- Validate prompt-tree DP geometry separately from flattened training rows and make FSDP2 activation-checkpoint replay re-enter sharding hooks without duplicating KV storage.

## Related Issue

N/A

## Test Plan

- `uvx --from pre-commit pre-commit run --files <all changed PR1 files>`
- `python -m compileall -q unirl/train_diffusion.py unirl/trainer/diffusion.py unirl/train/backend/fsdp/wrap.py unirl/rollout/engine/vllm_omni/engine.py`
- Latest-upstream top-of-stack Hydra compose:
  `python -m unirl.train_diffusion --config-name=diffusion/bagel/bagel_it2i_vllmomni --cfg job --resolve rollout_sleep_after_generate=false offload_train_during_reward=false`
- End-to-end smoke on 8x NVIDIA H20 96GB with `ByteDance-Seed/BAGEL-7B-MoT`, OpenGPT image-edit data, FSDP8, vLLM-Omni DP8/TP1, LoRA rank 64, 8 prompts x 8 samples, and two optimizer updates per rollout.
- Result: one complete rollout/reward/advantage/train step, reward `-1.0735`, ratio `1.0000±0.0000`, grad norm `0.0001`, peak HBM `88,430–88,538 MiB/card`, exit code 0, clean GPU teardown.
- No standalone test files are added; validation is pre-commit, config composition, compile checks, and the real DP8 smoke above.

## Compatibility / Risk

- Existing behavior is unchanged by default: train offload is disabled and rollout sleep remains enabled unless recipes opt in.
- `offload_train_during_reward` is additive; a reward on a disjoint `reward_fraction` slab never triggers train offload.
- No checkpoint or Sample/Part format changes.
- The highest-risk areas are FSDP2 activation-checkpoint wrapping and partial distributed lifecycle rollback; both were exercised by the top-of-stack H20 run.

## Reviewer Notes

- This is PR 1 of a three-PR stack. PR 2 adds the managed rank-affine reward child; PR 3 adds the BAGEL it2i consumer.

## Checklist


unirl/rollout/engine/vllm_omni/engine.py
```diff
@@ -178,34 +187,54 @@ def sleep(self) -> None:
...
+            except Exception:
+                self._wake_failed = True
+                logger.exception("vLLM-Omni rollback sleep failed after LoRA restore error")
+            else:
+                self._is_offloaded = True
+                self._wake_failed = False
+                self._weight_sync.mark_weights_released()
             raise
-        self._is_offloaded = False
+        self._wake_failed = False
```
CjhHa1: **P1 — A wake retry can still silently serve base weights.** `restore_lora_after_wake()` sets `_lora_loaded = False` when the re-push fails (`weight_sync.py:256`). On either a rollback-success retry or a degraded-wake retry, the next restore call then returns early at `weight_sync.py:235`; execution reaches this line, clears the failure latch, and `generate()` proceeds with `attach_lora=False`. Please preserve the intent that an adapter should be active separately from whether it is currently loaded, and clear this latch only after the cached adapter has actually been re-pushed successfully.

unirl/train/backend/fsdp/wrap.py
```diff
@@ -119,6 +146,24 @@ def fsdp_wrap(
...
 
+    if activation_checkpointing:
+        from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import (
+            CheckpointImpl,
+            apply_activation_checkpointing,
+            checkpoint_wrapper,
+        )
+
+        block_ids = {id(layer) for layer in block_instances}
+        apply_activation_checkpointing(
```
CjhHa1, line 148: **P1 — This changes parameter namespaces before post-wrap loading.** `CheckpointWrapper` makes the parent model expose names such as `layers.0._checkpoint_wrapped_module.proj.weight` through `named_parameters()`, while its `state_dict()` remains canonical (`layers.0.proj.weight`). The post-wrap loaders use `named_parameters()` names for checkpoint coverage and LoRA `base_layer` remapping (`sharded_load.py` and the HunyuanImage3 materializer), so AC + meta-init can report near-zero coverage or leave base weights uninitialized. I reproduced the namespace mismatch with PyTorch's wrapper. Please make the wrapper name-transparent or normalize this prefix in every post-wrap name consumer.

unirl/trainer/diffusion.py
```diff
@@ -50,6 +134,8 @@ def __init__(
         train_fraction: float = 0.5,
         reward_fraction: float = 0.0,
         enable_fsdp_offload: bool = False,
+        offload_train_during_reward: bool = False,
```
CjhHa1, line 276: **P2 — This option is not wired into async diffusion.** `train_async_diffusion.py` does not forward it, and `AsyncDiffusionTrainer._score_completed()` calls `reward.score_and_attach()` directly instead of entering `_reward_phase()`. With the default `reward_fraction=0`, async reward scoring can share the train slab, so setting this documented option is silently ignored and can still OOM. Please forward the option and wrap async scoring in `_reward_phase()`, or explicitly reject the unsupported combination.

unirl/trainer/diffusion.py
```diff
@@ -440,6 +528,67 @@ def _build_request_sample(
             request = request.with_parts([*request.parts[:-1], frontier])
         return request
 
+    def _offload_for_reward_phase(self) -> bool:
+        """Whether a colocated reward may temporarily borrow the train cards."""
+        return self._offload_train_during_reward and not self._reward_is_separate and not self._uses_ema
```
CjhHa1: **P2 — An explicitly enabled reward offload is silently disabled for every EMA/DiffusionNFT run.** By the time the reward phase starts, `_generate_for_training()` has already restored the live adapter in its `finally` block, so the generation-time EMA restriction does not automatically apply here. This also conflicts with the README's unconditional shared-slab semantics. Please support reward-phase offload after EMA restoration, or fail validation and document that this combination is unsupported instead of ignoring the requested policy.

CjhHa1 (commented):
I found two correctness issues that can cause silent wrong-weight generation or break post-wrap weight loading, plus two lifecycle-policy gaps. Please address these before merging.

unirl/rollout/engine/vllm_omni/engine.py
nussejzz: Thanks for your review!

`_lora_loaded` was both "an adapter should be active" (intent)
and "the adapter is currently pushed" (state). The failure path cleared it as state,
while the restore gate read it as intent, so both retry paths (rollback-success re-wake
and the degraded-wake retry) early-returned without re-pushing.

The fix gates `restore_lora_after_wake()` on `_last_lora_tensors` — the cached adapter
is the durable intent, and it survives a failed re-push. `_lora_loaded` remains
state-only: cleared on a failed push, set back to True only by a successful
`set_lora_from_tensors*`. Base-model runs (nothing ever cached) keep the early return.

Validated with a state-machine simulation covering six scenarios including both of your
retry paths, plus a negative control: with the old gate the simulation fails at
"PATH A retry must re-push, not early-return"; with the new gate all six pass. Also
re-verified end-to-end on 2 nodes × 8 H20 (sleep-mode 2-rollout run).

unirl/train/backend/fsdp/wrap.py
nussejzz: Thanks!

I add a small utility function: wherever a name needs to be compared, first delete the shell name segment `_checkpoint_wrapped_module` before comparing.

unirl/trainer/diffusion.py
nussejzz: `AsyncDiffusionTrainer` now fails validation at startup when `offload_train_during_reward` is set, with an error explaining that async scoring runs at reap time outside `_reward_phase()` (so the
option would otherwise be silently ignored and a train-slab reward could still OOM).

unirl/trainer/diffusion.py
nussejzz: The hidden `not self._uses_ema` clause is removed from `_offload_for_reward_phase()`; instead the trainer now raises at startup when `offload_train_during_reward` is combined with an algorithm that sets `requires_ema_rollout` (EMA/DiffusionNFT), stating that the combination is unvalidated against `backend.ema` state rather than silently ignoring the policy.

haonan3:
Pushed a maintainer refresh to unblock this stack — the branch was written against a three-week-old main and had been un-mergeable since #232 landed.

**`d748311f` — merge of current main (semantic resolution of `unirl/trainer/diffusion.py`, 11 hunks).** The load-bearing decisions, so review can focus there:

- This branch's per-rollout method takes main's name **`_rollout_and_score`**; main's window-level `train_step` (#232: accumulation windows, pooled logging, transfer-queue reclaim seam) is the only `train_step`. The five residency helpers graft under it unchanged; the per-rollout timer stays in the window step.
- **Reward is optional on main**: the DP-geometry validation passes `reward_dp_size=1` without a `reward:` block, and `_rollout_and_score` enters `_reward_phase()` only when a reward exists — reward-less (OPD) recipes skip the offload window entirely.
- **evaluate/_eval_pass keep #346's contract** — the frozen `eval_sampling` overlay (this branch's `eval_cfg_text_scale` rebuild is gone; the attribute no longer exists on main) and the eval media path (`media_prefix` / `_log_eval_media` / the rewards rehydration write-back), composed with this branch's chunk-level wake/sync/sleep plumbing. Media logging sits outside `_reward_phase()` so the upload doesn't hold the train-offload window open.

**`fd485512` — three follow-ups:**
- **`ac_wrap_order` now defaults to `outside`.** The old forward-monkeypatch checkpoint never re-entered FSDP hooks on recompute, so `outside` is the composition all 71 existing AC recipes (including the shipped BAGEL family) have been running — defaulting to `inside` silently flipped them all. `inside` stays as the per-recipe opt-in; the bagel it2i recipe pins it in PR3 (the order its smoke validated), the HI3 recipes keep their explicit `outside` pins, and the wrap.py comment now describes the two orders without the REQUIRED claims that contradicted the shipped recipes.
- flux2_klein's deferred zero-init compares canonical names (the deferred op runs post-wrap) and fails loudly on an unmatched absent param — same bug class as the earlier post-wrap-loader fixes.
- `sleep_task`/`wake_task` check the per-stage `OmniACK`s: at the pinned vllm-omni, worker handlers swallow exceptions into `status="ERROR"`, so the new partial-transition latch never saw worker-side failures.

Verified: full `pre-commit run --all-files` (the CI lint job) green; Hydra compose of `bagel_vllmomni`, `bagel_vllmomni_async`, `sd3_opd_multi_teacher` under the real import chain; AST cross-check that both entry points pass exactly the merged constructor kwargs; exactly one `def train_step` in the merged trainer. **Not covered: a GPU smoke.** The two silent-regression risks to watch on the next real run: the eval filmstrip still uploads (`eval_interval>0` + `logging.log_media: true`), and the async startup rejection for `offload_train_during_reward` still fires.

haonan3:
GPU smoke results for the refreshed stack (8x H20, torch 2.11.0+cu126, tree = the current heads; runs on `bagel_trainside_lora` — 7B BAGEL, LoRA, colocated local PickScore, `reshard_after_forward: true`, `activation_checkpointing: true`, 14-step trainside replay):

**Run 1 — `ac_wrap_order` default (`outside`), `offload_train_during_reward=true`, 2 rollouts, eval every step:**
```
EVAL step 0  (4 spp, 14 steps, 512x512, cfg=1.0 eta=0.0)  reward=0.7780
rollout 1/2  reward=0.7826  gn=0.0001  ratio=1.0000±0.0000  clip=0.12
EVAL step 1  reward=0.7780
rollout 2/2  reward=0.6841  gn=0.0001  ratio=1.0000±0.0000
EVAL step 2  reward=0.7774
```
**Run 2 — `ac_wrap_order: inside` + `report_to_wandb` + `log_media=true`, 1 rollout:**
```
EVAL step 0  reward=0.7780      <- identical to Run 1's deterministic eval
rollout 1/1  reward=0.7864  ratio=1.0000±0.0000
EVAL step 1  reward=0.7773
```

What this establishes:

- **Both AC compositions are correct on the multi-use replay path.** `ratio=1.0000±0.0000` in both orders means replay log-probs match rollout log-probs exactly through the checkpointed recompute, and the deterministic eval reward is bit-stable across orders (0.7780 == 0.7780). The default flip to `outside` (= the pre-knob composition all existing AC recipes ran) is validated on the exact 7B/reshard/multi-step case the old comment claimed required `inside`; `inside` works as the opt-in.
- **The merged trainer path is live end to end**: window `train_step` → `_rollout_and_score` → `_generate_for_training`, `_reward_phase` offload/onload around colocated scoring (`offload_train_during_reward=true` exercised every rollout and eval chunk; `perf/{wake_up,sleep,reward}_time_s` all report).
- **#346's eval media path survived the merge**: the offline wandb run contains `files/media/images/eval/*.png` (8 files) and `media/images/rollout/*.png` — the filmstrip regression flagged during the resolution is provably absent.

Environment notes for anyone re-running: the box needed `torch==2.11.0+cu126` + matching `torchvision==0.26.0` (torch 2.7's FSDP2 still enforces uniform param dtype and rejects the repo's fp32-LoRA-over-bf16-base wrapping — repo floor, not a stack issue), plus `opencv-python-headless`/`sentencepiece` for the BAGEL vendor transforms; `BAGEL_PATH` must point at a local snapshot (the bundle reads `llm_config.json` directly, hub ids do not resolve). Not yet GPU-covered: the vLLM-Omni it2i engine path and the managed EditReward child (both need the author's engine/venv assets) — the trainside slice of the stack is what this validates.

CjhHa1:
@haonan3 I am starting the remaining validation runs on `debug-16gpu-0807` against the current PR stack:

1. Current-head vLLM-Omni BAGEL it2i end-to-end smoke (rollout → reward → train → clean teardown).
2. Real vLLM-Omni sleep/wake ACK success-path validation, including stage/task correlation.
3. Controlled ERROR-ACK / partial-transition recovery validation.
4. Current PR1+PR2+PR3 managed EditReward + vLLM-Omni integration smoke.
5. Current-head HI3 meta-init + `ac_wrap_order=outside` smoke if the required model/runtime assets on the debug host are available.

I will post the commands, exact heads, and results here once the runs finish.

CjhHa1:
@haonan3 The remaining current-head validation runs have completed successfully on `debug-16gpu-0807` (worker-0: 8×H20). The runtime tree was PR #304 `4c84180` plus the PR #302 delta through `4d1be5f`; PR #303 was `00d4ce9`. I used eight deterministic synthetic source images, so the reward values below are smoke signals only, not quality benchmarks.

### 1. Real ACK contract + injected recovery

Environment: torch `2.11.0+cu129`, CUDA `12.9`, vLLM `0.20.0`, vLLM-Omni `0.20.0`.

- Actual `OmniACK` fields were `task_id/status/stage_id/rank/freed_bytes/metadata/error_msg`.
- Nested object and msgpack-dict SUCCESS ACKs passed.
- Nested ERROR ACKs, transport error envelopes, empty/no-reporter results, wrong stage IDs, and stale task IDs all failed closed.
- Injected partial-wake and LoRA-restore + rollback double faults recovered through the transition latch.
- Marker: `PR302_ACK_AND_RECOVERY_OK`.

### 2. BAGEL it2i vLLM-Omni + managed EditReward

Shape: vLLM-Omni DP8/TP1, BAGEL LoRA rank 64, batch `8×2`, two optimizer updates, two rollouts, default sleep-after-generate policy.

- All eight rank-affine EditReward children reached resident readiness and each served both score requests with HTTP 200.
- The real lifecycle path ran end to end: generate → level-1 sleep ACK → reward/train → wake ACK + LoRA restore → second generate → sleep ACK → clean teardown.
- Rollout 1: reward `0.2466`, grad norm `0.0002`, ratio `1.0000±0.0000`, clip `0.00`.
- Rollout 2: reward `0.5773`, grad norm `0.0001`, ratio `1.0000±0.0000`, clip `0.00`.
- Exit code `0`.

### 3. HI3 meta-init + `ac_wrap_order=outside`

Shape: HunyuanImage-3 80B meta-init/FSDP8, batch `8×2`, two optimizer updates, one rollout.

- Materialization verified all 64 LoRA `base_layer` parameters loaded finite.
- A standalone EditReward server on launcher GPU1 served all eight requests with HTTP 200; the existing HunyuanOCR workload on launcher GPU0 was left untouched.
- Reward `-1.7689`, grad norm `0.0001`, ratio `1.0000±0.0000`, clip `0.00`.
- Exit code `0`.

Artifacts and full logs are under `/apdcephfs/private_aimicahchen/Experiments/pr302-current-head-validation-0811/`. All planned PR #302 current-head validation is now green.

CjhHa1 (approved):
LGTM

PR #303, opened 2026-08-03
feat(reward): add rank-affine managed image scorers

nussejzz:
## Summary

- Add `ManagedScorerProcessBackend`, which launches one scorer child per reward rank with an explicit Python executable and inherited rank-local GPU visibility.
- Serve one image/image-edit scorer over an inherited loopback socket without creating a nested Ray GPU allocation.
- Add bounded request chunking, stable identity/version echo, deterministic idempotency, protocol validation, and graceful lifecycle endpoints.
- Support `resident` and `per_call` GPU residency, and make EditReward model path, dtype, head type, batching, and device movement explicit.

## Related Issue

N/A

## Test Plan

- `uvx --from pre-commit pre-commit run --from-ref origin/main --to-ref HEAD`
- `python -m compileall -q unirl/reward unirl-reward-service/reward_service`
- `python lint/check_recipe_targets.py` (`2,408` recipe `_target_` paths resolved on this tree).
- Latest-upstream Hydra compose for the managed BAGEL recipe.
- End-to-end smoke on 8x NVIDIA H20 96GB using eight rank-affine EditReward children in `/root/venvs/editreward`, each scoring its local one-prompt/eight-image shard.
- Result: all eight children started and served the complete rollout/reward/advantage/train step; reward `-1.0735`, no OOM/traceback/actor crash, exit code 0, clean child and GPU teardown.
- No standalone test files are added; the managed child, wire path, lifecycle, and cleanup were exercised by the real DP8 run.

## Compatibility / Risk

- Existing local and externally hosted `RemoteRewardBackend` paths remain available and keep their defaults.
- Managed mode initially supports only `image` and `image_edit`; text/video/audio carriers are deliberately rejected.
- The child must be given an explicit compatible Python environment and exactly one visible GPU unless multi-GPU use is intentionally enabled.
- `resident` keeps scorer weights on GPU; `per_call` performs blocking CPU/GPU model movement and should be used only when memory requires it.
- Wire additions are versioned/additive; strict identity echo is enabled only for managed children.

## Reviewer Notes

- Rebased onto current `main` (includes #302, #316 one-line docstrings, and #366 `PrimitiveValue`). The branch merges cleanly.
- Review `unirl/reward/*` + `unirl-reward-service/*`. #304 is the BAGEL it2i consumer and production recipe; it must use `history_kind` / `gpu_residency`.

## Checklist

### DIFF
--- a/unirl-reward-service/reward_service/server.py
+++ b/unirl-reward-service/reward_service/server.py
@@ -14,81 +14,19 @@ return scores. The per-reward deadline is `server.score_timeout_s`.
 from __future__ import annotations
 
 import asyncio
-import base64
-import io
 from collections import defaultdict
 from contextlib import asynccontextmanager
-from pathlib import Path
 
 from fastapi import FastAPI, HTTPException
-from PIL import Image
 
 from reward_service.config import ServiceCfg
 from reward_service.logging_utils import get_logger
-from reward_service.schemas import HistoryTurn, RewardRequest, ScoreRequest, ScoreResponse
+from reward_service.schemas import RewardRequest, ScoreRequest, ScoreResponse
 from reward_service.scorers import ScoreItem
+from reward_service.wire import request_to_item
 from reward_service.workers.pool import WorkerPool
 
 logger = get_logger(__name__)
-
-
-def _decode_image(b64: str) -> Image.Image:
-    try:
-        raw = base64.b64decode(b64)
-        return Image.open(io.BytesIO(raw)).convert("RGB")
-    except Exception as e:
-        raise HTTPException(status_code=400, detail=f"invalid image_b64: {e}") from e
-
-
-def _resolve_video(turn: HistoryTurn) -> bytes | str | None:
-    """Return the video source for a turn, in the form scorers expect.
-
-    Three branches, kept zero-copy when possible:
-
-    * ``video_path`` set → return the path string after a stat check.
-      The scorer (e.g. videoalign / decord) opens it directly on the
-      shared filesystem, no bytes round-trip.
-    * ``video_b64`` set → base64-decode and return the raw bytes; the
-      scorer is responsible for spilling to a tempfile if its decoder
-      needs a path.
-    * Neither set → ``None`` (turn has no video, e.g. an image-only
-      request that happens to ride alongside a video request in the
-      same batch).
-    """
-    if turn.video_path is not None:
-        p = Path(turn.video_path)
-        if not p.is_file():
-            raise HTTPException(
-                status_code=400,
-                detail=f"video_path does not exist or is not a regular file: {turn.video_path}",
-            )
-        return str(p)
-    if turn.video_b64 is not None:
-        try:
-            return base64.b64decode(turn.video_b64)
-        except Exception as e:
-            raise HTTPException(status_code=400, detail=f"invalid video_b64: {e}") from e
-    return None
-
-
-def _request_to_item(req: RewardRequest) -> ScoreItem:
-    if not req.history:
-        raise HTTPException(status_code=400, detail="history must not be empty")
-    history: list[tuple[str, Image.Image | None]] = []
-    videos: list[bytes | str | None] = []
-    any_video = False
-    for turn in req.history:
-        image = _decode_image(turn.image_b64) if turn.image_b64 is not None else None
-        history.append((turn.text, image))
-        video = _resolve_video(turn)
-        if video is not None:
-            any_video = True
-        videos.append(video)
-    return ScoreItem(
-        history=history,
-        videos=tuple(videos) if any_video else None,
-        metadata=req.metadata,
-    )
 
 
 def _bucket_by_reward(
@@ -172,23 +110,18 @@ def create_app(cfg: ServiceCfg) -> FastAPI:
     @app.post("/score", response_model=ScoreResponse)
     async def score(body: ScoreRequest) -> ScoreResponse:
         pool: WorkerPool = app.state.pool
+        if body.protocol_version != "1":
+            raise HTTPException(status_code=400, detail=f"unsupported protocol_version={body.protocol_version!r}")
         if not body.requests:
             return ScoreResponse(results=[], errors=[])
 
-        items = await asyncio.to_thread(
-            lambda: [_request_to_item(r) for r in body.requests]
-        )
+        items = await asyncio.to_thread(lambda: [request_to_item(r) for r in body.requests])
         buckets = _bucket_by_reward(body.requests, items, pool)
 
         timeout_s = cfg.server.score_timeout_s
-        name_to_ref = {
-            name: pool.dispatch(name, bucket_items)
-            for name, (_, bucket_items) in buckets.items()
-        }
+        name_to_ref = {name: pool.dispatch(name, bucket_items) for name, (_, bucket_items) in buckets.items()}
         gathered = dict(
-            await asyncio.gather(
-                *(_await_ref(pool, name, ref, timeout_s) for name, ref in name_to_ref.items())
-            )
+            await asyncio.gather(*(_await_ref(pool, name, ref, timeout_s) for name, ref in name_to_ref.items()))
         )
 
         results: list[dict[str, dict[str, float]]] = [dict() for _ in body.requests]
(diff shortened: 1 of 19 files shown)


unirl/reward/managed_process.py
```diff
@@ -0,0 +1,329 @@
...
+                require_identity_echo=True,
+                expected_scorer_version=config.scorer.version,
+            )
+            super().__init__(config=remote_spec, base_device=base_device)
+            # per_call means "on GPU only while scoring" — including BEFORE the
+            # first score call. The child boots resident (model loaded to CUDA),
+            # so without this initial offload it holds its full footprint through
+            # the whole first generation phase and can OOM a colocated rollout
+            # before any scoring ever happens.
+            if self.process_config.lifecycle == "per_call":
```
CjhHa1: **High — `per_call` can still OOM during bootstrap.** `_start_child()` does not return until `/health` reports a resident scorer, and EditReward defaults to constructing directly on CUDA. This offload therefore happens only after the training model and scorer have already been simultaneously resident. In the memory-sharing case where they fit only one at a time, the child can OOM before reaching this line. Please propagate an initial-offload/CPU-start policy into scorer construction (or require the equivalent of `offload_between_calls=true`) when `lifecycle="per_call"`.

unirl-reward-service/reward_service/server.py
```diff
@@ -202,6 +135,10 @@ async def score(body: ScoreRequest) -> ScoreResponse:
                 continue
             for i, sub_metrics in zip(indices, bucket_scores):
                 results[i][name] = sub_metrics
-        return ScoreResponse(results=results, errors=errors)
+        return ScoreResponse(
+            results=results,
+            errors=errors,
+            identities=[request.identity() for request in body.requests],
```
CjhHa1, line 154: **Medium — the full gateway cannot satisfy a scorer-version pin.** `RewardRequest.identity()` defaults `actual_scorer_version` to `None` and overwrites the version carried by the request, so a normal `RemoteRewardBackend` configured with `expected_scorer_version` rejects every response (`None != expected`). The direct managed server avoids this by supplying the actual version. Please either return the actual per-scorer version here, preserve the requested value when no actual version is available, or explicitly reject version pinning for the multi-reward gateway.

unirl/reward/managed_process.py
```diff
@@ -0,0 +1,329 @@
...
+
+    def compute_rewards(self, request: RewardRequest) -> RewardResponse:
+        if self.process_config.lifecycle != "per_call":
+            return super().compute_rewards(request)
+        with self._lifecycle_lock:
+            self.onload()
+            try:
+                return super().compute_rewards(request)
+            finally:
+                self.offload()
```
CjhHa1: **Medium — cleanup can mask the real scoring failure.** If scoring raises because the child crashed or became unreachable, this unconditional offload usually fails too, and its lifecycle `RequestException` replaces the original score exception. Please preserve the active exception (for example, log/tolerate offload failure only while unwinding an existing error) while still treating offload failure after a successful score as fatal.

unirl/reward/managed_process.py
```diff
@@ -0,0 +1,329 @@
...
+            "managed single-scorer client.required_rewards must equal (scorer.name,)",
+        )
+        require(
+            self.client.input_kind == "image",
+            f"managed image scorer requires client.input_kind='image'; got {self.client.input_kind!r}",
+        )
+        require(self.lifecycle in {"resident", "per_call"}, "lifecycle must be resident or per_call")
+        require(self.process.startup_timeout > 0, "startup_timeout must be positive")
+        require(self.process.shutdown_timeout > 0, "shutdown_timeout must be positive")
+        require(self.idempotency_cache_size > 0, "idempotency_cache_size must be positive")
```
CjhHa1, line 87: **Low — the idempotency cache may be smaller than one transport chunk.** If `client.request_batch_size > idempotency_cache_size` and a completed response is lost, inserting that chunk evicts some of its own rows. Retrying the same request then recomputes the evicted rows, defeating full retry deduplication and potentially repeating an expensive timeout. Please validate `request_batch_size <= idempotency_cache_size` for managed children, or retain all rows from an in-flight chunk atomically.

CjhHa1 (commented):
Inline review of the managed scorer lifecycle and wire protocol.

haonan3:
Maintainer refresh on top of the PR1 merge (`0068b58a` / `9c02e114` bring current main + the trainer resolution through the stack), plus `662b4a2f` with three substantive changes:

- **Group reaper guard.** `_spawn_group_reaper()` now arms only when the frontend leads its own process group. The managed parent always uses `start_new_session` (armed), an interactive foreground launch owns its group via job control (armed, safe) — but a wrapper script or a `Popen` without `start_new_session` shares the *launcher's* group, and sweeping that killed unrelated processes. The reaper also inherits stderr into the per-child log and prints before sweeping, so a vanished process tree leaves evidence. The thread reaper stays: it covers the one case the sibling can't (parent gone while the frontend hangs in graceful shutdown), and it already had the leadership guard.
- **per_call boot protocol cleanups.** `supports_offload` is rejected on the **class, before construction** — a misconfigured scorer no longer loads 16 GiB onto the exact GPU the protocol protects just to be told no. The "came up resident despite --boot-offloaded" branch was unreachable (an old service tree exits at argparse before binding the socket) and is folded into the exit-code error as a log hint.
- **EditReward columns named by config semantics.** With in-model pooling + `output_dim: 2` + the uncertainty loss, `reward()` returns `[pooled reward, pooled log-sigma]`, so the scorer now derives instance `sub_metric_names = ("reward", "log_sigma")` from the loaded config instead of labeling the uncertainty channel `edit_quality`; per-head configs keep the semantic names. Blast radius: only configs that pin the uncertainty head — in-tree that's exactly the new managed recipe, so no existing dashboard keys change. The single-item fallback now shapes through the same `_shape_reward` as the batched path, and your docs note (`8ed59d63`, thanks — merged) was reworded to match the naming behavior.

Verified with disposable harnesses against the real `create_direct_app`/`compute_rewards`/`_shape_reward` (run + results in the PR1/PR2 discussion; not committed per repo policy): reaper skipped when not leader / armed when leader; boot-offload rejection fires pre-construction; pooled-uncertainty naming and column-0 value unchanged; plus re-runs of the earlier chunk-eviction / event-loop-stall / exception-masking checks — all green. Known leftovers deliberately not touched: the `per_call` path still has no in-tree consumer (recipe pins `resident`), and the ruff reformat inside `unirl-reward-service/` (a lint-excluded vendored tree) is entangled with real edits — flagging both for you rather than churning further.

unirl/reward/managed_process.py
haonan3: Fixed by c7680247 (child boots with --boot-offloaded; EditReward honors UNIRL_SCORER_BOOT_OFFLOADED and constructs on CPU, so the GPU is never touched during bootstrap) with follow-ups in 662b4a2f: the supports_offload rejection moved to the class, before construction, and the unreachable old-server detection branch folded into the exit-code error. Resolving.

unirl-reward-service/reward_service/server.py
haonan3: Fixed by e40bdbf4: the gateway rejects a scorer_version pin with an explicit 400 pointing at the managed single-scorer path — it has no truthful per-scorer version to echo, so failing loud beats echoing None and failing every response downstream. The direct managed server keeps echoing its real version. Resolving.

unirl/reward/managed_process.py
haonan3: Fixed by d86ec91f: an offload failure is tolerated only while a scoring exception is already unwinding (the original error propagates; the lifecycle error is logged), and stays fatal after a successful score. Verified with a source-extracted harness: crash-mid-score now surfaces the scorer RuntimeError, not the transport error. Resolving.

unirl/reward/managed_process.py
haonan3: Fixed by 391480bc: the cache holds at least one full in-flight chunk (cache_limit = max(cache_size, len(requests))), so a retry after a lost response is a pure cache hit. Measured: 16-row chunk against cache_size=4 recomputed 12 rows on retry before, 0 after. Resolving.

haonan3 (approved):
LGTM

PR #304, opened 2026-08-03
feat(bagel): add it2i vLLM-Omni managed reward recipe

nussejzz:
## Summary

- Add a Sample/Part-native BAGEL `bagel_it2i` vLLM-Omni adapter that preserves source-image alignment for every GRPO sibling.
- Build worker and replay contexts from the same aspect-preserving VAE/ViT transforms, deterministic VAE posterior mean, fixed-noise keys, and matched SDE cutoff.
- Rebuild source contexts differentiably for LoRA replay while preserving upstream navit inference-mode guards and copy-on-write KV storage.
- Add the first end-to-end managed reward consumer recipe: BAGEL FSDP8 + vLLM-Omni DP8/TP1 + rank-affine EditReward DP8, with safe rollout sleep by default and measured all-resident opt-in.

## Related Issue

N/A

## Test Plan

- `uvx --from pre-commit pre-commit run --files <all changed PR3 files>`
- `python -m compileall -q unirl/models/bagel unirl/rollout/engine/vllm_omni`
- `python lint/check_recipe_targets.py` (`2,458` recipe `_target_` paths resolved on the top-stack tree).
- Hydra compose:
  `python -m unirl.train_diffusion --config-name=diffusion/bagel/bagel_it2i_vllmomni --cfg job --resolve rollout_sleep_after_generate=false offload_train_during_reward=false`
- End-to-end latest-upstream smoke on 8x NVIDIA H20 96GB with `ByteDance-Seed/BAGEL-7B-MoT`, `TIGER-Lab/EditReward-MiMo-VL-7B-SFT-2508`, `XiaomiMiMo/MiMo-VL-7B-SFT-2508`, OpenGPT image-edit data, FSDP8, vLLM-Omni DP8/TP1, LoRA rank 64, 8 prompts x 8 samples, and two optimizer updates.
- Result: one complete rollout/reward/advantage/train step, reward `-1.0735`, ratio `1.0000±0.0000`, grad norm `0.0001`, peak HBM `88,430–88,538 MiB/card`, exit code 0, and zero residual GPU memory after teardown.
- The 8xH20 smoke ran on the pre-refresh tree; the refresh since merged current main and adjusted the recipe (retired eval key dropped, ac_wrap_order pinned) — see the refresh comment for what was re-verified.
- No standalone test files are added; validation is pre-commit, compile/compose checks, and the real DP8 smoke above.

## Compatibility / Risk

- The new modality and recipe are additive; existing BAGEL T2I behavior remains available.
- The shared context-helper refactor retains upstream's load-bearing eval-mode guard for navit packed inference.
- It2i intentionally disables packed rollout because each request carries a distinct source image.
- `bundle.enable_vit=true` is required for trainer-side source-context replay.
- The recipe defaults to rollout sleep; all-resident mode must be explicitly enabled only after measuring full geometry memory.
- No checkpoint format change; conditions add raw source-image material required to rebuild cross-process replay contexts.

## Reviewer Notes

- Stacked on #302 and #303 and refreshed onto current main. Review the PR3 slice: `unirl/models/bagel/*`, `unirl/rollout/engine/vllm_omni/*`, `unirl/algorithms/bagel_flow_unigrpo.py`, and the recipe — `git diff <PR2-head>..HEAD`. Must not merge before both dependencies.
- The two upstream BAGEL conflicts were resolved semantically: upstream's eval-mode guard and this PR's differentiable source-context rebuild are both retained.

## Checklist

### DIFF
--- a/examples/diffusion/bagel/bagel_it2i_vllmomni.yaml
+++ b/examples/diffusion/bagel/bagel_it2i_vllmomni.yaml
@@ -1,0 +1,282 @@
+# @package _global_
+# BAGEL-7B-MoT image-EDIT GRPO — vLLM-Omni + managed EditReward DP8.
+#
+# The vllm_omni-rollout sibling of examples/diffusion/bagel/bagel_editreward.yaml:
+# identical model / precision / algorithm / data / reward, but generation runs in
+# eight colocated vLLM-Omni worker subprocesses (one per GPU, TP=1 → DP=8) instead
+# of the in-process TrainsideRolloutEngine, and the freshly-trained LoRA is pushed
+# into them each rollout. Train and one environment-isolated EditReward child per
+# rank remain resident; rollout residency is an explicit memory-policy choice.
+#
+# What it2i adds over bagel_vllmomni.yaml (all handled in unirl/{models,rollout}/bagel):
+#   - modality bagel_it2i — the adapter ships each sample's SOURCE image on the
+#     request (multi_modal_data["image"]) AND mirrors it onto the deferred
+#     conditions (input_images), because the trainer rebuilds the KV contexts at
+#     replay and must prefill the same source image the worker did.
+#   - RLBagelPipeline builds the three it2i KV contexts inside the worker with the
+#     vendored ImageTransform pair and injects them through vLLM-Omni's KV channel,
+#     so upstream's own img2img prefill never runs. That prefill would (a) override
+#     the output canvas with the resized SOURCE dims — breaking the driver-authored
+#     x_T shape — and (b) squash the source to a fixed 980x980 SigLIP square (~4x the
+#     ViT tokens, aspect ratio dropped) instead of the aspect-preserving navit
+#     transform the trainer replays with. A conditioning mismatch is NOT something
+#     replay can absorb: the reward would be earned under one conditioning while the
+#     gradient is taken under another, and no importance ratio corrects that.
+#   - bundle.enable_vit: true — the und ViT tower, needed on the TRAINER to rebuild
+#     the contexts at replay (the worker always builds its own).
+#   - packed rollout is off for it2i (upstream's grouped generate_image is cfg=1 t2i
+#     only), so every GRPO sibling is its own worker request.
+#
+# Deploy (one rank-affine child per train worker; no extra Ray GPU slot):
+#   export BAGEL_PATH=/path/to/BAGEL-7B-MoT
+#   export EDITREWARD_PYTHON=/path/to/editreward/venv/bin/python
+#   export EDITREWARD_CHECKPOINT=/path/to/EditReward-checkpoint
+#   export EDITREWARD_BASE_MODEL=/path/to/MiMo-VL-base
+#   export EDITREWARD_SERVICE_ROOT=$PWD/unirl-reward-service
+#   bash examples/run_experiment_single_node.sh diffusion/bagel/bagel_it2i_vllmomni
+#
+# Compose check (CPU):  python -m unirl.train_diffusion \
+#   --config-name diffusion/bagel/bagel_it2i_vllmomni --cfg job --resolve
+# Smoke first — it2i colocate is the tightest memory shape in the family (7B MoT +
+# ViT on the trainer AND a full BAGEL in each worker, plus source-image KV):
+#   override batch_size=8 num_rollouts=2 eval_interval=2 sampling.samples_per_prompt=2
+
+num_devices: 8
+batch_size: 8                  # prompts/rollout (x samples_per_prompt = images/rollout)
+adv_use_global_std: false      # per-group GRPO normalization
+num_rollouts: 10000            # intentionally large; stop manually
+
+# Periodic eval on the eval set (deterministic: eval_eta=0), logged under eval/*.
+# Eval prompts are chunked (eval_chunk_prompts) to stay within the it2i driver-memory
+# budget. Same values as the trainside editreward run.
+eval_interval: 25
+eval_num_prompts: 16
+eval_samples_per_prompt: 4
+eval_chunk_prompts: 8
+eval_cfg_text_scale: 1
+eval_eta: 0.0
+
+transport_kind: colocate_store
+workers_per_device: 1
+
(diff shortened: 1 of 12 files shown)


nussejzz:
Clipboard_Screenshot_1785728659

examples/diffusion/bagel/bagel_it2i_vllmomni.yaml
```diff
@@ -0,0 +1,282 @@
...
+          rm_head_type: ranknet_multi_head
+      client:
+        _target_: unirl.reward.remote.RemoteRewardSpec
+        base_url: managed://rank-affine
+        required_rewards: [editreward]
+        reward_weights: {editreward: 1.0}
+        batch_size: 8
+        request_batch_size: 8
+        timeout: 600.0
+        sub_metric_reduce: mean
```
CjhHa1: **[P1] Do not average EditReward uncertainty into the training reward**

The selected EditReward config uses `pooling_strategy: mean`, `output_dim: 2`, and `loss_type: uncertainty`. The two semantic heads are already pooled inside the model, so the returned columns are the aggregated reward mean and log-uncertainty, not separate instruction-following and visual-quality scores. `EditRewardScorer` currently labels those columns as `edit_following` / `edit_quality`, and this `sub_metric_reduce: mean` then averages the actual reward with its uncertainty parameter. The official EditReward inference path consumes only `reward[0]`.

This changes the optimization target for the new recipe. Please expose/consume only the score channel (or at minimum use the first channel here) and keep uncertainty as diagnostic metadata rather than a reward sub-metric.

unirl/models/bagel/diffusion.py
```diff
@@ -584,7 +645,10 @@ def build_forward_kwargs(
         :class:`~unirl.algorithms.bagel_flow_unigrpo.BagelFlowUniGRPO`'s velocity-MSE
         loop, which evaluates ``v_theta`` / ``v_ref`` across the SDE window.
         """
-        gen, cfg_text, cfg_img, image_shape = self._resolve_single(conditions)
+        gen, cfg_text, cfg_img, image_shape = self._resolve_single(
+            conditions,
+            differentiable=torch.is_grad_enabled(),
```
CjhHa1, line 487: **[P1] Build the UniGRPO reference context under reference weights**

For image-conditioned `BagelFlowUniGRPO`, `build_forward_kwargs()` is called with grad enabled before the algorithm enters `_disable_lora()` / `_reference_weights()`. This makes `_resolve_single()` build the source-image KV cache with the live policy; that prefill executes trainable `*_moe_gen` projections. The same `forward_kwargs` is then reused for `v_ref`, so the reference velocity uses base denoising weights but policy-produced conditioning K/V.

When `mse_weight > 0`, this is not a frozen-base reference objective. Please build separate reference `forward_kwargs` inside the disabled-LoRA/reference-weight context (and under `no_grad`), then build policy contexts separately for `v_theta`. The new FlowGRPO recipe does not trigger this, but the it2i stage now exposes the bug to the supported UniGRPO path.

haonan3:
Maintainer refresh: `6b88d24e`/`45a18f8c` carry current main + the PR1/PR2 resolutions through the stack; `676bad2f` fixes two things found while re-verifying against current main:

- **`force_rebuild` is now gated on `input_images`.** Rebuilding the reference conditioning is only sound where raw material can reproduce the stored context — exactly the image-conditioned case the flag exists for (it2i's source prefill runs the trainable `*_moe_gen` projections). Prompt-only conditioning runs the frozen und path AND may bake in text the bare prompt cannot reproduce: t2ti stores `init + system + think + prompt`, so the unconditional rebuild in `862f7f84` silently reconditioned `v_ref` on a bare-prompt context for the one shipped recipe that actually runs `BagelFlowUniGRPO` with `mse_weight>0` (`bagel_trainside_unigrpo`, t2ti). `_resolve_single` now enforces its docstring's promise and refuses a forced rebuild without raw material. (Full-FT caveat noted in the comment: prompt-only stored contexts remain the pre-existing approximation there.)
- **Recipe vs current main.** `eval_cfg_text_scale: 1` is a retired key on main (#346) and would fail at startup after this merge — dropped; the eval overlay inherits the training `cfg_text_scale: 1`, so nothing changes semantically. `ac_wrap_order: inside` is pinned — the composition this recipe's 8xH20 smoke actually validated, now that the knob defaults to the pre-existing `outside` (see PR1). The `sub_metric_reduce` comment follows the scorer's config-derived column names from PR2.

Verified: `python -m unirl.train_diffusion --config-name=diffusion/bagel/bagel_it2i_vllmomni --cfg job --resolve` composes clean under the full import chain (no retired keys in the resolved config), full pre-commit green on the tree. Not covered: a GPU smoke of it2i×UniGRPO — that combination still has no recipe, which is exactly why the gate fails closed.

examples/diffusion/bagel/bagel_it2i_vllmomni.yaml
haonan3: Fixed in two layers: 419b35c4 switched the recipe to sub_metric_reduce: first (verified end to end: first -> -1.0735 score channel, mean -> -0.0867 midpoint of score and log-sigma), and 662b4a2f makes the scorer derive its sub-metric names from the loaded config — under the pinned pooled-uncertainty config the columns are now named (reward, log_sigma), so the uncertainty channel can no longer masquerade as a quality score anywhere downstream. Resolving.

unirl/models/bagel/diffusion.py
haonan3: Fixed by 862f7f84 (reference conditioning rebuilt inside the reference-weight/no-grad scope via force_rebuild, bypassing the stored-context shortcut) plus 676bad2f, which scopes the rebuild to what raw material can actually reproduce: image-conditioned it2i (the trainable-*_moe_gen case this thread is about) force-rebuilds; prompt-only t2ti keeps its stored frozen-und contexts, because a bare-prompt rebuild would silently drop the baked-in system+think text — _resolve_single now refuses a forced rebuild without input_images. Resolving.

haonan3 (approved):
LGTM

PR #306, opened 2026-08-03
feat(qwen3-omni): support standalone audio modality training

cboss6:
## Summary

This PR adds end-to-end standalone audio support for Qwen3-Omni without
changing the existing video and audio-in-video data contract.

- Accept `(audio, prompt)` media references and keep prompt media sparse in
  sample metadata.
- Render standalone audio as an audio content block for both vLLM rollout and
  Hugging Face train-side replay.
- Decode audio files with PyAV and forward audio features without requiring
  video frames.
- Track `use_audio_in_video` per sample so standalone audio and video samples
  can safely share a batch.
- Compute Qwen3-Omni TMRoPE positions per sample instead of applying one
  batch-wide modality flag.
- Add a self-contained DCASE 2025 Audio QA recipe, dataset converter,
  documentation, and focused tests.

### Standalone audio flow

```mermaid
flowchart LR
    A["UniRL JSONL<br/>MediaRef(audio, prompt)"] --> B["MultimodalRLDataSource"]
    B --> C["Sparse prompt-media metadata"]
    C --> D["build_omni_messages"]
    D --> E["Qwen3-Omni audio content block"]

    E --> F["vLLM-Omni adapter"]
    F --> G["Audio waveform in<br/>multi_modal_data.audio"]
    G --> H["Policy rollout"]

    E --> I["Train-side chat template"]
    I --> J["PyAV audio decode"]
    J --> K["Processor audio features<br/>and attention mask"]
    K --> L["HF policy replay"]

    H --> M["Reward and advantage"]
    L --> N["Per-sample TMRoPE<br/>use_audio_in_video=false"]
    M --> O["GSPO update"]
    N --> O
```

Prompt media references are stored in root metadata rather than dense
primitive parts. This keeps heterogeneous text, standalone-audio, and video
rows representable in the same batch while media is loaded only when the
rollout or replay path needs it.

### Standalone audio versus audio-in-video

```mermaid
flowchart TD
    A["Input sample"] --> B{"Media source"}

    B -->|"Audio file"| C["Standalone audio"]
    C --> D["Decode one audio stream"]
    D --> E["Audio placeholders and features"]
    E --> F["use_audio_in_video = false"]
    F --> G["Audio-only TMRoPE positions"]

    B -->|"Video file with audio track"| H["Audio-in-video"]
    H --> I["Decode video frames and audio track"]
    I --> J["Video and audio placeholders/features"]
    J --> K["use_audio_in_video = true"]
    K --> L["Joint temporal audio-video positions"]
```

The two paths can produce similar audio feature tensors, but they are not
positionally equivalent:

- **Standalone audio** has no video grid and sets `use_audio_in_video` to
  `false`.
- **Audio-in-video** carries video frames and `video_grid_thw`, and sets
  `use_audio_in_video` to `true` so Qwen3-Omni aligns audio with the video's
  temporal grid.
- The flag is recorded per sample and replayed per row. A mixed batch therefore
  cannot accidentally treat every audio input as video-associated audio.

### Data and recipe

The new DCASE converter:

- reads the Hugging Face parquet shards;
- extracts embedded audio into WAV files using atomic writes;
- normalizes valid answers to `A`, `B`, `C`, or `D`;
- emits UniRL prompt-first JSONL with `(audio, prompt)` media references; and
- rejects malformed choices, missing answers, and unsupported records.

The audio recipe is fully self-contained: all training, rollout, model, and
data-source parameters are visible in one YAML file and can be edited without
following Hydra inheritance.

### Training recipe

The new recipe is
`examples/ar/qwen3_omni_audio_dcase_gspo_lora_vllm_omni_1x4.yaml`. It is
intentionally self-contained: model, FSDP, optimizer, rollout, weight sync,
reward, GSPO, sampling, evaluation, and data-source parameters are all visible
in one file instead of being inherited from a video recipe.

```bash
DATA_PATH=datasets/dcase2025_audio_qa/train.jsonl \
EVAL_DATA_PATH=datasets/dcase2025_audio_qa/val.jsonl \
PYTHONPATH=$(pwd) \
python -m unirl.train_ar \
  --config-name=ar/qwen3_omni_audio_dcase_gspo_lora_vllm_omni_1x4
```

Key settings and their rationale:

| Area | Setting | Purpose |
| --- | --- | --- |
| Parallelism | 4 GPUs, FSDP training, vLLM-Omni TP=4 rollout | Reuses all four GPUs between policy optimization and rollout; sleep mode releases rollout memory while training runs. |
| Rollout batch | 8 prompts × 8 samples per prompt | Produces 64 trajectories per rollout and gives GSPO enough siblings for group-relative advantages. |
| Optimization | 500 rollouts, one update per batch, LR `1e-5`, max grad norm `1.0` | Keeps the update schedule conservative for LoRA RL fine-tuning. |
| Trainable weights | LoRA rank 32 / alpha 64 on `q/k/v/o_proj` | Updates the Thinker attention projections while keeping the base model compact and reviewable. |
| Frozen encoders | Vision tower and audio tower frozen | Trains reasoning over existing audio representations rather than changing the pretrained audio encoder. |
| Sequence limits | Prompt 16,384 tokens; generation 2,048 tokens | Leaves room for audio tokens and chain-of-thought while bounding rollout cost. |
| Standalone-audio switch | `use_audio_in_video: false` in bundle, pipeline, and rollout | Ensures audio receives audio-only position handling and does not require a video grid. |
| Reward | Exact A-D MCQA, `require_answer_phrase: true` | Scores semantic correctness only when the response contains a recognized answer phrase. |
| Evaluation | Every 5 rollouts, full validation set, one deterministic sample per prompt | Produces an `avg@1` curve that is comparable across checkpoints without sampling noise. |

### Training curves and observed results
Clipboard_Screenshot_1785745499

| Metric | Observation |
| --- | --- |
| Validation accuracy | `57.33%` at baseline, `64.67% at rollout 195 (`+7.34  pp` over baseline) |
| Training reward | Overall `58.11% ± 10.90 pp` per-step SD; first 50 rollouts `54.94%`, latest 50 `60.97%` (`+6.03 pp`) |
| Reward trend | OLS slope `+2.53 pp / 100 rollouts`, two-sided `p=0.031` |
| Gradient norm | Mean `0.160`, P95 `0.211`, maximum `0.279`, below the `1.0` clipping threshold |
| K3 divergence | Mean `0.00618`, maximum `0.0235`; no divergence or NaN signal |
| Runtime health | No OOM, NaN, or traceback marker in the parsed log |

The fixed validation curve provides the clearest evidence of improvement.
`rollout/reward_mean` is noisier because every rollout samples a different
training batch, so it combines policy quality with prompt difficulty. In the
longer snapshot, both the latest-50 training reward and the last-five fixed-set
evaluation mean are higher than their early-run counterparts. The raw reward
trend is now positive at `+2.53 pp / 100 rollouts` with nominal `p=0.031`.
Because rollout batches are not a fixed evaluation population and adjacent
steps are not independent, that p-value is descriptive rather than a formal
generalization claim.

### Scope

This PR intentionally does not enable standalone image input. Image model
plumbing, rollout limits, Video-R1 conversion, and the image training recipe
remain in the follow-up image PR.

## Test Plan
Not run; Reason: no need.

## Checklist


haonan3 (dismissed):
Thanks for pushing the standalone-audio path end to end. The per-sample audio-in-video flag and caching the processor outputs for replay are good directions, but I think the following need to be fixed before merge.

1. Please keep media in the typed Sample/Part contract. This PR changes the existing video/prompt path from a Videos input Part into root metadata["_media_refs"], then adds audio through the same private channel. build_omni_messages also accepts typed Audios/Videos, so this leaves two live representations and two ingestion paths for the same model input. Metadata is currently dataset/reward context; hiding model inputs there bypasses Sample.turns(), primitive validation, slicing/transport, and general tooling. The new standalone-audio recipe is homogeneous, so it can use a typed Audios (and the existing typed Videos) Part now. If heterogeneous sparse modalities are required, that should be designed once as a first-class framework contract in a separate PR, not as a Qwen3-specific metadata side channel.

2. The new audio recipe selects SDPA, but #274 established that Qwen3-Omni audio cu_seqlens are only honored by the FlashAttention-2 varlen path. SDPA/eager performs full attention instead of the intended segmented attention, so this is a semantic rollout/replay mismatch rather than normal kernel noise. Please switch the standalone-audio trainside model to flash_attention_2, or provide standalone-audio layer-wise evidence that its exact cu_seqlens layout makes SDPA equivalent. A low end-to-end K3 alone is not enough to establish that.

3. The per-row TMRoPE position_ids added in Qwen3OmniARStage.autoregress do not currently survive the pinned Transformers 5.6 generation helper. Qwen3OmniMoeThinkerForConditionalGeneration.prepare_inputs_for_generation resets model_inputs["position_ids"] to None, after which forward recomputes positions with the batch-wide use_audio_in_video flag. Replay uses the explicit per-row positions, so a mixed audio-in-video/non-audio-in-video trainside batch can still disagree between generation and replay. Please make the precomputed positions reach the first forward and validate the mixed matrix; otherwise remove the heterogeneous-safety claim.

4. Please canonicalize standalone audio decoding. The sparse URI branch decodes/resamples to the processor sample rate, while the typed Audios branch accepts a waveform without sample-rate metadata or resampling. Since both are accepted by build_omni_messages, the same 48 kHz source can produce different model inputs depending on which route a caller happens to use. One typed input contract should own decoding/resampling and feed both rollout and replay.

5. The shared thinker stage currently carries max_num_seqs: 1 because audio-in-video batching was numerically unstable. Reusing it for every standalone mode silently serializes audio/image/video rollout. Please split the audio-in-video workaround into a modality-specific stage config, or provide evidence that standalone audio also requires serialization.

Please also make CI green and fill the Test Plan with exact commands/results. At minimum I would expect processor ids/features/masks, audio-tower layer-wise parity, TMRoPE positions, rollout/replay K3, and 16 kHz/48 kHz + stereo/long-audio cases. The current PR is merge-conflicted, lint is failing, and PR-body validation is failing.

cboss6:
@haonan3 Thanks for the detailed review. I’ve addressed the blocking points as follows.

1. Keep media in the typed Sample/Part contract
Agreed — the metadata["_media_refs"] side channel is removed.

Prompt media now goes through a first-class sparse primitive:

MediaRef / MediaRefs under Part.primitives["media"]
exposed via Sample.turns() / Sample.prompt_media_refs()
data-source collation rejects _media_refs in metadata
build_omni_messages takes URI-backed MediaRefs as the canonical path (with a temporary deprecated URI-Videos.from_uris shim). Waveform Audios and decoded frame Images/Videos are rejected for Omni prompts so we don’t keep two live ingestion paths.

This also covers heterogeneous sparse rows (different image/audio/video mixes in one batch) without a Qwen3-only metadata bypass. Omni-specific cardinality stays in the Omni adapter; the global data layer only does schema/URI normalization. ARTrainer only allows "media" when the pipeline target is qwen3_omni.

2. SDPA → FlashAttention-2
Standalone-audio recipe now uses attn_implementation: flash_attention_2 (same as the audio-video vLLM-Omni recipes), so audio cu_seqlens stay on the FA2 varlen path from #274.

3. Per-row TMRoPE / mixed audio-in-video
You’re right that this isn’t a simple “keep position_ids alive” fix, and Qwen currently doesn’t support mixing video-only and audio-in-video in one batch.

We no longer claim heterogeneous safety for that mix. Instead:

_validate_audio_in_video_modes fail-loud rejects mixed video-only / audio-in-video rows
TODO notes that true mixed training needs a separate end-to-end design later
Uniform video-only or uniform audio-in-video batches remain supported.

4. Canonicalize standalone audio decoding
Single URI contract owns decode/resample for both rollout and replay:
prompt audio must be URI-backed MediaRef
decode once via UniRL load_audio_pyav → mono float32 at processor rate
HF gets the waveform; vLLM gets (waveform, sample_rate) so it doesn’t re-decode/resample
waveform Audios without a sample-rate contract are rejected

5. Split max_num_seqs workaround
Shared thinker stage (qwen3_omni_thinker_only_rl_1x4.yaml): max_num_seqs: 64 for standalone image/audio/video
Dedicated audio-in-video stage (qwen3_omni_thinker_only_rl_audio_video_1x4.yaml): max_num_seqs: 1, with a comment on the batch-sensitive K3 issue
Audio-video recipes point at the dedicated stage

leviking98z-rgb (approved):

PR #308, opened 2026-08-03
refactor(fastvideo): integrate upstream through runtime patches

Zcchill:
## Summary

Refactor the FastVideo rollout integration to run against a pinned, unmodified upstream FastVideo revision and install UniRL-specific behavior through fail-closed runtime patches.

This removes the runtime dependency on a separately maintained FastVideo fork and fixes the spawned-worker TCPStore failure path that previously caused otherwise healthy long-running jobs to terminate after repeated sleep/wake cycles.

## Motivation

Maintaining UniRL-specific behavior in a separate FastVideo repository made the effective runtime version difficult to identify and complicated reproduction, upgrades, and support for additional model families.

This change moves the integration boundary into UniRL:

- FastVideo is installed from a pinned upstream commit;
- UniRL validates the installed version, symbols, and signatures;
- runtime patches are installed idempotently in both parent and spawned workers;
- model-specific behavior is isolated behind adapters;
- unsupported versions and contracts fail closed.

### Runtime patch suite

The patch installer provides:

- RLData contract extensions for SDE metadata;
- custom sigma and timestep handling;
- Flow and Dance SDE transition math and SDE-index windows;
- trajectory, timestep, log-prob, and text-condition transport;
- strict full-parameter checkpoint hot-swaps;
- spawned-worker patch propagation;
- TCPStore address-collision recovery.

Compatibility and capability checks run before the FastVideo backend is exposed.

### Adapter and backend separation

The rollout engine is split into:

- a generic FastVideo backend responsible for lifecycle and execution;
- model adapters responsible for request construction, schedules, conditions, outputs, and responses;
- a Wan2.1 adapter containing the current Wan-specific assumptions.

This prevents Wan-specific scheduler, shape, and conditioning behavior from leaking into future FastVideo model integrations.

### Dependency model

The `fastvideo` optional dependency pins:

`hao-ai-lab/FastVideo@2095477eac7e289c7a7ab13acb367ca60687c304`

`FASTVIDEO_PATH` remains available as a development override. No FastVideo source file or installed `site-packages` file is modified on disk.

### Root cause

FastVideo recreates multiprocessing workers and their distributed process group after colocated sleep/wake cycles. Worker startup first selects an apparently free TCPStore port, but another process can claim it before the spawned worker binds the socket.

This time-of-check/time-of-use race produces `EADDRINUSE` in a child process. The original child failure path signals the parent process, causing the Ray rollout actor and training job to exit. Retrying only at the outer engine layer cannot reliably recover an exception originating asynchronously in the spawned worker.

### Fix

The multiprocessing runtime patch now:

- propagates UniRL patches into spawned workers;
- reports worker initialization failures through the ready pipe instead of terminating the parent;
- detects TCPStore address-collision failures;
- closes failed worker pipes and processes;
- reserves a new master port and retries worker startup;
- raises the original error when bounded recovery is exhausted.

The backend also retries initial generator boot with a newly reserved port.

### Validation

Formal long-running experiments encountered real TCPStore address collisions during repeated sleep/wake cycles. The patched workers recovered without terminating Ray or restarting the training jobs, and all four jobs continued completing later rollout and optimizer steps.

## Supported scope

This draft currently validates:

- Wan2.1 T2V 1.3B;
- Flow and Dance SDE strategies;
- replay old-log-prob mode;
- native rollout old-log-prob mode;
- single-node 8-GPU colocated rollout and training;
- full checkpoint-path weight synchronization.

Other model families require a dedicated adapter and are rejected instead of inheriting Wan-specific behavior.

## Validation

- 7 focused runtime-patch tests passed;
- all changed Python files passed compilation;
- Ruff check and format verification passed;
- 1,455 recipe target paths resolved;
- replay and native modes completed rollout, reward, backward, optimizer update, and 8/8 full-weight synchronization.

Formal 720p runs use `batch_size=48`, `samples_per_prompt=16`, 768 samples per rollout, two updates per batch, `forward_batch_size=4`, 14 denoising steps, and shift 5:

Clipboard_Screenshot_1787044556

Clipboard_Screenshot_1787044624

Clipboard_Screenshot_1785751943

Clipboard_Screenshot_1785751965

## Draft status

This PR is opened as a draft for architecture and compatibility review.

Before marking it ready:

- finalize the supported FastVideo compatibility contract;
- review the patch surface for future non-Wan adapters;
- consolidate reproduction instructions and final validation evidence.

### DIFF
--- a/examples/diffusion/wan21/wan21_t2v_dancegrpo_fastvideo.yaml
+++ b/examples/diffusion/wan21/wan21_t2v_dancegrpo_fastvideo.yaml
@@ -17,9 +17,9 @@
 #      weights and pushes the merged full model (FastVideo serves merged weights;
 #      it has no separate adapter path here).
 #
-# FastVideo dependency: https://github.com/Zcchill/FastVideo at or after
-# 7fe1d7db9a0b8aebb46679e7924f597431f23665. Set $FASTVIDEO_PATH to that
-# checkout; FastVideo is lazy-imported by the rollout engine only.
+# FastVideo dependency: install UniRL's `fastvideo` extra. pyproject.toml pins
+# the unmodified hao-ai-lab/FastVideo source snapshot validated by UniRL's
+# fail-closed runtime patches. FASTVIDEO_PATH remains a development override.
 #
 # This recipe defaults to replay log-probs. Native mode is enabled by setting
 # rollout.config.native_logprob=true and algorithm.old_logp_source=rollout;
--- a/examples/diffusion/wan21/wan21_t2v_fastvideo_replay.yaml
+++ b/examples/diffusion/wan21/wan21_t2v_fastvideo_replay.yaml
@@ -1,0 +1,176 @@
+# @package _global_
+# wan21_t2v.yaml (FlowGRPO trainside baseline) with ONLY the rollout engine
+# swapped trainside -> FastVideo VideoGenerator, in REPLAY log-prob mode.
+#
+# Purpose: head-to-head against the wan21_t2v.yaml trainside baseline under
+# IDENTICAL hyperparams (FlowSDEStrategy, shift=5, 720p/5-frame, video_pickscore,
+# LoRA targets, sampling schedule, num_updates=2). The intended experiment
+# isolates the rollout-engine effect. Replay mode recomputes both old and new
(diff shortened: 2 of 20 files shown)


Zcchill:
Superseded by #415. Closing this draft.

PR #309, opened 2026-08-03
feat(qwen3-omni): support standalone image modality training

cboss6:
## Related PR
Please merge below PR before merging current one:
[feat(qwen3-omni): support standalone audio modality training](https://github.com/Tencent-Hunyuan/UniRL/pull/306)

## Summary

This PR enables standalone-image RL training for Qwen3-Omni Thinker on top of
the current typed `Sample -> Sample` rollout contract, while preserving text,
standalone-audio, video, and audio-in-video inputs.

- Extend the Video-R1 converter with `--modality image|video|all` and emit
  `(image, prompt)` as a typed `MediaRef`.
- Enable one image per request in the shared Qwen3-Omni vLLM stage.
- Add a self-contained Video-R1 image GSPO + LoRA recipe.
- Exercise the existing typed sparse-media path end to end: `MediaRefs` input
  primitive, role-aware conversation rendering, shared media preparation,
  vLLM rollout, stored replay conditions, image feature fusion, and
  per-sample TMRoPE.

### Current standalone image architecture

```mermaid
flowchart TB
    A["Video-R1 JSONL<br/>media_refs: image / prompt / URI"] --> B["TextPromptDataset<br/>normalize_prompt_example"]
    B --> C["MediaRef + MediaRefs<br/>typed sparse Batch"]
    C --> D["MultimodalRLDataSource"]
    D --> E["Sample request lineage<br/>text input Part -> media input Part"]
    E --> F["Fork empty AR generation Part"]
    F --> G["Sample.turns()<br/>frontier-aligned role-aware turns"]
    G --> H["build_omni_messages<br/>image block before text block"]
    H --> I["prepare_omni_media<br/>URI -> one RGB PIL image"]
    I --> J["HF Qwen3-Omni processor<br/>expanded input IDs + pixel_values + image_grid_thw"]

    J --> K["Compress expanded image placeholders<br/>for vLLM re-expansion"]
    I --> L["multi_modal_data.image<br/>same decoded PIL image"]
    K --> M["vLLM-Omni GenerateCall"]
    L --> M
    M --> N["Policy rollout<br/>tokens + behavior log-probs"]

    J --> O["Cache exact expanded encoding"]
    N --> P["Output adapter fills AR Part<br/>TextSegment + decoded text + conditions"]
    O --> P
    P --> Q["Reward + group advantages"]
    Q --> R["GSPO re-types stored<br/>Qwen3OmniARConditions"]
    R --> S["Qwen3OmniARStage.replay<br/>image fusion + per-row TMRoPE"]
    S --> T["LoRA policy update"]
```

Prompt media is no longer copied through `metadata["_media_refs"]`. Dataset and
reward metadata stays in `Part.metadata`; model input media is represented by
`MediaRefs`, a `FieldKind.CONCAT` batch stored under
`Part.primitives["media"]`. The outer list remains batch-aligned while each row
contains only the URI references present for that sample, so text-only, image,
audio, and video rows can be selected, sliced, fanned out, and concatenated
without introducing a second model-input channel.

This PR covers the **prompt-image** channel used for visual question answering.
It is distinct from `(image, condition)`, where the data source decodes the
image on the driver into an `Images` tensor under
`Part.primitives["image"]` for diffusion, image-edit, Qwen-VL, and Qwen3.5
consumers. Qwen3-Omni prompt images intentionally use URI-backed `MediaRefs`
and reject decoded `Images` inputs.

For this recipe, URI materialization is performed once in the vLLM input
adapter. `prepare_omni_media` decodes the image and the Hugging Face processor
produces the expanded prompt IDs, `pixel_values`, and `image_grid_thw`. The
adapter sends compressed placeholder IDs plus the same decoded PIL image to
vLLM, while the output adapter stores the processor-derived replay fields on
the generated `Part.conditions`. GSPO replay consumes those stored conditions
directly; it does not reopen the URI or rebuild the chat template.

The native `Qwen3OmniPipeline` uses the same `Sample.turns() ->
build_omni_messages -> prepare_omni_media` path when generation is performed
train-side, and likewise writes its exact processor conditions to the generated
Part.

### Standalone image versus video

```mermaid
flowchart TD
    A["MediaRefs row"] --> B{"modality"}

    B -->|"image"| C["load_image_rgb<br/>RGB PIL"]
    C --> D["processor: pixel_values<br/>+ image_grid_thw"]
    D --> E["get_image_features"]
    E --> F["image placeholder mask<br/>+ image DeepStack levels"]

    B -->|"video"| G["sample_video_frames_pyav<br/>frames + effective FPS"]
    G --> H["processor: pixel_values_videos<br/>+ video_grid_thw<br/>+ second-per-grid"]
    H --> I["get_video_features"]
    I --> J["video placeholder mask<br/>+ video DeepStack levels"]

    F --> K["_merge_visual_multiscale<br/>preserve placeholder order"]
    J --> K
    K --> L["masked_scatter into<br/>Thinker input embeddings"]

    D --> M["_per_sample_rope"]
    H --> M
    M --> N["row-specific spatial / temporal<br/>TMRoPE positions"]
    L --> O["Thinker forward / replay"]
    N --> O
```

The image and video paths share the conversation, processor, condition, and
Thinker replay infrastructure while keeping modality-specific payloads:

- **Standalone image** carries `pixel_values` and `image_grid_thw`; it has no
  frame sampling or video time grid.
- **Video** carries `pixel_values_videos`, `video_grid_thw`, and temporal
  metadata derived from sampled frames.
- `Qwen3OmniARConditions` stores each media field as a per-sample CONCAT list
  with `None` for rows where that modality is absent.
- Replay concatenates only populated tensors for each tower, places image and
  video embeddings into their own placeholder slots, and merges DeepStack
  features in the original visual-token order.
- TMRoPE is calculated row by row from that row's image grid, video grid,
  seconds-per-grid, audio lengths, and audio-in-video flag. Image-only and
  video-only rows can therefore share one heterogeneous batch.

### Video-R1 data conversion

`datasets/video_r1_260k/convert_video_r1_260k_to_unirl.py` now exposes:

- `--modality image` for standalone-image rows;
- `--modality video` for the existing video path; and
- `--modality all` for a heterogeneous image/video dataset.

The converter keeps A-D multiple-choice samples, resolves media paths to
absolute URIs, validates the answer, skips missing media by default, and
reports retained samples by source and modality. Each source row contains
exactly one image or one video; Video-R1 does not attach both media types to the
same question.

```bash
python datasets/video_r1_260k/convert_video_r1_260k_to_unirl.py \
  --data-root /path/to/Video-R1-data \
  --out-dir datasets/video_r1_260k_image \
  --modality image \
  --sources Chart,General,Knowledge,Math,OCR,Spatial \
  --val-count 1000
```

### Training recipe

The new recipe is
`examples/ar/qwen3_omni_image_video_r1_gspo_lora_vllm_omni_1x4.yaml`. It is
self-contained so model, FSDP, optimizer, rollout, reward, GSPO, sampling,
evaluation, and data-source settings can be reviewed without following Hydra
inheritance.

```bash
DATA_PATH=datasets/video_r1_260k_image/train.jsonl \
EVAL_DATA_PATH=datasets/video_r1_260k_image/val.jsonl \
PYTHONPATH=$(pwd) \
python -m unirl.train_ar \
  --config-name=ar/qwen3_omni_image_video_r1_gspo_lora_vllm_omni_1x4
```

| Area | Setting | Purpose |
| --- | --- | --- |
| Parallelism | 4 GPUs, FSDP training, vLLM-Omni TP=4 rollout | Reuses four GPUs between optimization and rollout; sleep mode releases rollout memory during training. |
| Rollout batch | 8 prompts × 8 samples per prompt | Produces 64 trajectories for group-relative GSPO advantages. |
| Optimization | 500 rollouts, one update per batch, LR `1e-5`, max grad norm `1.0` | Uses conservative LoRA policy updates. |
| Trainable weights | LoRA rank 32 / alpha 64 on `q/k/v/o_proj` | Updates Thinker attention projections while keeping the visual and audio towers frozen. |
| Sequence limits | Prompt 16,384 tokens; generation 8,192 tokens | Leaves room for visual tokens and long-form reasoning while bounding rollout cost. |
| Image input | One image per prompt; `use_audio_in_video: false` | Activates the image processor path without decoding a video audio track. |
| GSPO clipping | Low `1e-3`, high `2e-3` | Applies the asymmetric clipping range used by the image recipe. |
| Reward | Exact A-D MCQA with `require_answer_tag: true` | Requires the converter's `<answer>X</answer>` format. |
| Evaluation | Every 5 rollouts, full validation set, one deterministic sample per prompt | Produces a fixed-set `avg@1` curve without sampling temperature noise. |

```mermaid
sequenceDiagram
    participant T as FSDP trainer
    participant S as LoRA weight sync
    participant A as Qwen3-Omni adapter
    participant V as vLLM-Omni TP=4
    participant R as MCQA reward

    T->>S: Publish current LoRA weights
    S->>V: Copy and verify adapter weights
    A->>A: Render turns, decode each image once, run HF processor
    A->>V: Compressed prompt IDs + PIL image, 8 samples per prompt
    V->>A: 64 token trajectories + behavior log-probs
    A->>A: Fill AR Part with TextSegment and cached replay conditions
    A->>R: Decoded responses + metadata.answer
    R->>T: Tagged A-D rewards and group-normalized advantages
    T->>T: Replay stored conditions and run one GSPO update
    Note over T,V: Repeat for 500 rollouts and evaluate every 5 rollouts
```

### Training curves and observed results

Clipboard_Screenshot_1786207931

The run provides end-to-end evidence that standalone images pass through data
loading, vLLM rollout, reward scoring, train-side replay, and repeated policy
updates without numerical instability. The evaluation curve finishes significantly above
its starting point.

The training reward remains broadly stable while the fixed-set evaluation
ends somewhat above baseline. This is consistent with a healthy training path;
the current evidence is most useful as validation of modality support and
optimization stability.

## Test Plan
Not run; reason: no need.

## Compatibility / Risk

- Prompt media must use URI-backed `MediaRef` / `MediaRefs`; the deprecated
  `metadata["_media_refs"]` channel is rejected. Decoded `Images`, `Videos`, and
  `Audios` are not accepted as Qwen3-Omni prompt media.
- Local paths and HTTP(S) image URLs are supported. `s3://` and `gs://` must be
  materialized before dataset loading.
- Qwen3-Omni currently accepts at most one prompt input per modality in a
  conversation, and the stage config limits image/audio/video to one each.
- Multimodal prompts that exceed `max_prompt_length` fail loudly instead of
  being silently truncated.
- The Video-R1 `--modality all` output may mix image-only and video-only rows in
  one batch, but every source row still contains exactly one visual input.

## Reviewer Notes

- The important rollout/replay alignment point is that vLLM receives compressed
  placeholders plus decoded media, while `Part.conditions` keeps the expanded
  Hugging Face processor encoding used by train-side replay.
- Suggested review order:
  `unirl/models/types/conversations.py`,
  `unirl/models/qwen3_omni/media.py`,
  `unirl/rollout/engine/vllm_omni/adapters/qwen3_omni.py`, and
  `unirl/models/qwen3_omni/ar.py`.

## Checklist


haonan3 (dismissed):
The Qwen3-Omni image fusion and condition fields look directionally aligned with the Transformers model, but I do not think this PR should merge in its current shape.

1. Please do not add image/prompt as root metadata["_media_refs"]. UniRL already normalizes image media into a typed Images input Part, and Qwen-VL consumes that route through Sample.turns() and build_vision_messages. The PR description says image/condition is diffusion-specific, but it is already used by the AR Qwen-VL/Geo3K path. build_omni_messages then accepts both typed Images and the private metadata route, producing exactly the dual implementation we should avoid. Reuse the typed Part contract for this homogeneous image recipe. If prompt-versus-condition semantics need to become first class, extend the generic media/Part contract explicitly rather than placing model input in metadata.

2. This branch is stacked on all commits from #306 but targets main, so the PR repeats the full standalone-audio implementation and inherits its blockers. Please land/fix #306 first and then rebase, or temporarily set the PR base to the #306 branch so this review can be image-only. In particular, the hidden media route and the trainside mixed-TMRoPE issue must not be duplicated into image support.

3. Please remove the unrelated changes to the existing Video-R1 recipe. Batch size 4 -> 8, eval interval 10 -> 5, evaluation set 100 -> all, temperature 1 -> 0, and path/default changes alter existing experiment behavior but are not required for standalone image input. Those changes should be justified and reviewed separately.

4. The shared stage YAML now enables image input, but it also contains the audio-in-video max_num_seqs: 1 workaround. This makes image/video-only jobs inherit audio-specific serialization. Split the stage presets by the behavior they actually need, rather than accumulating every modality and workaround in one global config.

5. Do not mark Qwen3-Omni as fully supported with a green check in README before this PR lands and the modality matrix is verified. The row also says Text/Image/Audio/Video/ -> Text; please remove the trailing slash and state clearly that this is multimodal input to text only, not speech/TTS output.

Please add a reproducible Test Plan and get PR-body validation green: image processor ids/features/DeepStack masks/TMRoPE parity against pinned Transformers, rollout/replay K3, and text-only/image/video request coverage.

cboss6:
@haonan3 thansk for detailed comments, I have updated my PR accordingly.

leviking98z-rgb (approved):

PR #315, opened 2026-08-04
refactor(agentic): split rollout engine from driver-side manager

celve:
## Summary

- make `AgenticRolloutEngine.generate(Sample) -> Sample` execute exactly one trajectory, with no queue, thread pool, group assembly, or trainer policy
- add a driver-local `RolloutManager` that owns bounded per-slot dispatch, autonomous progress during training, GRPO group assembly, blocking collection, quiescence, and policy-version synchronization
- express retention as a root-atomic `List[Sample] -> List[Sample]` filter, shared by incomplete-tail disposal and rollout-staleness handling, and migrate the barrier, partial, and fully async agentic trainers to the same manager

The implementation follows the reviewed [driver-side manager design](https://github.com/foshstl/LIN-693/blob/main/docs/design-engine-and-manager.md) and [implementation plan](https://github.com/foshstl/LIN-693/blob/main/plans/implementation.md).

## Related Issue

N/A (Linear LIN-693)

## Test Plan

- `git diff --name-only -z upstream/main...HEAD | xargs -0 uvx pre-commit run --files`
- `python3 -m compileall -q unirl`
- `python3 lint/check_recipe_targets.py` (2,452 recipe targets resolved)
- `python3 lint/check_experimental_boundaries.py`
- one-off, uncommitted CPU harness over fake handle slots: autonomous refill, per-slot caps, blocking collection, underflow restoration, carry/drop of incomplete roots, stale-root filtering, whole-root filter validation (including mutating filters), sync exclusion, explicit policy-version stamping across resumed turns, close idempotence, and infrastructure-error propagation all passed
- Not run; reason: Ray, PyTorch, the model checkpoint, datasets, and a GPU allocation are unavailable in this environment. Agentic ALFWorld/deep-research rollout and training smoke tests remain required.

## Compatibility / Risk

- `AgenticRolloutEngine` is now a synchronous single-trajectory engine. Its former coordinator methods and `AsyncAgenticRolloutEngine` are removed; all in-repo agentic callers are migrated.
- `rollout.config.per_worker_concurrency` and the dead `partial_rollout` flag are removed. `per_worker_inflight` is trainer/manager configuration. The six agentic recipes use `worker_max_concurrency: 10` for eight trajectory slots plus two control slots.
- `Part` gains optional `harness_status`. Newly generated agentic turns receive the manager's explicit policy version; carried turns keep their previous versions.
- The driver now owns a daemon progress thread and point-to-point `Handle.slot(...)` calls. The latter reuse the existing tensor localization/rebinding path, but need Ray/GPU validation under the configured transports.
- Checkpoint tensor formats are unchanged.

## Reviewer Notes

- Review the ownership boundary first: trainers own admission policy; `RolloutManager` owns scheduling and filtering; `AgenticRolloutEngine` owns one trajectory; the harness owns turns.
- PR #310 overlaps generic async policy-versioning and PR #318 owns rollout TP topology. This branch deliberately does not implement agentic TP topology; its `Slot` API is complementary, though both branches touch `Handle` and will require a rebase if either lands first.
- PR #314 adds the missing agentic reward-curve profiles. PR #317 is the superseded rank-zero/remote-manager direction and will be closed in favor of this PR.
- AI assistance: implementation, code audit, behavioral harness, docs, and PR text were produced with Codex. No generated test artifact is committed.

## Checklist


celve:
## Direction change — the first commit here is superseded

Design review moved the manager from a **driver-side plain object** to a **`Remote` entered at
rank 0**, owning allocation among the engine replicas it holds. That is closer to what the issue
actually asks for ("a separate scheduler layer" — a separate layer, not a driver-side one), and it
matches the prior art more honestly: slime's `RolloutManager` is a Ray actor, and neither slime nor
AReaL routes per-trajectory state through the controller.

What the revision buys:

- **Saturation is fixed structurally.** A manager with its own drive thread keeps placing work while
  the driver is inside the optimizer step. The driver-side version needed over-pushing into Ray's
  actor mailbox, which has no precedent in either reference system and which I could not verify.
- **Carried tails stop crossing the process boundary** — under `tail_policy: carry` the manager
  keeps and resubmits them. That also retires the must-test case of the old design (a carried
  partial placed on a different replica than produced it, needing cross-worker localize).
- **The pump collapses** into one blocking `collect()` RPC — no poll loop, no backoff, no
  `wait_any` on the driver.
- **Push vs. pull becomes an implementation detail** inside the manager, invisible to both neighbours.

What it costs, and why this PR shrinks:

**`@addressed` / `WorkerHandle` / `PendingWorkerCall` / `wait_any` are no longer needed.** They are
built on `Handle`, which only the driver has; a rank-0 manager holds raw actor handles and calls
workers the way `_fan` and `_pull` already do. Rather than keep a module in search of a caller, that
commit will be reduced to **`Handle.engine_replicas`** alone — the DP-head predicate
(`tp_rank == 0 and pp_rank == 0`), which the driver still needs to hand the manager the right
targets, and which is the TP-correctness fix independent of anything else.

The second commit — the move to `rollout/manager/` — **stands unchanged**. That package is still the
right home; it will hold a `Remote` rather than a plain object.

I will force-push the reduced branch. Reviewing before that lands is not useful; the move commit is
the only part that survives as-is.

CjhHa1:
## P1 follow-up fix

Fix commit: [CjhHa1/UniRL@a661d59](https://github.com/CjhHa1/UniRL/commit/a661d591a5f7b7be136a9da6db01357b0da44b0c) (`cherry-pick a661d591a5f7b7be136a9da6db01357b0da44b0c`). It is based on the current PR head `e12fda2`.

### Problems

- `RolloutPool` filled the lowest-index launcher to capacity before using the next one, so small drives could run entirely on one rollout worker while other GPUs stayed idle.
- Launch/localization ran while holding the pool condition, and a ready future released capacity before `result()` was validated. A slow localization could block quiescence, while a failed future could admit replacement work and then poison `pause()`/`drain()`, allowing active generations to race rollout sleep/shutdown.
- Async AR could launch past an off-cadence evaluation boundary. Eval then synchronized a newer policy, stale-filtered those already-admitted generations, but their generation IDs remained consumed, eventually causing `no admissible generations`.

### Changes

- Reserve launchers round-robin and perform the potentially blocking launch outside the condition.
- Keep completed calls capacity-accounted until the trainer acknowledges successful resolution; after failure, stop new dispatch while still allowing pause/drain/close to wait for outstanding calls.
- Clamp async AR admission to the next evaluation sync boundary.

### Validation

- focused CPU concurrency harness: balanced dispatch, blocked-launch quiescence, failed-result cleanup, and eval-ceiling cases passed
- `python3 -m compileall -q unirl/rollout/manager unirl/trainer/async_ar.py unirl/trainer/async_rollout.py`
- pre-commit hooks on all four changed files passed
- pre-push hooks passed

The earlier agentic partial/async staleness findings no longer apply after `e12fda2` removed those workflows.

zzhuoxin1508:
A few issues in the rollout-manager rework :

1.Wasted prefetch when weight_sync_interval=1 — next_hard_boundary/boundary_launch_slots (async_rollout.py) clamp launches by num_rollouts/eval/save but not by the weight-sync cadence, so on-policy runs launch a generation that keep_within_lag(0) always evicts at the next quiesce → ~2× rollout compute + one discarded data batch per step, no overlap. main's deleted launch_ceiling gated exactly this — please restore a budget==0 ⇒ don't prefetch clamp.

2.prefer_newer is dead code; CompleteGroups.take() is FIFO (buffers.py:114, filters.py:73) — collect trains the oldest in-budget group instead of the freshest (main used drain_freshest). Either wire prefer_newer into the filter chains or delete it and document the FIFO policy.

3.Jointly-invalid defaults (agentic.py:104 / dispatch.py:43) — worker_max_concurrency defaults to 1 while per_worker_inflight defaults to 8, but the pool requires worker_max_concurrency >= max(capacity)+2 (=10); any recipe that omits worker_max_concurrency raises at construction. Derive the default from per_worker_inflight.

4.Lost failure-isolation net + stale contract (engine.py:79 vs harness/protocol.py) — generate() no longer wraps harness.run(), yet protocol.py still promises "the hosting runtime keeps a last-resort net for harness bugs". Harmless with today's self-catching ToolAgentHarness, but an escaping error now aborts the whole batch — restore the net or fix the doc.

haonan3 (approved):
Follow-up commits (2fb0b013..a291c794) during review:

Correctness fixes on the new manager stack:
- bb136d92 — RolloutManager now poisons itself when resolving/routing a completed rollout fails: previously samples could be lost silently while empty/counts kept reporting clean state (the exact invariants the trainers assert on).
- 45974fbb / 6322ae6d — RolloutPool dispatch is balanced across launcher slots (fill-first left slot 1 idle under the default per_worker_inflight=8), and launches now run outside the pool lock via slot reservations, so pause/drain/counts no longer block behind slow localization.
- 80034e7e — restored freshness-gated admission (weight_sync_interval - batches_since_sync) lost with AsyncBatchControl: without it, every publication window launched one generation that keep_within_lag was guaranteed to discard — at weight_sync_interval=1 that meant ~2× rollout compute and every other data batch consumed but never trained.
- a291c794 — per_worker_inflight / worker_max_concurrency are validated before any runtime construction (the default combination previously failed only after the placement block built every remote).

PR #316, opened 2026-08-04
chore: collapse every docstring to one line, behind a lint guard

celve:
## Summary

`#305 chore: clean Python comments` removed 9,261 lines and **zero docstring lines** — it was a pure `#`-comment sweep. That worked (447 `#` comments remain across 107k lines), but docstrings were never in scope, and they are **25,086 lines, 23.4% of first-party Python**.

The cause is a wording gap. CLAUDE.md §3's only rule said *"keep only essential, concise one-line **comments**"*. A docstring is not a comment in Python's vocabulary, so that reads as `#` comments — which is exactly the sweep that ran. Nothing constrained docstrings and nothing linted them, so every new file arrived with fresh prose.

This adopts a uniform policy and makes it enforceable: **every docstring — module, class, and function — is at most one line, within the ruff `line-length`.** `lint/check_docstring_lines.py` enforces it from here. **-21,950 docstring lines.**

Four commits, reviewable in order:

1. **`docs:` relocation (no `.py` changes).** Prose that records something outside our control — what stock upstream does wrong and what breaks on a version bump, numerical/bit-identity constraints, implementer contracts — moved into the directory READMEs *before* anything was deleted. Three new: `sglang_diffusion/_patches/README.md` (26 patches as a what-breaks + DELETE-WHEN table), `vllm_omni/patches/README.md` (the DELETE-WHEN registry), `rollout/engine/README.md`. `types/README.md` restructured so the live `Sample`/`Part` contract is the body and the `RolloutReq` migration tables are an appendix — filename unchanged, five docstrings defer to that exact path.
2. **`refactor(cli):`** the 13 modules that pass their docstring as the argparse `description` — their headers are user-facing `--help`, not developer prose. Moved into a module-level `_HELP` constant, `--help` byte-identical. (`epilog=` does *not* work: it renders after the argument groups, reordering the output.)
3. **`chore:` mechanical collapse**, 515 files / 1,807 docstrings whose first line was already a self-contained summary. Script-generated, slicing source text rather than the decoded value so raw-string prefixes survive.
4. **`chore:` 283 hand rewrites + the guard.** These had summaries wrapping across 2+ lines (median first sentence 115 chars), so each was condensed by hand. Also drops 40 Linear ticket stamps from summaries and corrects four stale package summaries — `models/sd3/__init__.py` still called itself a *"parallel prototype … before the migration lands"*, and `SD3ModelBundle` existed nowhere but in that sentence.

CLAUDE.md §3 is updated to state the rule and point at the README as the home for anything that does not fit, explicitly including "don't turn it into a `#` block instead".

## Related Issue

N/A

## Test Plan

Verification leans on AST equality and exclusion proofs, since CI here is lint-only. All harnesses were one-off and are **not committed** (CLAUDE.md §5).

- **Only docstrings changed.** Parsed both revisions of every changed `.py`, blanked every `Module`/`ClassDef`/`FunctionDef` docstring, compared `ast.dump`: **531 files changed vs `upstream/main`, and the only 14 with any code difference are the 13 intended `_HELP` extractions plus the new guard script.**
- **`--help` byte-identical** for all 13 CLI modules — rendered both revisions through `argparse` with each file's real `formatter_class` under `COLUMNS=80` (9 `RawDescriptionHelpFormatter`, 4 default) and compared `format_help()`: **0 regressions.**
- **LLM tool prompts untouched.** The `"description"` literals in `rollout/env/tools/{calculator,sandbox,search,visit}.py` are prompt text that shapes training results and sit ~9 lines from near-identical docstrings, so they got their own check rather than relying on the AST proof: **0 changes** (2/2/2/3 literals, hashed).
- **The guard is not vacuous.** `python lint/check_docstring_lines.py` on `upstream/main`: exit 1, **2,106 violations**. On this branch: exit 0, `check-docstring-lines: 3143 docstrings, all one line.` It scans **587/587** first-party files, so a `SCAN_DIRS`/`SKIP_PARTS` typo cannot pass by scanning an empty set.
- `SKIP=no-commit-to-branch pre-commit run --all-files --show-diff-on-failure` — all hooks pass, and **ruff-format made no further adjustment** to any of the 515 swept files.

## Compatibility / Risk

No API, config, checkpoint, data-format, or resource change. The only runtime-visible surface is CLI `--help`, proven byte-identical above.

**The real risk is merge conflicts.** This touches 531 `.py` files and there are 41 open PRs. Every docstring edit sits at the top of a file or immediately under a `def`, so conflicts resolve trivially by taking the cleanup side — but the volume is real. Two mitigations: land it fast, and note that **commits 3 and 4 are regenerable** — on conflict, re-run the transform on a fresh base rather than resolving by hand (that is exactly how this branch was rebased onto `#156`).

One accepted deletion, called out so it is a decision and not an accident: `unirl/types/sampling.py` held the repo's only assertion-carrying doctest. Nothing executed it — there is no `tests/` tree and CI is lint-only — and CLAUDE.md §5 forbids recreating one.

The `.md` path-rot guard that would naturally pair with this is **deliberately not included**: it fails on landing today, because `models/bagel/__init__.py` already cites a `docs/BAGEL_MIGRATION_PLAN.md` that does not exist. Worth its own PR with those cases named.

## Reviewer Notes

**Read commit-by-commit.** Commit 3 is 515 files of uniform mechanical truncation — review the transform, not the diff. Commit 4 is where the judgment is: 283 hand-condensed summaries, and that is where knowledge could have been lost.

**Duplicate-work check:** `#305` is the merged predecessor (`#`-comments only); no open PR touches docstrings. `#173 chore(sglang_diffusion): simplify reviewer comments` is merged and unrelated.

**AI-assisted.** The mechanical passes were script-generated and AST-verified; the 283 rewrites and the README relocation were authored per-item. The full diff and every verification command above were run before publishing.

`unirl/utils/README.md` previously promised that each module there "is documented by its own docstring" — that sentence is corrected rather than left dangling. Guard-script rationale moved into `.pre-commit-config.yaml` comments, since CLAUDE.md keeps `lint/` to hook-referenced scripts only.

## Checklist


CjhHa1 (changes requested):
I don't think it is good to del all docstrings, it will lose many information

celve:
> I don't think it is good to del all docstrings, it will lose many information

It's not delete but collapse. And could you please give one example?

CjhHa1:
> > I don't think it is good to del all docstrings, it will lose many information
> 
> It's not delete but collapse. And could you please give one example?

For example:

Before:
```
"""Extract per-frame face embeddings.

        Args:
            video: ``(C, T, H, W)`` float in ``[-1, 1]``.
            with_grad: When True, the ArcFace forward is kept inside autograd
                (used for the GENERATED video on the BPTT path); when False,
                the whole computation runs under ``torch.no_grad`` (used for
                the REFERENCE video, which never receives gradient).

        Returns:
            embeddings: ``(1, T, 512)`` on ``self.device``. Differentiable in
                ``video`` when ``with_grad=True``.
            mask:       ``(1, T)`` on ``self.device`` (1 = face found).
        """ 
```

Now:
```
"""Extract per-frame face embeddings."""
```

This is no good as we lost informations. I think docstrings will be good if they tell us and agents the essential args and notifications

celve:
@CjhHa1 You're right, and the example you picked was the exact class of loss. Three commits pushed to address it.

The information that had no other home is the kind knowable **only at runtime** — a `torch.Tensor` annotation cannot say `[B, T, C, H, W]`. I mined every deleted line carrying a shape, mapped it back to its symbol, and put it back **inside the one line**. Your file:

```python
-"""Extract per-frame face embeddings."""
+"""Per-frame face embeddings from ``(C, T, H, W)`` in ``[-1, 1]``: ``(1, T, 512)`` + mask ``(1, T)``."""
```

101 sites restored. The ones that most needed it:

- **`types/segments/latent.py`** — ten tensor fields had no shape at all. Now annotated, plus the `sde_logp[:, s]` ↔ `sde_indices[s]` indirection, which is the fact that silently scores the wrong transition if a caller indexes by step index rather than SDE ordinal.
- **`models/types/batched_replay.py`** — the S steps are stacked **step-major** as `[S*B, ...]`. `_tile_conditions` is a `NotImplementedError` stub, so that docstring *was* the spec.
- **`models/bagel/`** — the only stage whose batch dim is fake: `[1, K, seq, C]` packed navit, not spatial.
- **`models/hunyuan_video/text_embed.py`** — `embed_llama` returns `[B, seq, 4096]`, `embed_clip` returns `[B, 768]`, behind an identical `TextEmbedCondition` annotation.
- **`algorithms/flowdppo.py`** — six bare `Tensor` kwargs whose body hard-codes "dims 0-1 are `(B, S')`".

**Nothing was written from memory.** A shape is restored only if its text exists in pre-deletion git history, and a gate checks it: every `[...]` on an added line must be a substring of that file's pre-deletion blob — **256 brackets checked, 0 untraceable**. Where upstream never stated a shape (ltx2 `replay`'s `prev_sample_means`) it stays unstated. A wrong bracket is worse than none.

This is also consistent with what the sweep already did — commit 04368672 had put shapes into 18 one-liners; it just wasn't applied to the rest. Shape-carrying docstrings: **127 → 285**.

Two supporting changes:
- `models/README.md` states the "final stored latent, fp32 forward, `[-1, 1]` → `[0, 1]`" convention once instead of eight times.
- CLAUDE.md — my wording was genuinely ambiguous. The "don't move prose into a `#` block" clause sat right under the one-line rule and could read as banning inline annotations, which §3 has always allowed. It now says shapes/dtypes/ranges are **expected**, names both vehicles, and the guard's failure message says where a shape goes.

Verification: AST equality with docstrings blanked — 102 files, **0 code differences**; `check-docstring-lines` reports 3143 docstrings all one line; `pre-commit run --all-files` green.

Happy to go further if specific files still read as too thin — the ledger of every deleted line is reproducible, so pointing at a file is enough.

haonan3 (approved):

PR #317, opened 2026-08-04
refactor(rollout): split the agentic rollout engine from a rank-0 manager role

celve:
## Summary

Splits the agentic rollout engine from the layer that schedules it (LIN-693). Three
commits; the first is #315 and is included here only because GitHub cannot base an
upstream PR on a fork branch — **merge #315 first**, and this shrinks to the last two.

**The problem.** `AgenticRolloutEngine` is 464 lines, of which 198 are a rank-0
coordinator: a task queue, a `next_task` pull server, a drain thread pool, per-drive
completion buffers, and a batch-level `generate`. It is simultaneously the thing that runs
a trajectory and the thing that decides which trajectories run where. Symptoms of the mix
are scattered around it — `RolloutOutput` is a union widened for this one implementer,
`abort(ids)` means something different here than on every other engine (with `del ids`),
and `worker_max_concurrency: 24` in the partial recipes exists because rank 0 serves its
own drain, every worker's pull, and the whole poll fan out of one actor's thread slots.

**The split.**

```
trainer          batch size, over-sample width, tail policy, sync cadence
   |               submit / collect / quiesce / bump_weight_version
AgenticManager   ROLE, rank-0 entry. Queue, placement, in-flight caps, group assembly,
   |             staleness buffer, turn-boundary interrupt. Drive thread.
AgenticRolloutEngine  ROLE. run_trajectory + set_stopping + lifecycle. No decisions.
   |
ToolAgentHarness      one trajectory (landed in #295)
SyncRolloutEngine     one turn
```

The engine goes 464 -> 197 lines. `run_trajectory` is today's `_run_one` with its body
unchanged; `generate` goes because driving a whole batch to completion is a scheduling
decision, and `self._n` goes with it because fanning a prompt into `n` GRPO siblings is
the manager's job.

**Why the manager is a role and not a driver-side object.** An earlier revision of this
design put the queue on the driver. The driver does not poll during its optimizer step,
so that drains the slab for seconds every rollout, and the mitigation (over-pushing into
Ray's actor mailbox) has no precedent in either system this repo cites for the pattern —
slime's `RolloutManager` is a Ray actor, and AReaL keeps a per-worker queue under its
controller queue. A manager with its own drive thread keeps placing work because it is not
the driver. It also closes an idle window that exists today: when a drive drains
naturally, `AsyncAgenticTrainer` submits nothing until the next save/sync boundary, so at
`weight_sync_interval > 1` the slab idles for up to `interval-1` steps.

Two details worth a reviewer's eye:

- **Placement targets engine replicas, not workers** — `tp_rank == 0 and pp_rank == 0`.
  A non-tp-zero rank is a shell whose `generate` returns `None`; least-outstanding would
  *preferentially* feed it, because failing instantly looks cheap. Identical to "every
  worker" today, so it costs one line and only matters the day agentic TP is wired.
- **Task identity is a manager-owned uid, not a Sample id.** `Part.fork(1, ...)` extends
  ids with `/{branch}`, so a prompt's `n` siblings carry identical ids —
  `PendingGroups` distinguishes them by counting to `n`. Keying replica affinity on a
  Sample id would pin a whole GRPO group to one replica and serialize it.

## Related Issue

LIN-693. Depends on #315.

## Test Plan

- `python -m compileall unirl/` — passes.
- `uvx ruff@0.15.1 check` + `format --check` (the pinned pre-commit version) — clean.
- `python3 lint/check_recipe_targets.py` — 2434 recipe `_target_` paths resolve.
- `python3 lint/check_experimental_boundaries.py` — ok.
- AST comparison proving `run_trajectory`'s body identical to `_run_one`'s, and
  `_attach_env_reward` unchanged.
- AST comparison proving the #315 move verbatim: all eight top-level definitions
  identical modulo the renames.
- An ad-hoc AST re-export check (every in-repo `from unirl.X import a` resolves) — this
  caught two real import bugs in #315 that nothing else here could.

Not run; reason: **nothing in this PR has been executed.** There is no Ray, torch or GPU
in the environment it was authored in — the touched modules do not import there — and this
repo's CI is lint-only. Every claim above is structural.

## Compatibility / Risk

**Draft, and not merely incomplete: the agentic trainers do not run on this branch.**
`agentic.py:126` still calls `rollout.set_workers`, `agentic.py:165` still calls
`rollout.generate`, and both partial and async still construct `AgenticManager` against
the pre-role signature. The remaining commits wire them, which is also where the twenty
ordering invariants get exercised. Do not merge before that lands.

Once wired, expected and *correct* metric movements, so a reviewer is not surprised:
`perf/generate_time_s` disappears on the barrier path (it wraps `rollout.generate`, which
no longer exists; the time reappears in the `step_time_s` residual), and `perf/step_time_s`
drops on the async path at `weight_sync_interval > 1` — that is the idle-slab bug above.

Decode-idleness becomes a maintained invariant rather than a structural one. Mitigated by
SGLang's native backend raising `"verb requires quiesced generation"` on any weight or
memory verb with in-flight generations — loud on native, which all six agentic recipes
use; silent on HTTP, which none do.

## Reviewer Notes

Still to come on this branch: build the manager role in the trainers' `placement(...)`
blocks on an isolated slot, rewire all three agentic paths (the barrier `generate` cannot
outlive the queue's move), fold in two latent bugs found nearby, and delete the
coordinator's remaining call sites.

Latent bugs found while investigating, each triaged rather than silently fixed:
`inner.tp_size>1` with `backend: http` spawns N N-way-TP servers on one GPU each unguarded
(fix belongs with the trainer rewiring); `agentic_async.py` wires weight sync with
`rollout.workers` and omits `tp_size`/`pp_size`, diverging from `async_ar.py`'s correct
shape (same); `_parallel_shape_from_init_kwargs` is a flat probe blind to TP nested at
`config.inner`, which also affects `ComposedRolloutEngine` (own issue);
`TensorWeightSync` picks its serializer by module-name substring so all four colocate
agentic recipes take the fallback path (own issue).

Prepared with AI assistance; every line needs human review before merge. Checked for
overlapping open PRs and issues — none besides #315, which this depends on.

## Checklist


celve:
Superseded by #315. The finalized design keeps RolloutManager as an ordinary driver-side object; it is not a Remote or rank-zero role. Closing this alternate implementation so review stays on the agreed direction.

PR #332, opened 2026-08-05
fix(reward): pass each VideoAlign prompt one video path

### LINKED ISSUE #328: Local VideoAlign treats MP4 paths as frame-image lists
## Observed behavior

The local VideoAlign scorer assigns the complete `video_paths` list to every Qwen-VL video block. In this message format, a list represents one video supplied as individual frame images; it does not represent multiple video files.

The shipped caller writes generated samples as MP4 files and invokes this path with `batch_size: 1` by default. A value such as `[".../sample_00000.mp4"]` therefore enters the per-frame image loader and fails before VideoAlign produces a score. `VideoAlignRewardScorer` catches that exception and reports the sample as unsuccessful, after which `RewardService` raises on the failure flag.

## Source-derived reproduction

For two paths and prompts, the current method constructs:

```text
prompt-a -> ["video-a.mp4", "video-b.mp4"]
prompt-b -> ["video-a.mp4", "video-b.mp4"]
```

An uncommitted construction probe ran the actual `_VideoAlignInference.reward` method from the base revision and intercepted its input at `_process_vision_info`. This confirms the structure above, but it is not an end-to-end inference reproduction.

The vendored VideoAlign implementation instead pairs `video_paths` and `prompts` with `zip` and gives each video block one string path.

## Expected behavior

Each prompt should receive the path at the same index:

```text
prompt-a -> "video-a.mp4"
prompt-b -> "video-b.mp4"
```

That sends MP4 paths through the video-file decoder and matches the vendored implementation.

## Impact and scope

This input-construction defect affects single-item and multi-item calls. It causes decoding to fail before scoring rather than producing silently mispaired rewards.

Correcting the video block advances execution to later, separate failures in the local scorer's model-call and logits handling. This issue covers only the prompt/video input construction; it does not claim the local scorer becomes functional end to end.

morluto:
## Summary

Local VideoAlign currently assigns the complete `video_paths` list to every Qwen-VL video block. In the Qwen-VL message format, that list means one video supplied as frame images, not a list of video files. Because the caller writes generated samples as MP4 files, even the shipped `batch_size: 1` path sends `[".../sample_00000.mp4"]` to the image loader and fails before a score is produced.

This change gives each prompt the MP4 path at the same batch index. That matches the vendored VideoAlign implementation, which pairs paths and prompts and passes one string path per video block.

This PR fixes only that input-construction divergence. The local scorer still cannot complete inference because of separate downstream model-call and logits-shape failures; those repairs are intentionally left for follow-up.

## Related Issue

Fixes #328.

## Test Plan

- Uncommitted source-construction probe against `origin/main` and `HEAD`, intercepting the actual `_VideoAlignInference.reward` method at `_process_vision_info` — base constructed `[[video-a.mp4, video-b.mp4], [video-a.mp4, video-b.mp4]]`; this branch constructed `[video-a.mp4, video-b.mp4]`. This checks message construction only, not decoding or model inference.
- `python -m compileall -q unirl unirl-reward-service/reward_service` — passed.
- `uvx --from ruff==0.15.1 ruff check unirl/reward/local/videoalign.py` — passed.
- `uvx --from ruff==0.15.1 ruff format --check unirl/reward/local/videoalign.py` — passed.
- `git diff --check origin/main...HEAD` — passed.
- End-to-end VideoAlign inference was not run. The known downstream model-call and logits-shape failures remain, so this PR does not claim an end-to-end runtime fix.

## Compatibility / Risk

This intentionally changes single-item and multi-item calls from the frame-image-list branch to the video-file branch. It does not change the reward request schema, configuration, or public model interface. It also does not make the local scorer functional beyond this corrected input boundary.

## Reviewer Notes

The diff is limited to aligning local chat-data construction with the vendored scorer. The downstream inference repairs are outside this PR.

AI assistance was used for investigation and drafting; I reviewed the exact diff and validation results. Duplicate-work check: this continues the existing #328/#332 contribution rather than opening parallel work.

## Checklist


haonan3:
**The pairing bug is real, and unlike most of this batch it lives in reachable code** — `VideoAlignRewardScorer` is wired through `unirl/reward/local/registry.py:31,51` and shipped in `examples/diffusion/wan21_t2v_videoalign_dancegrpo.yaml:75-78`. The one-line fix is also the right fix: the vendored reference (`unirl-reward-service/reward_service/scorers/_videoalign/inferencer.py:131-134`) pairs with `zip(video_paths, prompts)` and passes a single `"video": f"file://{video_path}"` string, and `git log` shows the local port (#97, `7bb74e7c`) simply dropped the `zip`. The caller slices `batch_paths`/`batch_prompts` in lockstep (`videoalign.py:620-623`), so `video_paths[idx]` is always aligned. We want this to land. Keeping it open — but two things need to happen first.

## 1. The PR body misdescribes the failure mode and needs to be corrected

In qwen-vl-utils' message format, `"video": <list>` does not mean "multiple videos" — it means **one video supplied as a list of frame images**. The vendored copy shows the branch: `_videoalign/vision_process.py:254` sends a `str` to the video-file reader and a list through per-element `fetch_image` (PIL). The only caller of `reward()` writes mp4 files (`sample_{idx:05d}.mp4`, `videoalign.py:614-618`), so `"video": [".../sample_00000.mp4"]` takes the frame-list branch and fails to decode — **on every call, including the shipped `batch_size=1` default**. The exception is caught at `videoalign.py:581-588`, rewards are zeroed with `successes=False`, and `RewardService` fail-fasts on the failure flags.

Two consequences for the Summary:

- *"The single-item behavior is unchanged"* is not true — single-item goes from **always failing** to actually decoding the video. This understates your own fix, and it also means the real path was never executed while preparing the PR.
- *"every item in a batch evaluate[s] the wrong prompt/video pairing"* (also #328's framing) never happens at runtime: the request crashes before any score is produced. The defect is a porting divergence from the vendored reference, and the observable symptom is a crash, not silently wrong rewards.

Please rewrite the Summary/Compatibility sections to state what the code actually does. The record matters for whoever reads this later.

## 2. As-is, the advertised outcome is not delivered — the scorer still cannot run after this diff

The fixed line feeds a pipeline that is broken further down, from the same port:

- `videoalign.py:526` calls `self.model(return_dict=True, **batch)`, but `_VideoAlignModel.forward` (`:285-293`) accepts no `return_dict` and **requires `position_ids`**, which the raw processor output does not contain → `TypeError` on every call. The helper that computes exactly those inputs, `_build_vl_inputs` (`:197`), has zero callers — dead code since #97.
- `videoalign.py:528` indexes logits as `[B, 3]`, but `forward` returns `torch.cat` over pooled `[B, 1, H]` (`:348-354`) → `[B, 1, 3]` (`[B, 3, 3]` with `use_special_tokens`), so `r[0].item()` raises on a 3-element tensor.
- Minor: `qwen_vl_utils` is imported at runtime (`:192`) but declared nowhere in `pyproject.toml`.

So merging this alone changes the crash site, not the outcome — "batch requests now use the intended mapping" cannot be observed. This is also why the "focused behavioral harness" in the Test Plan can only have asserted `chat_data` construction against a mock; the real path cannot execute end-to-end on either side of this diff.

## Two paths

Either is fine:

- **Extend this PR** to make the local scorer actually work end-to-end: this pairing fix, plus the model-call fix at `:526` (wire up or repair `_build_vl_inputs`), plus the head-indexing fix at `:528` — with one real inference run quoted in the Test Plan (a small checkpoint on a single short clip is enough).
- **Keep this PR as the one-liner**, but rescope the body honestly — "aligns prompt/video pairing with the vendored reference; the local scorer remains non-functional pending the model-call fix" — and we'll file the follow-up issue for the rest, credited to this investigation.

Same note as on the rest of the batch: before the next filing, run the code path you are changing, or state plainly that you could not. The pairing find here was genuine; the body's claims about runtime behavior were not checked against runtime.

morluto:
Thanks for spelling out the two options. I kept the small-scope path and updated the branch, PR body, and #328 accordingly.

The public record now describes the actual failure: the MP4-path list enters the frame-image branch and fails even at `batch_size=1`; it no longer claims silently wrong batch scores or unchanged single-item behavior. The one-line fix remains aligned with the vendored implementation, while the body explicitly states that the downstream model-call and logits-shape failures still prevent end-to-end scoring.

I also rebased onto current `main` and replaced the old behavioral-test claim with a clearly labeled source/construction probe plus the validation commands that actually ran.

leviking98z-rgb (approved):

PR #334, opened 2026-08-05
fix(sglang): preserve unset diffusion seed semantics

### LINKED ISSUE #329: Preserve stochastic diffusion behavior when seed is unset
## Observed behavior

The shared SGLang diffusion adapter converts an explicitly unset `DiffusionSamplingParams.seed` (`None`) into `seed=0` in the request sent to the diffusion engine. That makes an unset seed behave like a fixed seed instead of preserving the engine's stochastic default.

## Expected behavior

An explicitly configured integer seed, including `0`, should be forwarded unchanged. When the diffusion seed is `None`, the adapter should omit the seed from the SGLang request so the engine retains its global-RNG behavior.

## Evidence and scope

`DiffusionSamplingParams.seed` is typed as `Optional[int]`. The `ImageAdapter` sampling contract documents that an unset seed falls back to global RNG, and the SGLang denoising path distinguishes a missing seed from a fixed integer. The shared `ImageAdapter` request builder is inherited by the SGLang image and video-family adapters.

## Success criteria

- `seed=17` is forwarded as `17`.
- `seed=0` remains distinguishable from an unset seed.
- `seed=None` is absent from the engine request.
- The change remains limited to the shared SGLang diffusion adapter boundary.

morluto:
## Summary

An unset diffusion seed is a meaningful request: it lets the downstream SGLang engine use its global random-number generator. The shared image adapter previously converted seed=None to integer 0 before building the request. That silently changed an intentionally stochastic request into a deterministic one.

The adapter now omits the seed field when it is unset. Explicit seeds remain unchanged, including seed=0.

## Related Issue

Fixes #329.

## Test Plan

- Focused fake-engine boundary harness verified that seed=17 is forwarded unchanged and seed=None is omitted from the request.
- Focused Ruff validation passed.

## Compatibility / Risk

Explicit deterministic seeds keep their existing behavior. Only the previously incorrect None-to-zero conversion changes; no other rollout engine or sampling parameter changes.

## Reviewer Notes

The key contract is at the adapter boundary: absence of the seed must remain absence, while zero must remain an explicit seed.

## Checklist

### DIFF
--- a/unirl/rollout/engine/sglang_diffusion/adapters/image.py
+++ b/unirl/rollout/engine/sglang_diffusion/adapters/image.py
@@ -159,21 +159,23 @@ class ImageAdapter(ModelAdapter):
         """Sampling scalars + the σ schedule slice.
 
         ``diffusion.sigmas`` is length T+1 (terminal 0 included); SGLang's
-        ``set_timesteps`` wants the interior T. Pass the seed even when
-        ``initial_noise`` pins x_T verbatim: SGLang derives per-step SDE noise
-        deterministically (its ``_make_step_generators``, keyed on
-        ``denoise_seeds``) only when seed is not None — a None seed silently
-        falls back to global RNG.
+        ``set_timesteps`` wants the interior T. When set, pass the seed even
+        when ``initial_noise`` pins x_T verbatim: SGLang derives per-step SDE
+        noise deterministically (its ``_make_step_generators``, keyed on
+        ``denoise_seeds``). When unset, omit the seed so SGLang keeps its global
+        RNG semantics.
         """
-        return {
+        kwargs = {
             "num_inference_steps": int(diffusion.num_inference_steps),
             "guidance_scale": float(diffusion.guidance_scale),
             "height": int(diffusion.height),
             "width": int(diffusion.width),
             "num_frames": int(diffusion.num_frames),
             "sigmas": diffusion.sigmas.detach().cpu().tolist()[:-1],
-            "seed": int(diffusion.seed) if diffusion.seed is not None else 0,
         }
+        if diffusion.seed is not None:
+            kwargs["seed"] = int(diffusion.seed)
+        return kwargs
 
     def build_response(self, sample: Sample, raw: List[RawResult]) -> Sample:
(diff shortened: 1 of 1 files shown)


haonan3:
Closing this one — the premise is inverted, and the behavior the PR describes as the fix does not exist at the pinned engine version.

**The `None -> 0` coercion is deliberate, not an accident.** It landed together with the engine in `ceaba2bc` (#35), carried over verbatim from the retired v1 engine, whose comment stated outright that a `None` seed *"breaks per-step reproducibility / cross-engine alignment"*. The surviving docstring in `build_sampling` is a **warning about that failure mode**; issue #329 quotes it as though it described the desired behavior. This PR rewrites a warning into an endorsement of the thing it was warning about.

**Omitting the key does not restore "global RNG semantics."** At the pin (`sglang[diffusion]==0.5.12.post1`, `pyproject.toml:56`):

- `multimodal_gen/configs/sample/sampling_params.py:138` — `seed: int | list[int] = field(default=42, ...)`
- `multimodal_gen/runtime/pipelines_core/stages/input_validation.py:79` — `assert seed is not None` in `_generate_seeds`, before the per-output generators are built

So dropping `"seed"` from the request substitutes SGLang's **default constant 42** — a different deterministic seed, never the global RNG. The stated post-fix behavior is not what the code does.

**The branch is also unreachable.** `DiffusionSamplingParams.seed` defaults to `42` (`unirl/types/sampling.py:111`), and across the example configs every seed setting resolves to 42 — no config in the repo sets `seed: null`. There is no shipped configuration in which this diff changes anything.

Stepping back: per-step SDE noise reproducibility is load-bearing here, since replay recomputes log-probs that have to line up with the rollout trajectory. Making the seed optional at the adapter boundary is the wrong direction for this codebase even setting the two points above aside. Note also that `fastvideo/engine.py` keeps the explicit-seed contract, so this would have made the two diffusion engines diverge.

General note for the next batch: please confirm the trigger condition is reachable — name the config or the call chain that gets there — before filing, and check `git log -S` on the exact line first. The blame here points straight at a commit that explains why the current behavior is what it is.

PR #335, opened 2026-08-05
fix(advantages): keep normalization finite for invalid rewards

### LINKED ISSUE #330: Keep advantage normalization finite for non-finite rewards and singleton groups
## Problem

The reward-to-advantage paths can produce non-finite advantages when a reward is NaN or when normalization receives a singleton group or batch. At the `abc05c1` baseline, the retained `unirl/algorithms/normalizers.py` helpers use raw means and unbiased `std()`, while `Part.compute_advantages` has the same global-statistics issue and allows a non-finite reward to poison grouped statistics.

This is the focused S7/L1/L2 item from the audit umbrella in #326.

## Expected behavior

All retained normalization paths should keep their output finite:

- statistics use finite rewards only;
- non-finite reward rows receive zero advantage;
- a singleton group or global batch produces a finite zero-centered advantage;
- the policy matches the existing agentic filtering behavior.

For example, `[1.0, NaN, 3.0]` should produce `[-1.0, 0.0, 1.0]` under population standard deviation, and `[1.0]` should produce `[0.0]`.

## Scope

S7/L1/L2 only: reward and advantage normalization safety. Other audit findings remain separate.

morluto:
## Summary

Advantage normalization can turn one invalid reward into non-finite training values: NaN/Inf entered mean/std and poisoned the whole advantage tensor. Singleton groups/batches also blew up under unbiased `std()` (undefined for N=1).

This PR applies one policy across `Part.compute_advantages` and agentic grouping, via shared `finite_mean_std`:

```text
reward -> finite mask -> mean/std over finite rewards -> (r - mean) / (std + eps)
             invalid row ---------------------------------> 0
```

- Non-finite rows are excluded from stats and get advantage `0`.
- Finite singleton / empty-finite sets use std `1`, so advantage collapses to `0`.
- Exact all-equal groups still get advantage `0`.

Fixes #330.

## Intentional numeric changes (valid finite inputs)

These are deliberate alignments with the agentic path and the documented GRPO formula `(r - mean) / (std + eps)`, not accidental side effects of the NaN fix:

1. **Population std (`unbiased=False`)**  
   Previously, `scope="global"` / several helpers used PyTorch default sample std (`unbiased=True`). Advantages rescale by `√(N/(N-1))` (≈1.41 at N=2, ≈1.07 at N=8; negligible for large batches). Per-group path already used population variance.

2. **Denominator `(std + eps)` instead of `(var + eps).sqrt()`** on the Part group path  
   Old form floored the denom at `√eps ≈ 1e-4` (soft damping of near-ties). New form matches agentic / TRL / veRL-style `(σ + ε)` and only uses `ε` to avoid div-by-zero. Exact ties remain 0; near-ties (float noise) can now produce full-scale z-scores instead of being suppressed.

## Cleanup

- Removed unused `normalize_grouped` / `normalize_global` (zero in-repo callers).
- Removed unused `require_valid_group_ids` / `require_expected_group_sizes`.
- Shared helper: `unirl.types.advantages.finite_mean_std` (masks non-finite internally).
- `AgenticTrainer._group_advantages` now uses the same helper.

## Test Plan

- `[1.0, NaN, 3.0]` → `[-1.0, 0.0, 1.0]` (`Part` global + agentic-equivalent).
- Singleton `[1.0]` → `[0.0]`.
- All-NaN / ±Inf batch → all-zero advantages.
- Group with partial NaN: finite siblings normalize; invalid row → `0`.
- Cold import `unirl.types.sample` / `unirl.trainer.agentic` (no circular import).
- `ruff check` / `ruff format` on touched files.

## Compatibility / Risk

- In-repo: no callers of the deleted normalizer helpers.
- Downstream forks importing `unirl.algorithms.normalizers.normalize_*` directly will break (submodule-only; never exported from `unirl.algorithms`).
- Live training numerics change for valid finite rewards as noted above (population std + near-tie damping removed on Part group path). Binary all-correct / all-wrong groups unchanged (still advantage 0).

## Reviewer Notes

Check finite-mask + singleton handling together with the two numeric alignments. Output shape is unchanged; invalid-row behavior is explicit.

## Checklist

- [x] Reviewed the changed code and removed unused helpers.
- [x] Documented intentional std / eps behavior changes.
- [x] Spot-checked finite / singleton / Part↔agentic parity.


CjhHa1 (approved):
Fixed all nits

PR #336, opened 2026-08-05
fix(bagel): initialize strategy schedule during replay

### LINKED ISSUE #333: Initialize the strategy schedule during Bagel replay
## Problem

`BagelDiffusionStage.diffuse` initializes the SDE strategy with the rollout sigma schedule before stepping, but `BagelDiffusionStage.replay` previously omitted the same contract call. This matters for stateful `StepStrategy` implementations whose replay step needs the full schedule, and for replay paths that run on a fresh or reset stage instance.

## Expected behavior

Bagel replay should initialize the strategy with `segment.sigmas` after moving the schedule to the model device and before the first replay step. Stateless strategies such as the current default `FlowSDEStrategy` should remain unchanged because their inherited initialization is a no-op.

## Scope

The current Bagel pipeline defaults to `FlowSDEStrategy`, so the omission is not an observed failure in the default production path. The compatibility gap is at the public stage/strategy boundary: Bagel accepts a `StepStrategy`, and stateful strategies can require `init_schedule()` before stepping.

## Evidence

At the audited base revision `abc05c1`, the Bagel replay path moves `segment.sigmas` to the model device and enters the replay loop without calling `self.strategy.init_schedule(schedule)`, while the diffuse path does call it. A focused recording-strategy harness confirms the expected behavior on the proposed fix: replay records exactly one schedule initialization with the segment schedule, and the first replay transition runs only after initialization.

Success means Bagel replay initializes the schedule before replaying any stored transition, without changing Flow strategy behavior or broadening the diffusion strategy API.

morluto:
## Summary

Bagel diffuse initializes the strategy with the sigma schedule before generating transitions. Bagel replay did not perform the corresponding initialization, so replay worked for the current stateless FlowSDEStrategy but could use stale or missing schedule state for a stateful strategy.

Replay now calls init_schedule(schedule) after moving the stored schedule to the model device and before evaluating replay transitions. This makes the replay lifecycle match diffuse and preserves the current no-op behavior for FlowSDEStrategy.

## Related Issue

Fixes #333.

## Test Plan

- Recording-strategy replay harness passed through the production Bagel stage/step path and confirmed schedule initialization occurs before replay evaluation.
- Strategy contract check passed.

## Compatibility / Risk

No change for the current FlowSDEStrategy. Custom strategies that depend on init_schedule now receive the schedule before replay, as required by the stage contract.

## Reviewer Notes

The review order is: compare Bagel diffuse ordering with replay ordering, then check that schedule initialization precedes the first replay step.

## Checklist

### DIFF
--- a/unirl/models/bagel/diffusion.py
+++ b/unirl/models/bagel/diffusion.py
@@ -529,6 +529,7 @@ class BagelDiffusionStage(DiffusionStage[BagelDiffusionConditions]):
 
         schedule = segment.sigmas.to(device)
         sigma_max = schedule[1] if int(schedule.shape[0]) > 1 else schedule[0]
+        self.strategy.init_schedule(schedule)
 
         gen, cfg_text, cfg_img, image_shape = self._resolve_single(conditions)
         gi, gi_cfg_text, gi_cfg_img = self._build_generation_inputs(gen, cfg_text, cfg_img, image_shape, device=device)


haonan3:
Closing this one — it is a no-op on every path that can run, and in the one case it could affect, it removes a guard we want to keep.

**No-op today.** `init_schedule` has exactly two implementations: the base no-op (`unirl/sde/kernels.py:38-40`) and `DPM2Strategy` (`kernels.py:551-553`). Every strategy that can reach Bagel `replay()` inherits the `pass`, so the added call does nothing on any runnable path — the description says as much.

**The one strategy it could affect is structurally barred from replay.** `DPM2Strategy.step` returns `(result, None, None)` (`kernels.py:577`), and Bagel `replay` raises on a `None` log-prob (`unirl/models/bagel/diffusion.py:559-563`). `unirl/sde/README.md:65-66` documents this as intentional — *"replay is for stateless strategies"*. All Bagel/unified recipes pin `FlowSDEStrategy`.

**And in that case the change is actively harmful.** `DPM2Strategy.step` opens with

```python
if self._sigmas is None:
    raise RuntimeError("DPM2Strategy requires init_schedule() before stepping.")
```

`init_schedule` sets exactly that `self._sigmas` (`kernels.py:551-553`). So the added line silences the tripwire without making stateful replay correct: `replay` iterates a sparse, non-zero-starting `step_indices` subset (`flowgrpo.py:249-252`) over a freshly reset `_DPMState`, so a multistep solver's history would be rebuilt from the replayed subset rather than the true rollout history. That turns a loud crash into a silently wrong importance ratio — the opposite of what you want from a fail-fast.

**The asymmetry is not a Bagel bug.** All 12 diffusion stages call `init_schedule` in `diffuse()` only and none in `replay()`. This change would make Bagel the sole outlier rather than fixing an inconsistency.

General note for the next batch: per CLAUDE.md §2, "nothing speculative" — a change whose stated benefit is for a hypothetical future strategy, and which is a proven no-op for every strategy that exists, does not clear the bar. Please confirm the trigger condition is reachable before filing.

PR #338, opened 2026-08-05
fix(weight-sync): bound bucketed ZMQ operations

### LINKED ISSUE #337: [Bug] Bound bucketed weight-sync ZMQ operations
## Observed behavior

The bucketed CUDA-IPC/shared-memory weight transport uses a ZeroMQ REQ/REP handshake for initialization metadata, bucket metadata, and acknowledgements. On `main`, none of those operations sets `RCVTIMEO` or `SNDTIMEO`, so ZeroMQ's default is an infinite wait.

A receiver-side exception after accepting a request but before sending its acknowledgement leaves the trainer's corresponding send or receive blocked. The shipped colocated full-weight path is reachable through `examples/diffusion/sd3/sd3_vllmomni_full_ipc.yaml` and `IPCWeightSync.sync()`; that method is broadcast to every train rank and starts the receiver RPC in a daemon thread while the main thread runs the sender pump.

## Evidence

A real PyZMQ 27.1.0 / libzmq 4.3.5 harness exercised the shared-memory sender over an `ipc://` socket with no peer:

```text
baseline c998564: remained blocked after 0.600s
candidate #338 with a test-only 100 ms bound: TimeoutError after 0.254s
candidate shared-memory leaks=[]
```

This establishes the unbounded socket operation and the candidate's bounded failure/cleanup behavior. A CPU shared-memory round trip also completed with one tensor and matching values. No full CUDA engine reproduction was run.

## Expected behavior

Every bucketed transfer send and receive should have a finite internal bound. Expiry should identify the protocol phase and release the sender-owned CUDA/shared-memory buffer, shared-memory segment, and socket without reporting a successful transfer.

A configuration option is not required; a fixed 600-second per-operation bound covers the shipped transfer shape without adding unused API plumbing.

## Scope and limitation

This issue covers ZeroMQ operations and failures or crashes that let the receiver RPC unwind. It does not provide whole-sync cancellation.

`IPCWeightSync.sync()` still performs an unbounded `thread.join()`. If a rollout worker is genuinely hung inside `collective_rpc`, the socket operation can time out while the trainer subsequently waits forever at that join. Bounding or cancelling that RPC/thread is separate follow-up work.

## Proposed fix

#338 applies the internal operation bound, translates timeout failures, and moves buffer ownership ahead of the initialization acknowledgement so abort cleanup can remove `/dev/shm/diffrl_weights_*` segments.

morluto:
## Summary

The bucketed CUDA-IPC/shared-memory transport previously used ZeroMQ's infinite default for every REQ/REP send and receive. A receiver-side failure between a request and its acknowledgement could therefore leave each broadcasting trainer rank blocked in the sender pump.

This change:

- applies a fixed 600-second `RCVTIMEO` and `SNDTIMEO` to the sender and receiver sockets before bind/connect;
- translates `zmq.Again` into an operation-specific `TimeoutError`;
- records CUDA/shared-memory buffers before the initialization acknowledgement, so abort cleanup can release them;
- uses zero linger on aborted sockets while allowing a completed receiver up to the same bound to deliver its final acknowledgement.

The timeout is intentionally internal. No configuration or public API surface is added.

## Related Issue

Fixes #337.

## Test Plan

- Uncommitted real-socket harness with PyZMQ 27.1.0 / libzmq 4.3.5, using a test-only 100 ms bound:
  ```text
  baseline c998564: remained blocked after 0.600s without a peer
  candidate 7f7ba16: TimeoutError after 0.254s
  candidate shared-memory leaks=[]
  ```
- Real `ipc://` shared-memory round trip on CPU (CUDA synchronization calls stubbed because no GPU was available):
  ```text
  1 tensor received, values match, receiver finished
  ```
- `python -m compileall -q unirl` — passed.
- `python -m ruff check unirl/distributed/weight_sync/transfer/bucketed_transfer.py` — passed with Ruff 0.16.2.
- `python -m ruff format --check unirl/distributed/weight_sync/transfer/bucketed_transfer.py` — already formatted.
- `git diff --check origin/main...HEAD` — passed.

The harness remains uncommitted per the repository's test-artifact policy.

## Compatibility / Risk

There are no signature, recipe, or configuration changes. The 600-second bound applies to each socket operation, so a slow but progressing multi-bucket transfer can take longer overall.

This does not bound the existing `thread.join()` in `IPCWeightSync.sync()`. If a rollout worker is genuinely hung inside `collective_rpc` rather than failing and unwinding the receiver RPC, the trainer can still block at that join. Whole-sync cancellation is follow-up scope.

## Reviewer Notes

The core review path is `BucketedWeightSender._init_buffer()`, the socket operation wrapper, and the receiver's completed-versus-aborted cleanup. No pytest module or timeout plumbing remains.

AI-assisted; I reviewed the complete one-file diff and the real-socket evidence. The maintainer's overlap check in the review comment found no competing PR for this fix.

## Checklist


haonan3:
**The bug here is real and well-found** — this is the strongest of the batch, and I'd like the fix to land. But the diff needs to shrink by about half before it can. Leaving this open rather than closing it; see the two paths at the bottom.

Unlike #340, there is no open PR this could fold into — I checked everything currently touching this area. #316 touches `bucketed_transfer.py` but only collapses docstrings; #225 adds a separate `ckpt_engine_transfer.py`; #227 is SGLang LoRA sync. So this stays its own PR.

## Confirming the finding

The sender blocks unbounded at `bucketed_transfer.py:128`, `:148`, `:184` and the receiver at `:251`, `:276` — no `RCVTIMEO`/`SNDTIMEO` anywhere. A REQ socket gets no peer-death signal, so any receiver-side exception between `recv_pyobj()` and `send(b"")` parks the trainer permanently, and that window is easy to enter (`ipc_receive_mixin.py:149/156` raise on a bad `name_remap`, which the one shipped IPC recipe configures).

What makes it worse than it looks: `sync()` is `Dispatch.BROADCAST`, so every train rank runs the same pump. A deterministic receiver failure hangs **all** ranks at once, nobody reaches a collective, and torch's `ProcessGroupNCCL` watchdog never fires — the job sits there silently. And the comment at `full/ipc.py:112` (*"surface, don't let the pump hang forever"*) shows this was already known and that the existing mechanism can't work: catching the exception in a daemon thread does nothing for the main thread's blocking `recv()`. Good catch.

600 s is a defensible bound for the shipped shape (~5 GB bf16 into 2048 MB buckets is ~3 buckets, seconds each).

## What has to change

**1. Remove `unirl/distributed/weight_sync/transfer/test_bucketed_transfer.py`** (147 of the +302 lines). CLAUDE.md §5: the `tests/` tree was removed by policy (#99/#267) — *"do not recreate it under any name"*. `main` has zero pytest files under `unirl/`, and `[tool.setuptools.packages.find] include = ["unirl", "unirl.*"]` would ship this module — with its module-level `import pytest`, a dev-only dependency — into every wheel. There is also no pytest job in `.github/workflows/`, so it would never run. Run the harness, quote the commands and output in the Test Plan, don't commit it.

**2. The tests wouldn't be load-bearing anyway.** `_TimeoutSocket`/`_FakeContext` replace zmq wholesale and the fake raises `zmq.Again()` itself, so two of them assert only that the `except zmq.Again` line exists; `assert socket.options == [(zmq.RCVTIMEO, 1250), (zmq.SNDTIMEO, 1250)]` mirrors the implementation's `setsockopt` call order back at itself; two monkeypatch `torch.cuda.synchronize` so no IPC path runs at all.

**3. Drop the `timeout_s` fan-out.** Seven signatures across `full/ipc.py`, `composed/engine.py`, `synchronous.py`, and the whole `vllm_omni` backend Protocol gain a parameter no recipe sets. Setting the socket options inside `bucketed_transfer.py` is sufficient; a knob can be added when someone needs it. (Plumbing correctness itself is fine — I traced every implementation and dispatch path including the `collective_rpc` string dispatch, and there's no `TypeError` — it's just unused surface.)

**4. `_send` / `_recv` each take a `socket` argument neither body uses**, and the two functions are otherwise byte-identical. One helper, no dead parameter. (`ruff` misses it because `select` doesn't include `ARG`.)

**5. `DEFAULT_ZMQ_TIMEOUT_S` is defined and exported, then hardcoded as bare `600.0` in five other files** (`full/ipc.py:51`, `composed/engine.py:375`, `synchronous.py:86`, `vllm_omni/engine.py:214`, `vllm_omni/weight_sync.py:70`). Changing the default currently means editing six places.

**6. `_validate_timeout` guards impossible inputs** — non-finite and >24.8-day values that no caller can produce — plus a test for the `1e306` branch. CLAUDE.md §2: no error handling for impossible scenarios.

**7. It doesn't close the scenario end-to-end.** After the sender raises `TimeoutError`, `finally: thread.join()` at `full/ipc.py:127` is still unbounded and that thread is parked inside `collective_rpc`. A worker that is *hung* rather than *dead* still hangs the trainer, just on a different line. Your error text says "crashed or stopped responding" — only the "crashed" half is delivered.

One more thing worth putting in the body: moving `self.buffer`/`self.shm` ahead of the init ack also fixes a real `/dev/shm/diffrl_weights_*` leak on abort (`main` sets them only after `recv()`). That's a genuine improvement and reviewers shouldn't have to find it themselves.

## Two paths

Either is fine by us:

- **You revise it** — items 1–6 are mostly deletions, so the result should be roughly a 30-line diff in `bucketed_transfer.py` alone. Item 7 can be a follow-up.
- **You'd rather not** — say so and we'll take the fix in-house against #337, with the finding credited to you. No hard feelings either way; the diagnosis is the valuable part here and it's yours.

Let us know which you prefer.

morluto:
Thanks for the detailed review. I revised this along the first path and force-pushed a clean one-commit branch on current `main`.

Items 1–6 are addressed: the committed pytest module and all `timeout_s` plumbing/validation are gone; the transport now has one internal 600-second bound and one helper for all ZMQ operations. The final diff is limited to `bucketed_transfer.py` (36 additions, 15 deletions). It keeps the early buffer/shared-memory ownership so initialization failures clean up `/dev/shm`, uses zero linger on abort, and gives a completed receiver bounded linger for its final acknowledgement.

I also narrowed #337 and this PR to per-operation ZMQ failures. The existing unbounded `thread.join()` for a genuinely hung `collective_rpc` remains explicit follow-up scope.

The replacement uncommitted harness used real PyZMQ sockets: `main` remained blocked past 600 ms with no peer, while the candidate with a test-only 100 ms bound raised `TimeoutError` in 0.254 s and left no shared-memory segment. A real shared-memory IPC round trip also completed with matching tensor values. Compileall, Ruff check/format, and `git diff --check` pass.

leviking98z-rgb (approved):

PR #340, opened 2026-08-05
fix(train): defer unified FSDP gradient sync

### LINKED ISSUE #339: fix(train): synchronize deferred FSDP gradients in UnifiedModelTrainStack
## Observed behavior

When `training.fsdp.defer_grad_sync=true`, the shared `UnifiedModelTrainStack` optimizer step runs the AR and image algorithms over their microbatches, but the unified path does not toggle FSDP2 gradient synchronization around those backward calls. The single-stage `TrainStack` already applies the intended no-sync accumulation contract.

## Source-derived reproduction

For one shared optimizer step with two AR microbatches followed by one image microbatch, `abc05c1` makes no `set_grad_sync` calls:

```text
AR μ0 → AR μ1 → image μ0 → optimizer step
          (no FSDP sync scheduling in UnifiedModelTrainStack)
```

The expected contract is to defer reduce-scatter through every non-final microbatch and enable it before the final backward:

```text
AR μ0       AR μ1       image μ0
sync off →  sync off →  sync on → optimizer step
```

## Expected behavior

`UnifiedModelTrainStack` should apply one shared synchronization schedule across both algorithms: `set_grad_sync(False)` for non-final microbatches and `set_grad_sync(True)` for the final microbatch of each optimizer step.

## Scope

This is limited to the unified-model training stack's deferred FSDP gradient synchronization. Existing optimizer ordering, loss scaling, and non-deferred behavior should remain unchanged.

## Success criteria

- The AR and image backward calls share one ordered sync schedule.
- Only the final microbatch enables FSDP gradient synchronization.
- The change is covered by a focused behavioral regression harness or test.

morluto:
## Summary

UnifiedModelTrainStack performs one shared optimizer step from two ordered streams: AR microbatches followed by image microbatches. The generic training stack already honors defer_grad_sync, but the unified stack never toggled the backend. As a result, FSDP reduce-scatter could run once per microbatch instead of once for the combined optimizer step.

The unified stack now uses one global synchronization schedule:

```text
AR μ0 -> AR μ1 -> ... -> image μ0 -> ... -> image μlast -> optimizer.step
 off      off                    off                 on
```

The final image microbatch remains the synchronization point for the shared step, matching the existing stack contract and preserving the AR/image accumulation order.

## Related Issue

Fixes #339.

## Test Plan

- Dependency-isolated source harness observed no synchronization calls on the base revision and [False, False, True] for two AR microbatches followed by one image microbatch after the change.
- `python -m py_compile unirl/train/unified_model_stack.py`

## Compatibility / Risk

No optimizer, loss-scaling, or microbatch ordering changes. When deferred synchronization is disabled, the backend remains a no-op for these toggles; when enabled, one reduce-scatter is requested at the final global microbatch.

## Reviewer Notes

Review the absolute microbatch offset passed from the AR loop into the image loop. That offset is what makes the final image microbatch the single synchronization point for the shared optimizer step.

## Checklist

### DIFF
--- a/unirl/train/unified_model_stack.py
+++ b/unirl/train/unified_model_stack.py
@@ -253,6 +253,18 @@ class UnifiedModelTrainStack(Remote):
         )
         results: Dict[str, TrainStepResult] = {"ar": ar_result, "image": image_result}
         any_backward = ar_backward or image_backward
+        last_micros = image_result.micros or ar_result.micros
+
+        # Deferred reduce-scatter must be performed by the final microbatch.
+        # If that microbatch has no loss while an earlier one did, the accumulated
+        # gradients are still rank-local and must not be stepped silently.
+        if any_backward and last_micros and not last_micros[-1].has_backward and self.fsdp_backend.grad_sync_deferred:
+            raise RuntimeError(
+                "UnifiedModelTrainStack._train_one_step: defer_grad_sync deferred the gradient "
+                "reduce-scatter to the last microbatch, but it reported no backward while "
+                "an earlier microbatch did. Disable training.fsdp.defer_grad_sync or "
+                "investigate the empty final microbatch."
+            )
 
         if any_backward:
             # Defragment between multi-updates before NCCL gradient clipping.
--- a/unirl/train/unified_model_stack.py
+++ b/unirl/train/unified_model_stack.py
@@ -253,18 +253,6 @@ class UnifiedModelTrainStack(Remote):
         )
         results: Dict[str, TrainStepResult] = {"ar": ar_result, "image": image_result}
         any_backward = ar_backward or image_backward
-        last_micros = image_result.micros or ar_result.micros
-
-        # Deferred reduce-scatter must be performed by the final microbatch.
-        # If that microbatch has no loss while an earlier one did, the accumulated
-        # gradients are still rank-local and must not be stepped silently.
-        if any_backward and last_micros and not last_micros[-1].has_backward and self.fsdp_backend.grad_sync_deferred:
-            raise RuntimeError(
-                "UnifiedModelTrainStack._train_one_step: defer_grad_sync deferred the gradient "
-                "reduce-scatter to the last microbatch, but it reported no backward while "
-                "an earlier microbatch did. Disable training.fsdp.defer_grad_sync or "
-                "investigate the empty final microbatch."
-            )
 
         if any_backward:
             # Defragment between multi-updates before NCCL gradient clipping.
--- a/unirl/train/unified_model_stack.py
+++ b/unirl/train/unified_model_stack.py
@@ -168,6 +168,8 @@ class UnifiedModelTrainStack(Remote):
         micro_slices: List[Tuple[int, int]],
         *,
         training_progress: float,
+        microbatch_offset: int,
+        total_microbatches: int,
     ) -> tuple[TrainStepResult, bool]:
         """Backward one algorithm's Part over the given absolute ``micro_slices``
         (no zero_grad / no optimizer step).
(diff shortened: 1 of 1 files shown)


haonan3:
Thanks — the parity gap you found here is **real**, and I want to be clear about that up front. I'm closing this PR because the change belongs in #301, not because the finding is wrong. Details below, and I've left a note on #301 so it gets picked up there.

## The finding is genuine

`TrainStack._run_update` honors `training.fsdp.defer_grad_sync` (`unirl/train/stack/base.py:258`), and `UnifiedModelTrainStack` never calls `set_grad_sync` at all. `git show bdd64c0a:unirl/train/unified_model_stack.py` confirms the unified stack already had the same micro loop when the feature landed and simply was not updated, and there is no comment anywhere documenting an intentional opt-out. So a user who sets `defer_grad_sync: true` on a unified recipe today gets a knob that silently does nothing. That is worth fixing.

## Why it has to happen in #301, not here

#301 (`perf(unified-model): per-track micro-batch geometry for AR and image`, @leviking98z-rgb) predates this PR and rewrites the exact loop you are editing — its hunk `@@ -210,10 +267,19 @@ def _backward_part` is the `for start, end in micro_slices:` body.

More importantly, this is not a rebase conflict. #301 gives AR and image **independent** micro-batch geometry (`_optimizer_step_slices(bs, micro_batch_size=...)`), so the absolute-offset arithmetic here — `microbatch_offset + i == total_microbatches - 1`, with `total = len(ar_slices) + len(image_slices)` computed under one shared micro size — does not survive that change and has to be re-derived from the new geometry. Landing both independently would produce a plausible-looking but wrong sync point.

CLAUDE.md §5 asks for a duplicate-work check before filing; #301 has been open since 2026-08-02 and would have turned up.

## Two things that need to be different when it lands there

**1. The fail-closed guard has to come with the toggle.** `stack/base.py:256-293` is toggle **plus** a `RuntimeError` when the last micro reports no backward while earlier ones did — those two are co-designed, and `unirl/train/readme.md:85-89` explains why. In the unified stack the final micro is always an image micro, and both image algorithms return `has_backward=False` when `segment.sde_indices` is empty (`flowgrpo.py:176-177`, `bagel_flow_unigrpo.py:312-313`) — a batch-shared, all-or-nothing property. Without the guard, AR gradients accumulated with sync off get stepped on unsynced.

Your branch history shows you saw this: `83c4d56e` added the guard and `da241f3b` reverted it 11 minutes later. **The revert rationale in that subject line — "avoid rank-local empty-microbatch guard" — is a good catch**, since `has_backward` is rank-local and a rank-local `raise` can diverge or hang ranks. But the same objection applies to the existing guard at `stack/base.py:281-293`, so it deserves its own issue rather than being silently dropped here (the PR body still claims the change is "matching the existing stack contract", which it isn't once the guard is gone). If you want to open that issue, it would be a useful one.

**2. It's `perf`, not `fix`.** Nothing is broken today: `_grad_sync_enabled` initialises to `True` (`base_backend.py:295`) and `set_grad_sync` is its only mutator, so the unified stack currently reduce-scatters every microbatch and produces **correct** gradients — there is just one NCCL call per micro instead of one per step. It is also inert in every shipped config: `_defer_grad_sync = defer_grad_sync and not reshard_after_forward` (`base_backend.py:294`), all 20 occurrences in `examples/` are `false`, and all four `UnifiedModelTrainStack` recipes set `reshard_after_forward: true`. Framing that as a correctness bug in #339 overstates it, and it also means the change needs a parity check before anyone enables the flag — `compileall` and `git diff --check` don't cover it.

Closing in favor of #301. Thanks again for spotting the gap — it's a real one.

Issue #264, opened 2026-07-27
[Question] HI3 unified-model RL: any official experiments / recommended recipes?


niehen6174:
Hi, We're trying to run RL on HunyuanImage-3.0-Instruct with `train_unified_model` (recipes like `hi3_vllmomni_veomni_ep`).
In `hi3_vllmomni_veomni_ep.yaml` the sampling block says *"These small values are for the dump-first observe run"*, and PR #61 only mentions a short smoke test (PickScore ~0.85).
  

<img width="1256" height="549" alt="Image" src="https://github.com/user-attachments/assets/7390b870-5f61-4ec8-8e56-a73022c1ff08" />

**Question:** Has the team run longer HI3 unified RL experiments (e.g. 100+ rollouts)? If so, is there a recommended recipe or hyperparameter set we should follow (reward, lr/clip, LoRA, joint vs freeze_ar)?
 
Any pointers would be helpful — happy to share our run logs if useful.
Thanks!

CjhHa1:
I've assigned a teammate to follow your path

haonan3:
The pure text-to-image half of this question now has a validated recipe: #342 (just merged) adds `examples/unified_model/hi3_trainside_t2i.yaml`, measured well past smoke scale — **199 rollouts on 8x H20** with HunyuanImage-3.0-Instruct + PickScore, rollout reward rising **0.759 → 0.863** (first-30 mean 0.759 ± 0.060 → last-30 0.863 ± 0.016, plateau around rollout 80, no GRPO group ever collapsing to constant reward).

The hyperparameters you asked about are all pinned in the recipe: LoRA r=8 / α=16 on `qkv_proj` + `o_proj`, lr 2e-4, FlowGRPO `clip_range: 2e-5`, cfg=1, 16 inference steps with 3 SDE steps in the first half, 4 samples/prompt, seed 42. Two notes:

- Keep `forward_batch_size: 1` for now — the upstream batched chat template silently tokenizes only the first sample per call; #300 addresses that.
- This covers the diffusion-only T2I track (the AR objective stays frozen). The unified two-track (AR + diffusion) recipe you also asked about is still open on our side.

The recipe header carries a smoke override (`num_rollouts=2 eval_interval=0 logging.report_to_wandb=false`) if you want a quick sanity pass first. If you run it longer, we'd be glad to see your logs.

niehen6174:
[...]

PR #342, opened 2026-08-06
feat(hi3): pure text-to-image trainside FlowGRPO recipe

heguangxin:
## Summary

Issue #264 asks whether HI3 RL has a recipe validated past a short smoke run. This adds one for the pure text-to-image half: `examples/unified_model/hi3_trainside_t2i.yaml`, measured over 199 rollouts on 8x H20 with rollout reward rising from 0.759 to 0.863.

Three trainside gaps had to close first. Each failed silently rather than loudly:

- **The prompt shape was not selectable.** vLLM-Omni's `t2i_vanilla` is three settings, not one: `bot_task=image`, `sys_type=en_vanilla`, `sequence_template=pretrain`. Trainside t2i read only `bot_task`, so a recipe could set the other two and be ignored without any error, falling back to the Instruct checkpoint's `en_unified` / `instruct`. `modes/t2i.py` now forwards both from the root Part's `control`.
- **The system prompt tokenized as its own section.** The checkpoint wrapper emits system and user as two text sections and tokenizes each independently, shifting BPE boundary merges even with no separator between them: the issue-264 prompt came out 102 tokens against the official builder's 101. For the `pretrain` direct-T2I shape only, the system text is now folded into the prompt. That branch requires `cfg_factor == 1`, no message list, no CoT text and no cond image, so CFG and multimodal paths cannot reach it.
- **Driver-authored noise never reached t2i.** `diffusion.py` has taken `sde_sample_keys` since #186, but only `modes/it2i.py` supplied them and `params.noise_group_ids` was never populated from the Part. So `sampling.seed` governed neither x_T nor the per-step SDE draw; both fell back to ambient RNG, which makes a run depend on batch composition and chunk order. `_prepare_seeded_sampling` mirrors the it2i helper, minus the cond-VAE generators t2i has no use for.

## Related Issue

Refs #264. This answers the diffusion-only T2I half. The unified two-track (AR + diffusion) recipe the issue also asks about is not covered here.

## Test Plan

Re-verified against `main` at `c998564`, from the repo root.

- `SKIP=no-commit-to-branch pre-commit run --all-files --show-diff-on-failure` — all 17 hooks pass, 37s.
- `python -m unirl.train_diffusion --config-name unified_model/hi3_trainside_t2i --cfg job --resolve` — exit 0.
- Resolved-config equivalence against the measured run: 134 keys on each side, no structural difference, and the only three value differences are environment-local (`logging.entity`, the two PickScore `${oc.env:...}` paths). The recipe reproduces the run behind the curve rather than approximating it.
- Concat failure behind the `tokenizer_output` gate, reproduced CPU-only:

  ```python
  import torch
  from unirl.distributed.tensor.batch import _concat_value

  def tok(n):
      return {"tokens": torch.zeros(1, n, dtype=torch.long)}

  _concat_value([tok(101), tok(101)], batch_sizes=[1, 1])  # -> tokens (2, 101)
  _concat_value([tok(101), tok(117)], batch_sizes=[1, 1])  # -> RuntimeError: Sizes of tensors must
                                                           #    match except in dimension 0
  ```

- Training run: 8x H20, HunyuanImage-3.0-Instruct, PickScore, seed 42, 199 rollouts. `rollout/reward_mean`, 64 images per point (16 prompts x 4 samples):

  | rollouts | 0-19 | 40-59 | 80-99 | 120-139 | 160-179 | 180-198 |
  | --- | --- | --- | --- | --- | --- | --- |
  | mean | 0.7626 | 0.8336 | 0.8577 | 0.8623 | 0.8586 | 0.8626 |
  | sd | 0.0581 | 0.0340 | 0.0264 | 0.0161 | 0.0174 | 0.0144 |

  First 30 rollouts average 0.7594 +/- 0.0597, last 30 average 0.8633 +/- 0.0157: +0.104, with the spread of the per-rollout mean down to a quarter of its starting width. Within-rollout spread (`rollout/reward_std`) holds near 0.062 throughout and `rollout/zero_std_group_ratio` is 0 at every point, so no GRPO group ever collapsed to a constant reward.

  Over the 398 optimizer updates, `train/ratio_mean` sits 3.1e-6 below 1 (below 1 on 347 of them), `train/approx_kl` averages 4.0e-11, and mean `clipfrac_lt_one` is 0.043 against `clipfrac_gt_one` 0.001. Update 0 is on-policy by construction (`diffuse_kv_cache: false`, fbs=mbs=1) and its `ratio_mean` averages 1.7e-6 from 1 — the rollout-versus-replay agreement this recipe exists to expose.

## Compatibility / Risk

One new recipe file and two modified modules. No config schema, checkpoint, data format, API or resource change.

Two previously-silent behaviours change on the trainside t2i path: a request carrying a driver-authored noise recipe now honours it (matching it2i since #186), so trajectories for such a run change; and with `diffuse_kv_cache: false` the unread `tokenizer_output` is no longer attached. No recipe under `examples/` drives trainside t2i today, so the blast radius is this PR's own file.

Default paths stay byte-identical: `sys_type` and `sequence_template` default to `None`, the fold branch additionally requires an explicit `sequence_template`, and `_prepare_seeded_sampling` returns its inputs unchanged when `params.seed is None` or the Part carries no noise recipe.

## Reviewer Notes

**Overlap with #300.** Two contact points, both agreed. The `tokenizer_output` gate in `modes/t2i.py` is line-equivalent to the ones #300 applies to `it2i` and `t2ti` — same condition, differing only in a redundant parenthesis and the comment above it. #300 needs it for B>1 batching; this needs it because `batch_size > num_devices` puts two prompts on one rank and `Part.concat` then merges two `TokenizerEncodeOutput` objects of different token length. By agreement this PR lands first, and #300 has dropped `modes/t2i.py` from its diff. We also both edit `text_embed.py`, in the same function for different reasons: this adds the `sequence_template` parameter and the pretrain fold, #300 wraps the `apply_chat_template` call in its batching shim. Those compose — the merged form keeps `sequence_template=effective_sequence_template` inside #300's `with` block.

The `FieldKind.SHARED` declaration on the conditions class does not prevent that, which is the easy thing to misread. `modes/t2i.py` stores `diff_conds.to_dict()`, so the Part holds a bare `TokenizerEncodeOutput` under `Part.conditions`, which is itself `FieldKind.CONCAT` (`types/sample.py:93`) — the field kind is dropped at the `to_dict()` boundary. `TokenizerEncodeOutput` extends the diffusers `BaseOutput`, i.e. an `OrderedDict`, so `_concat_value` recurses in and `torch.cat` fails. The repro above is the real dispatch path. (`cond_vit` is unaffected: `to_dict()` keeps it as an `ImageEmbedCondition`, which takes the `Batch` branch where field kinds still apply.)

**`forward_batch_size: 1` is not a throughput choice.** Above 1 the upstream batched chat template tokenizes only the first sample of the call — `tokenization_hunyuan_image_3.py` passes a literal `prompt_list=[[]]` and zips it against a length-B kwargs list, so samples 1..B-1 never reach the tokenizer and nothing raises. #300 fixes that; until it lands this recipe must stay at 1, and the YAML says so.

## Checklist


haonan3 (approved):
LGTM

PR #345, opened 2026-08-07
feat(sft): support Qwen3-Omni audio SFT

zzhuoxin1508:
## Summary

Enable supervised fine-tuning of Qwen3-Omni on standalone WAV question-answering data,
without adding a second SFT path.

Data flow:
`DCASE parquet -> supervised JSONL (audio + prompt + response) -> ARSupervisedTrackBuilder -> existing Qwen3-Omni Turn/embed path -> response-only SFT loss`

Audio (plus the text question) is the input; the supervised target is text only. The bundle
loads `Qwen3OmniMoeThinkerForConditionalGeneration` — the Talker is never instantiated — and
both towers stay frozen, so only the Thinker's text-decoder attention LoRA trains.

- `unirl/train/sft/track_builder.py` — collect URI-backed `role="prompt"` media from supervised
  records and pair it with the batched text as existing `user` Turns, then call the chat stage's
  existing `embed()`. Records carrying `media_refs` that the selected path cannot consume now
  raise instead of falling through to the text-only embed (see Compatibility / Risk). The legacy
  Qwen-VL `role="condition"` image path is unchanged.
- `unirl/models/qwen3_omni/chat_template.py` — `embed_sft_prompt()`, which wraps supervised
  text + media into the same Turn pair rollout produces, so SFT and rollout share one encoding.
- `unirl/data/sft.py` — document the AR (omni) manifest row shape.
- `datasets/dcase2025_audio_qa/convert_dcase2025_audio_qa_to_unirl.py` — add `--supervised`
  (response targets + a final-answer prompt matching that target) with its own default output
  directory, preserving the original RL conversion behavior.
- `examples/sft/validation/qwen3_omni_audio_sft_lora.yaml` — 4-GPU Qwen3-Omni Audio QA SFT
  recipe: vision and audio towers frozen, Thinker attention LoRA, audio-in-video disabled
  because this recipe uses standalone WAV input.

## Related Issue

N/A

## Test Plan

- Qwen3-Omni standalone WAV SFT, 200 steps on 4 GPUs, `eval_loss 0.70389 -> 0.07697`
  ([W&B run](https://wandb.ai/zhouzhuoxin1508-tencent/unirl-qwen3-omni-sft/runs/sr6sdw4i)):
  ```
  SFT_DATA=datasets/dcase2025_audio_qa_sft/train.jsonl \
  SFT_EVAL_DATA=datasets/dcase2025_audio_qa_sft/val.jsonl \
  ENTRY=train_sft bash examples/run_experiment_single_node.sh \
    sft/validation/qwen3_omni_audio_sft_lora
  ```
  Note: the response is `The answer is X.` (~6 tokens, 5 of them fixed template), so this loss
  drop mostly measures answer-format acquisition.
- Audio ablation on the SFT'd model, 64 samples: correct audio `79.7%`, shuffled audio `56.3%`,
  silence `57.8%`. This is the load-bearing result — it confirms the adapted model is genuinely
  conditioned on the waveform.
- Candidate-likelihood evaluation, 64 samples: base `75.0%`, SFT `79.7%`. **Within noise**
  (48/64 vs 51/64; binomial SE ≈ 5.2pp). This PR does not claim a task-accuracy improvement;
  a larger eval set is follow-up work.
- Qwen2.5-VL image SFT regression: 100 steps, `eval_loss 1.31315 -> 0.82027` — the
  `role="condition"` image path is unaffected.
- Unified DCASE converter CLI:
  ```
  python datasets/dcase2025_audio_qa/convert_dcase2025_audio_qa_to_unirl.py --help
  ```
  confirms `--supervised` defaults to `datasets/dcase2025_audio_qa_sft` and the RL default
  stays `datasets/dcase2025_audio_qa`.
- Routing harness (run, not committed) over 8 (chat stage × manifest) combinations, asserting
  that `role="prompt"` media reaches `embed_sft_prompt`, mixed and text-only omni batches still
  embed, the Qwen-VL image branch is untouched, and every unconsumed-media shape raises:
  ```
  omni  audio role=prompt       -> embed_sft_prompt
  omni  mixed prompt/text-only  -> embed_sft_prompt
  omni  audio role=condition    -> RAISE ValueError
  omni  text-only manifest      -> embed(texts)
  vlm   image role=condition    -> images branch
  vlm   image role=prompt       -> RAISE ValueError
  llm   text-only               -> embed(texts)
  llm   stray media_refs        -> RAISE ValueError
  ```
- `SKIP=no-commit-to-branch pre-commit run --files <changed files>` — all hooks pass.

## Compatibility / Risk

- **New failure mode, deliberately loud.** A supervised record whose `media_refs` no path can
  consume now raises. Previously such media was silently dropped and training continued
  text-only: SFT loss still falls on MCQA answer priors, and `DP_SCATTER` made the check
  per-record, so a single mis-roled row could diverge one rank undetectably. Manifests that
  relied on the silent drop will now fail fast; prompt media must use `role="prompt"`.
- **Converter default output directory changed for `--supervised` only.** It previously shared
  the RL default (`datasets/dcase2025_audio_qa`), so running `--supervised` without `--out-dir`
  overwrote the RL manifests in place with differently-prompted rows. Non-supervised behavior
  and any explicit `--out-dir` are unchanged.
- No checkpoint-format, API, or resource-requirement changes.

## Reviewer Notes

- The supervised prompt intentionally drops "Reason step by step" from the RL prompt, since the
  target is a direct answer. The answer format `The answer is X.` is still matched by
  `MCExactMatchRewardScorer`'s phrase regex, so an SFT → RL handoff works on the reward side,
  but the two prompt distributions differ by design.
- **Overlaps with #243** (`perf(sft): optimize supervised data preparation`), which rewrites the
  same `_embed_prompts` to add prefetching. Audio is currently decoded synchronously inside the
  training step, which is exactly what #243 moves off the critical path for images — worth
  agreeing on merge order and folding audio into that prefetch.
- Known limitation: free-form autoregressive evaluation produces incomplete answers for both the
  base and adapted Qwen3-Omni model. The SFT/replay path and candidate-likelihood evaluation
  work; this PR does not change the model generation implementation.
- Pre-existing and out of scope: `MediaRef` rejects `role="target"`, which `unirl/data/sft.py`,
  `DiffusionSupervisedTrackBuilder`, and `unirl/utils/prepare_sft_t2i.py` all still use — worth a
  separate issue.
- AI assistance was used for part of this change and for the review pass that added the
  unconsumed-media guard; the diff has been read line by line.

## Checklist


zzhuoxin1508:
Clipboard_Screenshot_1786107215

haonan3 (approved):

PR #347, opened 2026-08-07
feat(minimax-h3): t2va model package for Flow-GRPO (trainside FSDP)

celve:
## Summary

Adds `unirl/models/minimax_h3/` — a **t2va** (text → video + stereo audio) model
package for MiniMax-H3 — plus the trainside Flow-GRPO recipe that drives it.

MiniMax-H3 is a ~31B guidance-distilled omni-modal DiT that denoises video and
stereo audio **jointly in one packed sequence**. Three properties shaped the
package and are checkpoint contracts rather than tuning knobs:

1. **Guidance is distilled** — no CFG, no negative prompt, one forward per step.
   `guidance_scale` is inert.
2. **Two sigma schedules in one sequence** — video `shift=12`, audio `shift=3`.
   Only the video grid is pinned onto the `Part`; the audio grid is derived
   in-stage from the same static formula.
3. **Mixed-dtype weights** — `_keep_in_fp32_modules` (patch projections,
   timestep MLP, output heads, rope) stay fp32 while the block stack is bf16.
   This is why `root_wrap: false` is load-bearing in the recipe: it keeps those
   tensors outside every `fully_shard` group so `MixedPrecisionPolicy` never
   all-gathers a bf16 compute copy. H3 conditions on `t = 1 - sigma`, and bf16
   near 1.0 has ~0.008 spacing — roughly 40% relative error on sigma at the
   low-noise end.

Alongside the package, four small changes outside it:

| file | change | risk |
| --- | --- | --- |
| `unirl/models/types/meta_init.py` | `finalize_meta_init(..., keep_in_fp32=None)` — **opt-in** mixed-dtype pinning for meta-init bundles | none; `None` default reproduces the previous uniform cast exactly |
| `unirl/reward/local/clap.py` | `ClapProcessor` takes `audio=`, not the removed `audios=` | bug fix; `clap` raised `TypeError` before this |
| `unirl/reward/local/video_pickscore.py` | new `frame_selection: "first" \| "middle"` | default `"first"` keeps every existing recipe byte-identical |
| `unirl/reward/local/t2av_composite.py` | forward `frame_selection` to inner specs that declare it | additive |

## Related Issue

N/A on GitHub. Tracked internally as **LIN-690 — Support MiniMax H3**.

## Test Plan

Hardware for all training runs: 1 node × 8 × H20 (95 GB), TaiJi GZ.

**Lint / format** (pinned `ruff` v0.15.1, matching `.pre-commit-config.yaml`):

```
uvx ruff@0.15.1 check  <19 changed .py files>   # All checks passed!
uvx ruff@0.15.1 format --check <19 changed .py> # 19 files already formatted
```
Full `pre-commit run --all-files` was **not** run — `pre-commit` is not
installed on this machine. `ruff check .` across the whole repo reports 9
pre-existing errors, none in files this PR touches.

**Hydra config + `_target_` resolution**, re-run against the **rebased** tree
(in a detached `git worktree`, so the live training checkouts were untouched):

```
  --config-name=diffusion/minimax_h3/minimax_h3_t2va_trainside --cfg job --resolve
# EXIT=0
# checked 20 _target_ block(s) -> RECIPE_KEYS_OK
```

**Numerical parity against the reference implementation** (CPU, no checkpoint):

- sigma grid vs `MiniMaxH3Scheduler` for both the video (`shift=12`) and audio
  (`shift=3`) schedules: **max |Δ| = 0.000e+00** (bit-identical).
- velocity-sign convention vs H3's own Euler step: **7.8e-07**, with an
  un-negated control that diverges by **1.97** — i.e. the test discriminates.
- confirmed against the real checkpoint: negated final latent σ = 1.04 vs
  un-negated 2.96.

**Single-GPU rollout smoke test:** one `generate()` outside Ray/FSDP, 768×768 /
124 frames, 10 steps — video and stereo audio decode, per-stage timings sane.

**End-to-end Flow-GRPO training:** multiple concurrent arms, longest **81
rollouts** (~20 h continuous), no OOM, no errors, checkpoints written and
reloadable. Reward `0.5·videopickscore + 0.5·clap` on `datasets/pickscore/`.

**Reward does not improve — see Reviewer Notes.** That is why this is a draft.

**Rebase note.** The branch was rebased onto `upstream/main` after the training
runs, picking up 12 commits. One real conflict, in
`unirl/reward/local/video_pickscore.py`: upstream's #198 switched
`Images(pixels=...)` to `Images.from_dense(...)` and #305 removed the
section-divider comments. Both upstream changes were adopted; the net diff of
that file against `main` is now exactly the `frame_selection` feature and
nothing else. `unirl/algorithms/flowgrpo.py` also gained
`_require_replay_anchor_for_batched_replay`, which this recipe satisfies (it
does not set `batch_replay_steps`, and it already uses
`old_logp_source: replay`). Lint, format, config and `_target_` resolution were
all re-run post-rebase and pass; **the multi-hour training numbers above were
produced on the pre-rebase tree.**

## Compatibility / Risk

- **Additive for existing recipes.** No default changes outside the new package
  except the two reward files above, both of which preserve current behaviour.
- **Vendored third-party code** (~3.1k of the 4.9k lines) under
  `unirl/models/minimax_h3/vendor/`, taken from **huggingface/diffusers PR
  #14355 @ `abc5e9bf`**, an *unreviewed draft branch*. Provenance, the exact
  file mapping, what was deliberately not vendored, and the mechanical
  import-rewrites are recorded in `vendor/VENDOR_COMMIT.txt`. **Expect API
  churn** — no released diffusers ships MiniMax-H3, and pinning the git branch
  was rejected because it collides with the `sglang[diffusion]` diffusers
  override already in `pyproject.toml`. Re-sync before trusting parity numbers.
- **Geometry is not free.** H3 was released for a 768px short edge, both axes a
  multiple of 32, 5–15 s at 24 fps. The pipeline **raises** on anything other
  than 768×768 / 124 frames rather than silently re-resolving. There is no cheap
  low-res setting, so this recipe cannot be smoke-tested on small GPUs.
- **`eta` defaults to 0.2, not the usual flow-GRPO 0.7.** 0.7 destroys this
  checkpoint — it is guidance-distilled, so its trajectory is sharp and it was
  never trained to recover from states the SDE noise knocks it onto. At 0.7 the
  rollout produced blue/white noise and every reward was scored on garbage. The
  recipe records the measured sweep behind 0.2.
- **`sde_indices: [0, 3, 6, 9]`** is a measured memory constraint, not a
  preference: `replay()` builds one autograd graph across every replayed step,
  and at ~22k packed rows all 10 steps need >100 GB/rank and OOM an H20 *after*
  a ~10-minute rollout.
- No new dependencies; no checkpoint-format or data-format changes.

## Reviewer Notes

**Draft because the reward curve is flat.** The port is numerically correct and
trains stably, but across four independent arms (30–81 rollouts each) no
configuration shows reward improvement:

| arm | group size | n | slope/step | t |
| --- | ---: | ---: | ---: | ---: |
| F | 8 | 40 | +0.000136 | +0.49 |
| G2 | 8 | 30 | +0.000122 | +0.23 |
| H | 4 | 81 | −0.000007 | −0.07 |

Diagnosis so far, from LoRA checkpoints rather than the reward curve
(`lora_B` initialises to exactly zero, so `‖B‖` measures policy movement
directly):

- The update is **alive** — `‖B‖` grows monotonically every step, reaching
  `‖ΔW‖ ≈ 3.84` by rollout 30.
- The motion is **directed, not diffusive** — net displacement over path length
  `D/L` = 0.70–0.91 against random-walk references of 0.27–0.58.
- So the policy is moving coherently and `videopickscore` is indifferent to
  where it points. The open question is what the gradient is actually
  optimising, not whether the optimizer works.

Already ruled out along the way: GRPO group size 2 (for two samples the
normalised advantage is algebraically always ±1, so magnitude is discarded — the
recipe now defaults to 8), a dead `clap` term (image captions vs generated
ambient audio is near-random by construction), blank frame-0 scoring (H3 clips
can open on a fade; scoring frame 0 cost −0.0815, ~6.7× the entire policy
signal at eta 0.2), and `lr` (AdamW is scale-invariant, so the small
`grad_norm` does not imply a stalled update).

**Reviewers should focus on:** the FSDP wrap interaction in the recipe comments
(`root_wrap: false` + `module_prefix: transformer_blocks` are jointly
load-bearing — an unscoped LoRA target list injects adapters into
`token_refiner.refiner_blocks.*`, outside every `fully_shard` group, whose grads
no collective would DP-sync); and the vendored-code churn risk above.

**AI assistance:** this branch was developed with Claude Code. Every changed
line has been executed on real hardware, not just type-checked. **Duplicate-work
check:** no open PR or issue on `Tencent-Hunyuan/UniRL` mentions minimax, h3, or
t2va.

## Checklist

      Diff is 21 files / +4883 / −19, no checkpoints, datasets, media, logs or
      credentials.
      The recipe carries the design rationale inline. No test files: the
      `tests/` tree was removed by repo policy (#99/#267), so the verification
      harnesses used here were run and their results quoted above rather than
      committed.

## EXP
NFT:
image
Flow GRPO: 
image


CjhHa1 (approved):
LGTM

PR #348, opened 2026-08-09
feat(reward): EditScore family scorer served by in-child vLLM

nussejzz:
## Summary

Adds an **EditScore** scorer to the reward service — the first VLM-as-judge reward served by a **full vLLM engine inside a managed scorer child** — plus two recipes exercising it under real training.

One class covers the whole EditScore family (VectorSpaceLab, ICLR 2026): `backbone` (qwen25vl / qwen3vl × HF / vLLM), base checkpoint, LoRA adapter, `score_range`, `num_pass` and `tensor_parallel_size` are constructor params, so swapping EditScore-7B for EditScore-Qwen3-VL-8B (reported on the official card as matching the original 72B) is a config change, not a code change. The wrapper absorbs the package's deployment quirks so recipes stay declarative:

- one-time LoRA→base merge is cached (`cache_dir` on shared storage; ~17GB, ~45s merge for 8B);
- `gpu_memory_utilization` / `enable_sleep_mode` / extra engine kwargs are injected around the package's hardcoded `LLM()` call — required whenever the judge shares a GPU with training;
- `enable_sleep_mode` maps `onload`/`offload` to vLLM `wake_up()`/`sleep(level=1)` (measured: offload 5.1s, 47GB→1.0GB idle; wake 0.3s; scores bit-identical across the cycle) and forces `mm_processor_cache_gb=0` — the frontend multimodal cache survives sleep while the engine-core receiver cache does not, and a post-wake hash-hit corrupts scores silently;
- `max_image_side` (default 1536) caps vision tokens so dataset-native ~3000px sources cannot overflow `max_model_len` mid-run;
- `report_sub_metrics` narrows the emitted metrics — recipes emit only `overall` (`sqrt(min(SC) × PQ)`), the calibrated scalar the model card defines;
- batched scoring: all SC / PQ prompts of a pass go through `batch_inference` as two batched calls instead of 2·N sequential generates (2.6× faster cold-vs-cold), with per-item fallback on any batched-path failure.

Stacked on #303 (managed scorer child / lifecycle protocol), sibling of the #304 BAGEL recipe, rebased onto the current review-fix stack. The scorer already speaks the new per_call boot protocol (`enable_sleep_mode` ⇒ `supports_offload` ⇒ a `--boot-offloaded` child sleeps the engine before first reporting ready). The parent-side protocol is untouched: a child hosting an entire vLLM engine is invisible to the trainer.

## Related Issue

N/A — groundwork and motivation are in #302 / #303 / #304 (stacked PRs).

## Test Plan

All on 8× H20 (96GB), EditScore-Qwen3-VL-8B (base `Qwen/Qwen3-VL-8B-Instruct` + `EditScore/EditScore-Qwen3-VL-8B-Instruct` LoRA), scorer child in a dedicated venv (vllm 0.20, torch 2.11 cu129).

- Standalone scoring semantics (single GPU, no training): identity pair → `prompt_following=0` / `consistency=10` (detects the un-applied edit), unrelated pair → `consistency=0`, malformed item → NaN without failing the batch; ~3.3s/sample at 2×1024² images.
- Sleep-cycle probe: score → `offload` → `onload` → score; scores bit-identical across the cycle, idle footprint 1.0GB.
- per_call smoke: `--config-name=diffusion/bagel/bagel_it2i_managed_editscore_smoke`, 2 rollouts, `offload_train_during_reward=true` — rewards 4.32 / 3.85, `ratio=1.0000±0.0000`; child lifecycle exactly 3×offload + 2×onload + 2×score per rank (rollout engine sleeps after generate, trainer offloads during scoring, judge wakes only to score).
- 200-rollout all-resident convergence run (train FSDP8 + vLLM-Omni rollout DP8 + judge DP8, `gpu_memory_utilization=0.23` sized from measured peaks; opengpt image-edit data): 200/200 zero faults, reward 4.91 (first-20 mean) → 6.63 (last-20 mean), peak 8.79; GPU peak 91.5/97.9GB uniform across ranks, ~87s/rollout. Curve: https://wandb.ai/dingzuhao7-tencent/bagel-editscore/runs/52sq0aop
- Batched-vs-sequential parity (cold-vs-cold on disjoint 8-item sets): 2.6× speedup, 8/8 within one decile step (7/8 bit-equal; batch composition perturbs bf16 logits under temp-0.7 sampling).
- Rebased-stack smoke (post #303 boot protocol): engine-backed child comes up `state=offloaded`, 2 rollouts reproduce baseline scale (4.39 / 3.89) — first real-training validation of `--boot-offloaded` with an engine-hosting scorer.

## Compatibility / Risk

- New optional scorer: registers via the existing optional-import pattern; environments without the `editscore` package are unaffected (registry warns and skips).
- Child venv needs `vllm` + `editscore` + `peft`; torch build must match the host driver (cu-suffix pinned to the training venv's).
- No parent-side protocol or API changes; recipes are additive.
- Memory limit documented rather than hidden: the QIE NFT recipe variant does not fit 96GB parts (EMA rejects `offload_train_during_reward`, resident trainer ~75GB + judge wake ~27GB) — it targets H200-class nodes and points H20 users at the BAGEL recipe pattern.
- Batched scoring can move a score by one decile step vs solo requests (bf16 batch-composition effect, noted in code); `batched: false` restores strictly sequential behavior.

## Reviewer Notes

- `report_sub_metrics` is the same shape as the fix applied to EditReward in #304 (`sub_metric_reduce: first`): the RL signal is one calibrated scalar; the other sub-metrics stay available for logging.
- Follow-up (not in this PR): `tensor_parallel_size > 1` for the 72B variant needs the multi-GPU child unlock (`allow_multiple_visible_devices`) plus a rank-0-spawns / others-share-URL rendezvous — planned as the next PR.
- Three upstream-worthy findings against the `editscore` package / vLLM are documented in the scorer comments (LLM kwargs not exposed; mm cache desync across sleep; slept EngineCore not detecting parent death).
- AI-assisted development was used for implementation and test harnesses; all changes were reviewed and validated by the author.

## Checklist


CjhHa1 (approved):
LGTM

PR #320, opened 2026-08-04
feat(sft): add WAN2.1 UCF-101 full-transformer SFT

leviking98z-rgb:
## Summary

- Add target-video supervised fine-tuning for WAN2.1: bounded video decode -> WAN VAE clean latent -> flow-matching SFT.
- Add a reproducible **UCF-101 full-transformer SFT** recipe and deterministic manifest builder.
- Reuse one bounded PyAV/tensor video loader across SFT, WAN V2V, and Qwen3-Omni while keeping the SFT builder as a thin manifest-to-Part adapter.
- Make PyAV available to the core Linux install for worker-side target-video decoding.

## Related Issue

N/A

## UCF-101 Full-Transformer SFT

Model and dataset:

```text
WAN2.1-T2V-1.3B
UCF-101 18-class sports/action subset
Full transformer finetuning
VAE and text encoder frozen
```

Training setup:

```text
2,367 training videos / 128 held-out validation videos
17 frames, 256x448
1 node x 8 NVIDIA H20 96 GB GPUs
Global batch size 8
Learning rate 2e-6
300 optimizer steps
Checkpoints at steps 100, 200, and 300
```

Results:

```text
Eval loss at initialization: 0.21981
Best eval loss, step 260: 0.16230  (-26.2%)
Final eval loss, step 300: 0.16264
Final train loss: 0.16008
Final grad norm: 0.30942
```

### Training curves

The screenshot comes from the completed W&B run. `eval/loss` drops sharply in the first 20 steps and continues improving to its best value at step 260. Raw per-step `train/loss` is noisy because each diffusion batch samples a different timestep; `train/loss_ema_50` shows the underlying downward trend.

WAN2.1 UCF-101 Full SFT W&B training curves

W&B run: [`leviking98z-zhejiang-university/unirl-pr320-wan21-video-sft/jgifozoc`](https://wandb.ai/leviking98z-zhejiang-university/unirl-pr320-wan21-video-sft/runs/jgifozoc)

### Qualitative comparison

The matched-prompt/matched-seed panels below compare the untouched WAN2.1 base model with the step-300 full checkpoint at START / MIDDLE / END. They demonstrate UCF-101 distribution adaptation rather than reconstruction of a source clip.

**Archery**

WAN2.1 Base vs UCF-101 Full SFT: archery

**Surfing**

WAN2.1 Base vs UCF-101 Full SFT: surfing

**Basketball dunk**

WAN2.1 Base vs UCF-101 Full SFT: basketball dunk

[...]

PR #352, opened 2026-08-11
feat(sft): interleaved image-text agent-trajectory SFT (qwen_vl + bagel)

haonan3:
## Summary

SFT on interleaved image-text **agent trajectories**: retrieved images enter the history mid-conversation, are seen by the model, and condition its later turns. Agent records may now carry OpenAI-style content-part lists (`[{"type": "text", ...}, {"type": "image", "image": <uri>}, ...]`) in history messages; the final assistant turn stays text-only — supervision is text CE, never an image loss, and the diffusion side is untouched.

Both backbones run the same manifests end to end:

- **Qwen2.5-VL** — new `QwenVLChatTemplateStage.embed_messages` renders multi-image histories through the processor (any number of images, any position; overlong prompts raise instead of truncating, since truncation desyncs vision spans). Replay was already multi-image capable; conditions now always emit batch-aligned per-row `pixel_values`/`image_grid_thw` lists — a collapsed `None` from an image-free DP build shard desynced the concat'd list from batch rows and broadcast the whole batch's features into every scatter shard.
- **BAGEL** — new `BagelChatTemplateStage` renders histories into Bagel's native ordered splits (bare `<|im_start|>`-wrapped text chunks + ViT-only image splits, no role tokens; target convention `encode(text) + <|im_end|>`); `pack_und_forward_inputs` generalized from `[image | prompt | response]` to ordered interleaved splits (image blocks full-attn at one rope position, text causal — the vendored `forward_train` layout); inference-mode replay prefills every split up to the last image and grad-forwards the trailing text.

Data path: `datasets/searchgen/prepare_sft.py` converts SearchGen-20K traces + corpus-1m into per-assistant-turn agent records (analysis → candidates → selection → refinement), extracting candidate images from the corpus TAR shards (parallel shard streaming, image-only SQL filter, missing-shard tolerance). The 283K generated images + judge scores are deliberately untouched — that is GRPO group-shaped preference data, not SFT. Recipes: `examples/sft/qwen_vl_agent_sft.yaml`, `examples/sft/bagel_agent_sft.yaml`; guide: `datasets/searchgen/README.md`.

## Related Issue

N/A

## Test Plan

GPU e2e (Taiji 8×H20, box conda torch 2.7.1 / transformers 4.56 / flash-attn 2.8.1, `pip install --no-deps -e .` + `opencv-python-headless`; data cooked by the moved converter, `datasets/searchgen/prepare_sft.py --max-traces 16 --candidates-per-query 3` → 30 train / 3 val records, 53 candidate images):

- `ENTRY=train_sft bash examples/run_experiment_single_node.sh sft/qwen_vl_agent_sft num_steps=1 batch_size=8 eval_interval=0 save_interval=0 backend.scheduler_cfg.warmup_steps=0` → `step 1/1 loss=1.98498 grad_norm=12.2500`, exit 0.
- `ENTRY=train_sft bash examples/run_experiment_single_node.sh sft/bagel_agent_sft num_steps=1 batch_size=8 eval_interval=0 save_interval=0 backend.fsdp_cfg.master_dtype=null` → `step 1/1 loss=1.86019 grad_norm=0.3496`, exit 0 — re-run against the final tree after the `system_instruction` removal; loss is bit-identical to the run before it, confirming the removed branch was unreachable. (`master_dtype=null` only for that box's torch 2.7 — pre-2.11 fsdp2 asserts one dtype per FSDP group; the recipe's fp32 LoRA masters target the pinned torch 2.11.)

CPU harnesses (run, not committed — tests/ tree removed by policy):

- Tiny random-init Qwen2.5-VL + real processor: record → `ARSupervisedTrackBuilder.build` → `SFT._masked_ce` → backward on interleaved 2-image records; CE = 11.935 ≈ ln(vocab) exactly as expected for random init; eval-pad masking; image-pad count == merged grid product per row.
- Real converted manifest through the same path (mixed image-ful/image-free batch), plus a `pytree_cat`/`pytree_chunk` harness reproducing the dp=8 build→concat→scatter seam: 8/8 shards feature/token aligned after the fix (all misaligned before).
- Bagel packer fake-model checks: `[vit,text]` byte-parity with the previous packing, interleaved `[text,vit,text]` and two-image layouts (rope/index/mask asserted), trailing-split helper edge cases; `BagelChatTemplateStage` against the real BAGEL tokenizer + `ImageTransform`s (split kinds/wrapping, target ends `<|im_end|>`, overflow raises).
- Review-fix harness (14 checks, all pass): unknown `text`/`image` part fields and unsupported part types rejected; image targets still rejected; **tool-call agent targets still normalize** (`content=None + tool_calls`, `content + tool_calls`, post-tool final answer) so #234's capability is provably intact; Bagel raises on `role="tool"`, on `tool_calls` in history, and on a `tool_calls`-carrying target, while still accepting a plain text target; and `embed_messages` no longer injects a configured `system_instruction` (agent manifests are authoritative, matching QwenVL).
- Review-round-3 harnesses (run, not committed): **(a) Qwen-VL tool rejection** — scratchpad venv (CPU torch 2.13 / transformers 5.15) + the real `Qwen2.5-VL-7B-Instruct` processor, 15 checks all pass: plain-text agent targets tokenize byte-identical to the previous generic-fallback path; a `tool_calls`-carrying target raises on the stage AND through the `ARSupervisedTrackBuilder` dispatch (the pre-fix fallback provably dropped it silently — same ids as text-only); `content=None + tool_calls` raises cleanly instead of a deep Jinja TypeError; `embed_messages` rejects `tools` / `role="tool"` / history `tool_calls`, while tool-free histories render byte-identical to the pre-fix path. **(b) Reference numbering** — old vs fixed converter over the full 96,848-trace release (shared candidate DB): `_full_trace` admission identical (59,559); 996 sampled traces (1/50) whose filtered-s2 list already matched `ordered_references` produce byte-identical conversations (0 diffs); the old code's silent misnumbering reproduces on 497 actually-rendered traces and all 497 now render with canonical numbering (0 newly dropped); 1,169 previously-dropped traces are rescued; all 2,619 fixed-code conversations pass the canonical-numbering invariants (refs `1..n` in `ordered_references` order, `Selected references` == s3 indices, citations in bounds). **(c)** e2e cook of the fixed converter (`--max-traces 40`): exit 0, 69 train / 3 val records, missing-shard tolerance exercised.
- `pre-commit run --files <changed>` — all hooks pass, including `lint/check_recipe_targets.py` and `lint/check_experimental_boundaries.py`.

## Compatibility / Risk

- Agent-record schema is extended, not changed: string contents and existing text-only agent manifests are byte-identical through the builder (verified); the omni `embed_sft_prompt` media path (#345) is orthogonal (legacy prompt/response records only).
- **The shared SFT manifest contract is not narrowed.** Tool-call assistant targets stay supported at the data layer — `prepare_sft_agent` (#234, `examples/sft/qwen3_agent_sft_lora.yaml`) supervises them by design — so "the target must be plain text" is enforced by the backbone that cannot render it (Bagel), not by `unirl/data/sft.py`. The only tightening in the shared layer is on this PR's own new shape: content parts now reject unknown fields instead of silently dropping them.
- Agent histories now render identically on both backbones: `BagelChatTemplateStage.embed_messages` no longer prepends a configured `system_instruction` when the conversation lacks a `system` message. `BagelPipeline` is its only construction site and never passed one, so that branch was unreachable on every live path; removing it makes Bagel match QwenVL (`system_instruction` belongs to the legacy `embed` path), so one manifest/config pair cannot render differently per model. The agent-capable chat-stage contract — `embed_messages`, `supports_message_images`, optional `tokenize_agent_target`, plus this precedence rule — is documented on `ARSupervisedTrackBuilder`.
- `QwenVLChatTemplateStage.embed` rollout path refactored into a shared packer — byte-identical behavior (reviewed line-by-line; truncation semantics preserved). The rollout Turn-path "at most one image turn" guard is intentionally untouched.
- `BagelPipelineConfig.max_prompt_length` was briefly added then removed as dead config; the live knob is `pipeline.max_prompt_length` (recipes wire the constructor directly).
- `pack_und_forward_inputs` signature changed (`prompt_ids`/`image` → ordered `splits`); its only caller (`BagelARStage._replay_train`) is updated in the same commit.
- The SearchGen converter moved from `unirl/utils/prepare_sft_searchgen.py` to `datasets/searchgen/prepare_sft.py` (the `datasets/ucf101`, `datasets/video_r1_260k` layout). Invocation changes from `python -m unirl.utils.prepare_sft_searchgen` to `python datasets/searchgen/prepare_sft.py`; both recipe headers and `.gitignore` are updated in the same commit, and no other references remain.

## Reviewer Notes

- AI-assisted (Claude Code); duplicate-work check: no open PR/issue covers interleaved agent SFT (closest are #320 WAN diffusion SFT and #345 omni audio SFT, both orthogonal).
- An adversarial review pass ran before publishing; its findings (image-only SQL filter for candidates — web HTML results silently dropped ~20% of traces; missing-shard tolerance; dead config; numbering-space weld in selection targets) are fixed in `817d3c13`.
- **Review follow-ups** (rebased onto `05abbe37` and re-verified): (1) the SearchGen converter now lives with its dataset, plus a `datasets/searchgen/README.md` covering the gated non-commercial corpus license; (2) content parts fail loudly on unknown fields; (3) Bagel fails loudly on `role="tool"` turns and on `tool_calls`-carrying targets — the one case where its text-only target tokenizer diverged silently from Qwen. A fourth suggestion — enforcing "target has no `tool_calls`" in `unirl/data/sft.py` — was deliberately **not** taken: it would have deleted #234's tool-call SFT capability, which the data layer declares and `prepare_sft_agent` produces.
- **Second review round** (Codex): the module contract now states the real invariant — the target may not be a *part list*, but may carry text and/or `tool_calls`, and the rendering stage must emit the whole target or raise — and the part-list shape is described as UniRL's own strict text/image schema rather than "OpenAI-style", since unknown fields are rejected. The `system_instruction` precedence divergence is removed (above). A structural `Protocol` for the chat-stage seam was deliberately deferred: all four AR stages implement `embed_messages`, but only 2 of 4 implement the full agent surface (`supports_message_images` / `tokenize_agent_target`), so pinning the seam now would be structurally misleading and would reach into `qwen3`/`qwen3_omni` — a separate change.
- **Third review round** (PR thread): both findings verified against the real template/data and adopted. (1) @leviking98z-rgb's blocking finding — the stock Qwen2.5-VL chat template references neither `tools` nor `tool_calls` (tokenizer and processor templates are identical), so the generic HF suffix path silently supervised only the text of a tool-call target; fixed per their option 2: `QwenVLChatTemplateStage` now rejects tool-shaped records exactly like Bagel (`embed_messages` guards + a stage-owned `tokenize_agent_target` that otherwise delegates byte-identically to the shared suffix helper). `unirl/data/sft.py` stays untouched — #234's Qwen3 tool-call SFT is unaffected. (2) @zzhuoxin1508's numbering finding — adopted via their first suggestion: on all 50,808 guard-passing traces `selected_reference_indices` is exactly `1..len(ordered_references)`, so the converter sources the reference list from `s4_generation_manifest.ordered_references` and drops traces with unresolved/reordered entries; the `queries[query_index]` alternative was not taken — `query_index` indexes the image-search-query subsequence, not the raw mixed list (naive mixed-list indexing disagrees on ~13% of rows). **Existing manifests must be re-cooked**; the curves above ran on manifests where 497 rendered traces (~1%) carried misnumbered references — loss impact negligible, curves not re-run.
- Inference-mode Bagel replay freezes the whole prompt when a history *ends* on an image (documented in the docstring); the agent recipes pin `replay_mode: train`, which packs everything into one grad `forward_train`.
- Generated-image + judge-score data (GRPO group shape) is intentionally left for a future RL PR.

## Checklist


haonan3:
## Training curves — real manifest, shipped recipe defaults

The Test Plan's 1-step runs prove the path executes end to end; they cannot show the model *learns* from the interleaved history. This is `sft/bagel_agent_sft` at its **shipped defaults** (`num_steps: 300`, `batch_size: 16`, `eval_interval: 20`) over a real SearchGen manifest, with loss measured on a held-out split.

The signal to read is the **eval** line: 153 records from traces the model never trains on. It falls across all 16 evaluations, and at 0.62 epochs the train and eval curves have not separated — so this is learning, not memorization of a small manifest.

### BAGEL-7B-MoT — bagel_agent_sft (und/AR experts, LoRA r64)

300 steps logged (`300/300`), 0.62 epochs; train loss 2.3395 → 1.2615 (5-step mean 1.2706).
Held-out eval loss 2.3920 → 1.4106 over 16 evaluations.

```mermaid
xychart-beta
    title "BAGEL-7B-MoT — bagel_agent_sft (und/AR experts, LoRA r64) — train (smoothed) vs held-out eval loss"
    x-axis "step" [0, 10, 20, 30, 40, 50, 60, 70, 80, 90, 100, 110, 120, 130, 140, 150, 160, 170, 180, 190, 200, 210, 220, 230, 240, 250, 260, 270, 280, 290, 300]
    y-axis "loss" 1.137 --> 2.506
    line [2.3395, 1.9804, 1.7760, 1.6177, 1.5437, 1.5077, 1.4884, 1.4371, 1.5005, 1.4157, 1.4163, 1.3888, 1.3961, 1.3431, 1.3556, 1.3664, 1.4362, 1.3544, 1.3881, 1.2987, 1.3013, 1.3534, 1.3574, 1.3124, 1.2868, 1.3010, 1.2969, 1.3130, 1.3752, 1.2906, 1.2706]
    line [2.3920, 2.3920, 1.7883, 1.7883, 1.6435, 1.6435, 1.5643, 1.5643, 1.5384, 1.5384, 1.5208, 1.5208, 1.4885, 1.4885, 1.4781, 1.4781, 1.4721, 1.4721, 1.4663, 1.4663, 1.4479, 1.4479, 1.4335, 1.4335, 1.4350, 1.4350, 1.4236, 1.4236, 1.4176, 1.4176, 1.4106]
```

| step | epoch | train loss | eval loss | grad_norm | lr |
| ---: | ---: | ---: | ---: | ---: | ---: |
| 1 | 0.00 | 2.3395 | — | 0.303 | 1.00e-04 |
| 20 | 0.04 | 1.6359 | 1.7883 | 0.463 | 1.00e-04 |
| 40 | 0.08 | 1.4749 | 1.6435 | 0.539 | 1.00e-04 |
| 60 | 0.12 | 1.3896 | 1.5643 | 0.512 | 1.00e-04 |
| 80 | 0.17 | 1.5951 | 1.5384 | 0.531 | 1.00e-04 |
| 100 | 0.21 | 1.3100 | 1.5208 | 0.478 | 1.00e-04 |
| 120 | 0.25 | 1.3361 | 1.4885 | 0.562 | 1.00e-04 |
| 140 | 0.29 | 1.3051 | 1.4781 | 0.519 | 1.00e-04 |
| 160 | 0.33 | 1.5030 | 1.4721 | 0.551 | 1.00e-04 |
| 180 | 0.38 | 1.4908 | 1.4663 | 0.555 | 1.00e-04 |
| 200 | 0.42 | 1.2660 | 1.4479 | 0.566 | 1.00e-04 |
| 220 | 0.46 | 1.3875 | 1.4335 | 0.578 | 1.00e-04 |
| 240 | 0.50 | 1.1588 | 1.4350 | 0.547 | 1.00e-04 |
| 260 | 0.54 | 1.4010 | 1.4236 | 0.594 | 1.00e-04 |
| 280 | 0.58 | 1.4726 | 1.4176 | 0.637 | 1.00e-04 |
| 300 | 0.62 | 1.2615 | 1.4106 | 0.527 | 1.00e-04 |


<details>
<summary>Data provenance and environment</summary>

Cooked with the moved converter, all defaults except the trace cap:

```bash
python datasets/searchgen/prepare_sft.py \
  --searchgen-root /path/to/Datasets/searchgen \
  --out-dir datasets/searchgen/processed \
  --max-traces 6000
```

- **25m19s, exit 0** → 7,674 train / 153 val records, 26,522 extracted images.
- 6,000 traces rendered; 3,391 dropped by the converter's missing-shard tolerance. The corpus mirror is missing 1 of 28 `cached-images` shards (~1,002 members), and a trace is dropped if *any* of its 12–20 candidate images is unavailable — hence the large fraction. This is the documented behaviour, not a failure.
- Prompt length against the recipe's `max_prompt_length: 8192`, measured over all 7,674 train records: **p50 1,583 · p95 4,949 · max 6,732** tokens (max 21 images in one record). **0 records overflow.** Worth stating explicitly, since both chat stages *raise* on overlong prompts rather than truncating — the converter's defaults do fit the recipe's declared budget.

Environment: Taiji 1×8 H20, box conda torch 2.7.1 / cu12.9, transformers 4.56, flash-attn 2.8.1, `pip install --no-deps -e .` plus `opencv-python-headless`. The only override is `backend.fsdp_cfg.master_dtype=null`, required by that box's torch 2.7 (pre-2.11 fsdp2 asserts one dtype per FSDP group); the recipe's fp32 LoRA masters target the pinned torch 2.11.

</details>

`sft/qwen_vl_agent_sft` is running now at its shipped defaults; its curve follows in this thread.

haonan3:
## Training curve — Qwen2.5-VL, full-parameter

Companion to the BAGEL LoRA curve above. `sft/qwen_vl_agent_sft` at its **shipped defaults** (`num_steps: 300`, `batch_size: 32`, `eval_interval: 20`), same manifest, same held-out split. This one trains the **full 7B LLM** (ViT frozen, no LoRA), so the two curves together cover both training modes the agent SFT path supports.

### Qwen2.5-VL-7B — qwen_vl_agent_sft (full-parameter LLM, ViT frozen)

300 steps logged (`300/300`), 1.25 epochs; train loss 2.3328 → 1.4141 (5-step mean 1.3767).
Held-out eval loss 2.4524 → 1.4931 over 16 evaluations.

```mermaid
xychart-beta
    title "Qwen2.5-VL-7B — qwen_vl_agent_sft (full-parameter LLM, ViT frozen) — train (smoothed) vs held-out eval loss"
    x-axis "step" [0, 10, 20, 30, 40, 50, 60, 70, 80, 90, 100, 110, 120, 130, 140, 150, 160, 170, 180, 190, 200, 210, 220, 230, 240, 250, 260, 270, 280, 290, 300]
    y-axis "loss" 1.213 --> 2.565
    line [2.3328, 2.1753, 1.7554, 1.5966, 1.6051, 1.5285, 1.4872, 1.4930, 1.5654, 1.4463, 1.4324, 1.4608, 1.4251, 1.4502, 1.4551, 1.4261, 1.4627, 1.4545, 1.4424, 1.4390, 1.3567, 1.3883, 1.4148, 1.4126, 1.3815, 1.3442, 1.3650, 1.3930, 1.3254, 1.3514, 1.3767]
    line [2.4524, 2.4524, 1.7856, 1.7856, 1.6240, 1.6240, 1.5741, 1.5741, 1.5410, 1.5410, 1.5246, 1.5246, 1.5112, 1.5112, 1.5035, 1.5035, 1.4992, 1.4992, 1.4958, 1.4958, 1.4942, 1.4942, 1.4929, 1.4929, 1.4935, 1.4935, 1.4930, 1.4930, 1.4928, 1.4928, 1.4931]
```

| step | epoch | train loss | eval loss | grad_norm | lr |
| ---: | ---: | ---: | ---: | ---: | ---: |
| 1 | 0.00 | 2.3328 | — | 10.562 | 1.00e-06 |
| 20 | 0.08 | 1.6959 | 1.7856 | 4.750 | 9.97e-06 |
| 40 | 0.17 | 1.6684 | 1.6240 | 4.875 | 9.74e-06 |
| 60 | 0.25 | 1.5256 | 1.5741 | 4.281 | 9.28e-06 |
| 80 | 0.33 | 1.6379 | 1.5410 | 4.500 | 8.63e-06 |
| 100 | 0.42 | 1.3856 | 1.5246 | 4.125 | 7.81e-06 |
| 120 | 0.50 | 1.3402 | 1.5112 | 4.750 | 6.85e-06 |
| 140 | 0.58 | 1.5091 | 1.5035 | 4.344 | 5.81e-06 |
| 160 | 0.67 | 1.4006 | 1.4992 | 4.281 | 4.73e-06 |
| 180 | 0.75 | 1.4949 | 1.4958 | 4.312 | 3.66e-06 |
| 200 | 0.83 | 1.4169 | 1.4942 | 4.469 | 2.66e-06 |
| 220 | 0.92 | 1.3795 | 1.4929 | 4.156 | 1.76e-06 |
| 240 | 1.00 | 1.4483 | 1.4935 | 4.312 | 1.02e-06 |
| 260 | 1.08 | 1.3708 | 1.4930 | 4.344 | 4.62e-07 |
| 280 | 1.17 | 1.3361 | 1.4928 | 4.219 | 1.17e-07 |
| 300 | 1.25 | 1.4141 | 1.4931 | 4.344 | 0.00e+00 |


Held-out eval falls monotonically 2.4524 → 1.4929 through step 220, then sits flat within ±0.0005 for the last 80 steps — saturated at 1.25 epochs on this manifest, not diverging. The eval curve never turns back up, so there is no overfitting to report even past one epoch.

### Cost of the two recipes

Measured over the runs above, same box (1×8 H20), same manifest:

| recipe | mode | batch | mean step | **per sample** | step-time spread |
| --- | --- | ---: | ---: | ---: | ---: |
| `qwen_vl_agent_sft` | full-parameter 7B | 32 | 11.6 s | **364 ms** | 1.6× |
| `bagel_agent_sft` | LoRA r64 | 16 | 20.2 s | **1260 ms** | 3.4× |

The full-parameter run is ~3.5× cheaper per sample than the LoRA one, which is worth stating because it inverts the usual expectation. Three structural reasons, none introduced by this PR:

- **BAGEL-7B-MoT is a mixture-of-transformers** — 28 GB of weights against Qwen's 16.6 GB, because it carries separate und and gen expert stacks. FSDP full-shard all-gathers every expert each step even though this path only runs the und experts.
- **Bagel stacks backend `activation_checkpointing: true`** on top of the packed forward; the Qwen recipe sets it `false` and relies on the bundle's HF gradient checkpointing alone.
- **Fewer micro-steps to hide the traffic behind** — `micro_batch_size: 1` means 16/8 = 2 micro-steps per device for Bagel against Qwen's 4, so the MoT all-gather is less overlapped.

The 3.4× step-time spread on Bagel (9.0 s – 30.5 s) is the same story from the data side: records carry 3–21 images, there is no length bucketing, and with only two micro-steps per device a single 21-image record sets the step time. Qwen's four micro-steps and more uniform processor packing keep it at 1.6×. Flagging rather than fixing — it is the recipe's cost structure, not a regression, and the `[vit,text]` byte-parity check in the Test Plan shows the non-interleaved packing is unchanged.

leviking98z-rgb:
Blocking finding: Qwen-VL silently drops `tool_calls` from accepted agent-SFT records.

The new [`QwenVLChatTemplateStage.embed_messages`](https://github.com/Tencent-Hunyuan/UniRL/blob/8ae7799d27ca2a13fa5d8cbc92d4518af6e1437e/unirl/models/qwen_vl/chat_template.py#L90-L125) accepts `tools` and agent histories without rejecting tool-call shapes. The stage also does not implement `tokenize_agent_target`, so target serialization falls back to the generic HF chat-template suffix path.

However, the default recipe uses `Qwen/Qwen2.5-VL-7B-Instruct`, whose official [processor chat template](https://huggingface.co/Qwen/Qwen2.5-VL-7B-Instruct/blob/main/chat_template.json) does not reference either `tools` or `message.tool_calls`. I reproduced this with the official template: an assistant target containing both text and `tool_calls` renders byte-for-byte identically after removing `tool_calls`. The result is that only the text is supervised and the structured call is silently lost.

This violates the invariant documented in this PR that a stage must render the whole target or raise. SearchGen itself does not contain tool calls, so the provided training runs are unaffected, but the shared agent-SFT contract explicitly accepts these records and the PR states that capability remains intact.

Please either:

1. provide a tool-capable Qwen-VL template plus stage-owned `tokenize_agent_target`, or
2. reject non-empty `tools`, `role="tool"`, and any history/target `tool_calls`, as the Bagel stage does.

A regression test should verify that a target such as `{"role":"assistant","content":"I will check.","tool_calls":[...]}` is fully serialized or rejected, never reduced to its text.

zzhuoxin1508:
`prepare_sft.py` may misalign reference indices. Some real selections lack `query_id` but can be recovered via `query_index`; the current code drops them and renumbers the remaining selections, causing `Reference Image N` in the refined prompt to point to the wrong image. Please use `s4_generation_manifest.ordered_references` as the canonical order, or recover `query_id` from `queries[query_index]` instead of filtering and compressing the indices.

zzhuoxin1508:
>  found two remaining issues:
>
> 1. The S3 task is inconsistent: some selected references have `used_in_refinement=false` and are not cited by the source `refined_prompt`, but the converter requires `Reference Image N` citations and appends `Selected references: [...]`. Please either filter/normalize these traces or relax the instruction.
> 2. sft.py accepts content=null for system/user messages, but the pinned Transformers 5.6 Qwen-VL chat template attempts to iterate that value and raises TypeError

leviking98z-rgb (approved):

PR #360, opened 2026-08-12
fix(drpo): reconcile reward docs with the recipe

haonan3:
## Summary

Reconcile the DRPO reward documentation with the shipped recipe and the paper.

- Document `MathVerifyRewardScorer` as the rule-based grader used by the DRPO DAPO-Math recipe.
- Replace stale recipe comments that compared the configured grader to the unused `MathBoxedRewardScorer` class with behavior-based wording.
- Fix the debug checklist row that still framed the failure mode around boxed answers. Under `math-verify` a missing `\boxed{}` is not a failure by itself, and the row's `/no_think` + `enable_thinking: false` remedy contradicted the recipe it documents (which sets `enable_thinking: true` in both the pipeline and rollout chat templates under a MUST-match comment). Reworded to the real symptom — the completion never reaches a final answer — with truncation as the first thing to check.
- Describe `math-verify` extraction precisely: recognized final-answer markers take priority; without one, it may fall back to the rightmost parseable expression. With `truncated_reward: keep`, a trace truncated mid-derivation can therefore score 1.0 off an intermediate value.
- Remove the unused `math_boxed.py` and `gsm8k_exact_match.py` scorers after confirming they have no in-tree callers or registry entries.

## Related Issue

N/A

## Test Plan

- `SKIP=no-commit-to-branch pre-commit run --all-files` on the branch with current `main` merged in — all hooks passed, including `check-recipe-targets`, `check-core-dependencies`, and `check-docstring-lines`.
- YAML parse-equality harness: `yaml.safe_load` of each of the 7 touched recipes before vs. after → identical for all 7; each parsed tree's `reward.backend._target_` is `MathVerifyRewardScorer`.
- Verified extraction behavior against both the minimum supported `math-verify==0.6.0` and current `0.9.0`: boxed and anchored unboxed `42` score true; an earlier anchored final answer outranks a later expression; without an anchor the parser can select the rightmost expression; a trace truncated at a matching intermediate value can score true.
- Verified the ground-truth plumbing: `datasets/dapo_math/prepare_dapo_math.py` writes `{"prompt", "metadata": {"answer": ...}}` and `MathVerifyRewardScorer` reads `meta["answer"]`.
- `git grep -i 'math_boxed|MathBoxed|gsm8k_exact_match|GSM8KExactMatch'` repo-wide → zero hits after removing the two unused modules.

## Compatibility / Risk

The 7 recipe edits are comment-only and parse-identical. Removing the two unregistered scorers has no in-tree runtime impact; an out-of-tree config that directly imports either old module path will need to migrate to `MathVerifyRewardScorer`.

## Reviewer Notes

AI-assisted; duplicate-work check done. The final diff removes the two orphan scorers rather than creating an unevidenced compatibility promise for hypothetical out-of-tree users.

Changed since the last review round: corrected the debug-checklist row and parser semantics, documented the truncation caveat, removed the orphan scorers, and merged current `main`.

## Checklist


CjhHa1 (approved):
LGTM

PR #361, opened 2026-08-12
refactor: clarify and enforce core ownership boundaries

haonan3:
## Summary

First low-risk implementation batch from the UniRL architecture audit, focused on clear ownership and dependency direction rather than broad file movement.

- Move the post-materialization callback contract from `train/deferred.py` to `models/types/post_materialize.py`. The callbacks describe *model-construction* state (register work while the module is still on meta; drain it after the weight load), not training policy — and two model bundles (`qwen3`, `flux2_klein`) were reaching up into `unirl.train` to get it. `_stamp` is renamed `defer_after_materialize` at the new home. **No compatibility shim**: `train/deferred.py` is deleted and all five call sites are migrated, per the direction #366 set.
- Same rationale, same home: the AC-wrap name canonicalization moves out of `train/backend/sharded_state.py` to `models/types/post_materialize.py` as public `canonical_param_name`. Its non-train callers are model bundles matching param names against checkpoint keys in deferred/materialize paths — `flux2_klein` (left half-fixed by the first revision) and a previously undisclosed identical reach-in in `hunyuan_image3`. The old private symbol is deleted, all call sites migrated (`sharded_load`, both bundles), and the constraint prose is now a `models/README.md` gotcha instead of a comment block.
- Move `tokenize_agent_target` from `unirl/data/sft.py` to `models/types/conversations.py`. It is pure tokenizer mechanics (record messages → final-assistant-turn token suffix) with no data-layer dependency, and its `qwen_vl` chat-template consumer (#352) made `models/` import `unirl.data` — the one edge that failed the new guard on the merged tree. All consumers (`qwen_vl`, `train/sft`, the `datasets/` CLI) migrate; no re-export left behind.
- Move the duplicated DTensor `local_view` (byte-identical copies in `train/ema.py` and `train/backend/sharded_state.py`, with `algorithms/bagel_flow_unigrpo.py` importing the EMA copy) to `unirl/distributed/local.py` — deliberately *not* `distributed/tensor/`, whose `__init__` eagerly hoists `worker_local` (module-level `import ray`): parking it there, as the first revision did, pulled 177 `ray` modules (~1.5 s) into previously torch-only train imports. `distributed/__init__.py` is the package whose stated contract is that a leaf import stays a leaf; after the move the same imports load **0** `ray` modules. `sharded_state.drop_meta_entries` now calls the helper instead of re-inlining the unwrap, and the stale `local_view` `__all__` re-export is gone.
- Make train stacks import `unirl.algorithms.base` contracts — and make that mean something at runtime: the package root now resolves concrete algorithms lazily (PEP 562 `__getattr__`/`__dir__`), so a base-contract import loads exactly `unirl.algorithms` + `unirl.algorithms.base` (the eager root previously pulled all 13 concrete algorithm modules regardless, making the `.base` switch a no-op). The package API is unchanged: `unirl.algorithms.GRPO`, Hydra `_target_` dotpaths, and `__all__` still resolve; no in-repo consumer imports concrete symbols via the root (grep-verified).
- Add `lint/check_core_dependencies.py`, a narrow AST dependency guard wired into pre-commit (`files: \.py$`, same gating rationale as `check-docstring-lines`). It locks only directions the current tree already satisfies — now including `train → algorithms/base`-only and `utils/` / `config/` never importing the loop tier — and deliberately does **not** hide known VeOmni/EP debt behind allowlists. `TYPE_CHECKING`-only imports are exempt (no runtime coupling; else-branches and function-local imports stay checked); bare `import unirl` / `from unirl import *` are rejected in ruled dirs (attribute-chain access through the shared package object evades prefix rules); a syntax-error file fails `ruff`/`check-ast` instead of crashing this hook; every `allowed` entry is a true sub-prefix of its owner's `forbidden` tuple, so each visibly shields something.
- Document the module taxonomy, dependency direction, and a concrete "where new code goes" table in `unirl/README.md` — stating exactly which directions the guard locks, and naming the unlocked debt (VeOmni/EP reach-ins from `models/` and `distributed/weight_sync/`, `cosmos3`'s train-side track-builder subclass, `algorithms`' `train/lora` adapter flips) so `ok` is not read as more protection than it provides. The dataset-CLI row points at the real repo-root `datasets/<dataset>/` (post-#375 layout), not a nonexistent `unirl/datasets/` path. The post-materialize contract and canonical-name constraint are `unirl/models/README.md` gotchas.

## Related Issue

N/A

## Test Plan

- `SKIP=no-commit-to-branch pre-commit run --all-files` on the branch with current `main` (`cc6cd395`, #375) merged in → all hooks Passed, including the new `check-core-dependencies`.
- `python3 lint/check_core_dependencies.py` → `ok` on the merged tree. The first revision failed exactly here once #352 landed on `main` (`models imports 'unirl.data.sft.tokenize_agent_target'`); fixed by the `conversations.py` move, not by a carve-out.
- **Dependency-guard harness** (one-off, uncommitted per `CLAUDE.md`): 24 synthetic modules planted under `unirl/` one at a time, guard run, module removed, verdict compared to expectation — the original 11 spelling cases (absolute/relative/star/alias, contract carve-outs, sibling self-import) plus: train→concrete-algorithm and `from unirl import algorithms` rejected while `algorithms.base` passes; `TYPE_CHECKING` skip in both `Name` and `typing.`-attribute forms, with the `else:` branch and function-local imports still checked; bare `import unirl` and `from unirl import *` rejected; live `utils/`→train and `config/`→models rejections with `config/`→utils accepted; a syntax-error file tolerated without a traceback. `RESULT: all cases behave as expected`; clean tree `exit=0`.
- Import-graph measurements (fresh interpreters): `import unirl.train.backend.sharded_state` and `import unirl.train.ema` → **0** `ray*` modules in `sys.modules` (first revision: 177, ~1.5 s extra). `from unirl.algorithms.base import StageAlgorithm` → loaded `unirl.algorithms*` modules = `['unirl.algorithms', 'unirl.algorithms.base']` (first revision: all 14). `ray` still loads under `algorithms.base` via its own `distributed.group.remote` dependency — pre-existing and intrinsic to the contract, unchanged here.
- Lazy-API checks: `from unirl.algorithms import FlowGRPO, GRPOConfig` works and loads only those modules; `importlib.import_module("unirl.algorithms.grpo")` (the Hydra `_target_` path) works; `__all__ ⊆ dir()`; unknown attribute raises `AttributeError`.
- Import smoke on every touched module (`models/types/conversations`, `models/types/post_materialize`, `data/sft`, `train/sft/track_builder`, `train/ema`, `train/backend/sharded_state`, `train/backend/sharded_load`, `train/backend/fsdp/backend`, `distributed/local`, `algorithms/bagel_flow_unigrpo`, `models/flux2_klein.bundle`, `models/qwen3.bundle`, `models/hunyuan_image3.bundle`, `models/qwen_vl.chat_template`), plus a stub-tokenizer functional test of the moved `tokenize_agent_target` covering the suffix-contract error path and the EOS-trim path.
- `git grep` for `train.deferred`, `tensor.local` / `tensor/local`, `_canonical_param_name`, and `data.sft import tokenize` → **zero** hits.

## Compatibility / Risk

Repo-internal API moves with no aliases left behind (#366): `unirl.train.deferred` is gone; `tokenize_agent_target` no longer exports from `unirl.data.sft`; `_canonical_param_name` is deleted in favor of `unirl.models.types.post_materialize.canonical_param_name`; `local_view` lives at `unirl.distributed.local`. All in-tree callers are migrated; none of these were public out-of-tree surface. `unirl.algorithms`' attribute surface, `__all__`, and recipe `_target_` dotpaths are byte-identical in behavior (PEP 562). The guard is stricter than the first revision (train→algorithms lock, `utils`/`config` rules, bare-import rejection) but passes on the current tree; `TYPE_CHECKING`-only imports are exempt by design.

## Reviewer Notes

Part of the repo architecture audit. AI-assisted (Claude Code); duplicate-work check done. History is merge-based on current `main` (the #375 rewrite of `unirl/utils/README.md` is resolved in the merge commit), then one commit per concern.

Changed since the second revision (review round):

- The first `check-core-dependencies` went red the moment #352 landed on `main` — and with `always_run: true` plus `--all-files` lint CI, merging it would have blocked every subsequent commit repo-wide. Fixed at the root (the tokenizer helper belongs to the contract layer both sides may import), and the hook is now extension-gated.
- `distributed/tensor/local.py` regressed train import graphs onto `ray` (measured); the helper now lives at `distributed/local.py`.
- The `.base` import switch was a runtime no-op behind the eager package root; the root is now lazy, so the isolation this PR claims is measured, not aspirational.
- The guard's advertised coverage and its `RULES` now match: train→algorithms locked with the `base` carve-out, `utils`/`config` rules added, dead self-permit carve-outs dropped, `TYPE_CHECKING` exempted, the bare-`import unirl` hole closed, and `unirl/README.md` names the unlocked debt — including `hunyuan_image3`'s previously undisclosed `_canonical_param_name` reach-in, which is migrated rather than listed.

Still intentionally out of scope, now named in `unirl/README.md` so they are not copied as precedent: VeOmni/EP model backedges, `cosmos3`'s track-builder subclass, `algorithms`↔`train/lora` adapter ownership, and SDE scheduler placement.

## Checklist


CjhHa1 (approved):
Reviewed the ownership-boundary refactor, lazy exports, dependency guard, and moved contracts. The targeted checks and full hook suite pass.

PR #174, opened 2026-07-03
feat(qwen3-5): support text and single-image inputs in AR training

NancyFyong:
## Maintainer Update (2026-08-01)

Merged current `main` (`73a3c58e`, including the TP/EP full-weight-sync support from #226) into the contributor branch and completed a correctness-focused review.

The follow-up fixes align SGLang rollout with train-side generation/replay, including:

- exact text/VL chat-template parity;
- exclusion of image/video placeholder IDs from both sampling distributions;
- Transformers 5.6 cached decoding and Qwen3.5 M-RoPE position handling;
- VeOmni 0.1.11 Qwen3.5-MoE selective registration and its no-cache GDN generation path;
- sharded loading of checkpoints that contain an auxiliary MTP head absent from the training model;
- main-supported VeOmni EP8 + SGLang TP8/EP8 recipe wiring.

All repository pre-commit hooks, all four Hydra recipe compositions, and out-of-tree real-model smoke checks pass. No test file was added.

## Summary

Add focused Qwen3.5 AR support: dense/MoE bundles, a text/single-image AR pipeline and stage, conditions/chat-template plumbing, four SGLang GRPO recipes, and Qwen3.5-only fail-fast runtime guards for supported FSDP2/VeOmni + SGLang paths.

## Verification Status

The long-running results below were collected on the split rollout/weight-sync integration stack before #226 merged. #226 is now on `main`, this branch has merged that exact main support, and the updated TP8/EP8 recipes use the merged interfaces. The long jobs were not repeated after this rebase; current-code numerical closure and loading paths were revalidated with tiny dense, MoE, and VL checkpoints.

| Recipe | Model | Hardware | Status | Notes |
|---|---|---|---|---|
| `qwen3_5_grpo_text_sglang` | Qwen3.5-9B (dense) | 8×H20 | ✅ Verified | Full GRPO, 161/800 rollouts, DAPO-Math-17k |
| `qwen3_5_grpo_vl_sglang` | Qwen3.5-9B (dense) | 8×H20 | ✅ Verified | Geometry3K image-to-text |
| `qwen3_5_moe_grpo_text_sglang` | Qwen3.5-MoE-35B-A3B | 8×H20 | ✅ Verified | Colocated SGLang TP8/EP8 + VeOmni EP8; train reward 0.7988 @ rollout 78, best 0.8555; eval avg@16 28.23% @ rollout 70 |
[...]

PR #364, opened 2026-08-12
fix(trainer): share model-blind AR preflight with async trainer

haonan3:
## Summary

Unify AR trainer pre-flight checks behind one model-blind helper and fix the async trainer's missing input-primitive initialization.

- Add `ar_preflight()` and call it from both `ARTrainer` and `AsyncARTrainer` before actors are created.
- Discover optional model-specific validation through `unirl.models.<package>.validation.validate_training_contract`, removing the trainer's direct dependency on Qwen3.5.
- Rename the Qwen3.5 validator to that generic hook name while keeping its validation behavior unchanged.
- Let pipeline classes declare additional accepted prompt primitives. `Qwen3OmniPipeline` declares `("media",)`, replacing the trainer's `"qwen3_omni"` target-string special case.
- Initialize `_allowed_input_primitives` in `AsyncARTrainer`. It intentionally skips `ARTrainer.__init__` but inherits `_build_request_sample`, so the previous code raised `AttributeError` on the first async dispatch.

Stacked on #357; retarget this PR to `main` after #357 merges.

## Related Issues

Follow-up to #174 and #306; no issue is closed directly.

## Test Plan

- Smoke-tested `ar_preflight()` with the shipped Qwen3, Qwen-VL, and Qwen3-Omni `from_bundle` targets.
  - Standard AR pipelines resolve `text`, `image`, and `video`.
  - Qwen3-Omni additionally resolves `media`.
- Verified all shipped Qwen3-Omni AR recipes use `Qwen3OmniPipeline.from_bundle`, whose bound classmethod resolves back to the declaring pipeline class.
- Verified Qwen3.5 dispatches through `validate_training_contract` and that the removed validator names have no remaining repository consumers.
- GitHub lint and CodeQL checks pass.

## Compatibility / Risk

- No recipe changes; accepted input primitives are unchanged for every shipped AR pipeline.
- Qwen3-Omni keeps its existing `media` input behavior, now declared by the pipeline rather than inferred from its target string.
- The Qwen3.5 validator's old internal names are removed. Out-of-tree imports must use `validate_training_contract`.
- Pre-flight now resolves the configured pipeline target on the driver. Invalid targets fail before GPU actors are allocated.

## Reviewer Notes

- `AsyncARTrainer` deliberately bypasses `ARTrainer.__init__` for its two-slab placement, so both constructors must call the shared pre-flight helper explicitly.
- Hydra targets may name either a pipeline class or a bound classmethod such as `from_bundle`; the helper uses the same `get_object`/`__self__` resolution pattern as the diffusion trainer.

## Checklist

### DIFF
--- a/unirl/models/qwen3_5/validation.py
+++ b/unirl/models/qwen3_5/validation.py
@@ -50,12 +50,12 @@ def _require_version(distribution: str, minimum: Version, *, reason: str) -> Non
         raise RuntimeError(f"Qwen3.5 {reason} requires {distribution}>={minimum} (installed: {installed}).")
 
 
-def is_qwen3_5_pipeline_config(pipeline_cfg: Any) -> bool:
+def _is_qwen3_5_pipeline_config(pipeline_cfg: Any) -> bool:
     """Whether a Hydra pipeline config targets the Qwen3.5 package."""
     return "unirl.models.qwen3_5." in _target(pipeline_cfg)
 
 
-def validate_qwen3_5_training_contract(
+def validate_training_contract(
     *,
     pipeline_cfg: Any,
     backend_cfg: Any,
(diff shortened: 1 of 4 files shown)


CjhHa1 (approved):
LGTM

PR #370, opened 2026-08-13
feat(vllm-omni): offload the MoE workspace pool on sleep

YSunLIN:
# feat(vllm-omni): offload the MoE workspace pool on sleep

### Motivation

In colocate layouts, the rollout engine and FSDP trainer share GPUs. vLLM's reusable MoE `WorkspaceManager` scratch buffer grows to the largest fused-expert GEMM it sees, but it is allocated outside the sleep-managed CuMem pools. Large MoE runs can therefore retain several GiB of VRAM through the train phase and push the trainer into OOM.

### Positioning

The fix enrolls workspace allocations in vLLM's existing CuMem machinery. It remains a quarantined, idempotent compatibility patch under `unirl/rollout/engine/vllm_omni/patches/`, with no config, trainer, checkpoint, or data-format changes.

### Design

`compat_moe_workspace.py` patches `WorkspaceManager._ensure_workspace_size`:

1. Each DBO ubatch owns an independently replaceable CuMem pool key (`moe_workspace:<ubatch_id>`), so one slot cannot replace another slot's live pool.
2. Allocation callbacks use the shared logical tag `moe_workspace` while the pool dictionary keeps the per-ubatch key.
3. Before growth replaces a pool, the patch drains CUDA work, drops the old workspace, and releases the retained `MemPool` before its `CUDAPluggableAllocator`. This ordering avoids the Torch 2.11 MemPool abort caused by destroying a live pool or finalizing the allocator first.
4. UniRL's worker-level `sleep(level=1)` discards the workspace with the other non-weight CuMem allocations; its `wake(tags=None)` remaps every allocation. No executor sleep bookkeeping is needed.

When sleep mode is off, `CuMemAllocator.instance` is `None` and the original workspace method runs unchanged. Repeat installs are sentinel-guarded, and import/API drift falls back to stock behavior.

## Test Plan

- CPU-import parity: importing `compat_moe_workspace` does not import vLLM or Torch; a vLLM-less install call returns without error.
- H20 runtime on the pinned stack (`torch 2.11.0+cu129`, `vllm 0.20.0`, `vllm-omni 0.20.0`):
  - two DBO slots grew from 1→2 GiB and 1.5→3 GiB;
  - two consecutive sleep/wake cycles each freed and remapped 5.00 GiB;
  - both workspaces remained writable after every wake.
- Repository pre-push suite, Ruff, format, AST, and one-line-docstring checks pass.

## Compatibility / Risk

- **Behavior:** changes only workspace allocations when the CuMem sleep allocator exists.
- **Scope:** rollout-engine workers only; no trainer, config, checkpoint, or YAML changes.
- **Internals:** depends on `WorkspaceManager`, `CuMemAllocator`, and `dbo_current_ubatch_id` from the pinned vLLM runtime.
- **Retirement:** delete the patch once vLLM allocates the reusable workspace in a sleep-managed pool itself.

## Reviewer Notes

Review the workspace growth lifecycle first: the old tensor must be released before replacing its per-ubatch pool, and the retained `MemPool` must be finalized while its pluggable allocator is still alive. The device-wide synchronization is intentional at vLLM 0.20 because its CuMem free callback can unmap memory immediately.


CjhHa1 (approved):
LGTM

PR #375, opened 2026-08-18
refactor(datasets): move offline data prep to datasets/, shrink utils to real leaves

haonan3:
## Summary

Second implementation batch from the UniRL architecture audit, and the migration
[#361](https://github.com/Tencent-Hunyuan/UniRL/pull/361) explicitly deferred: it wrote the
rules (`One-off dataset download/conversion CLI → datasets/<dataset>/`, "`utils/` is only for
stable, domain-agnostic leaves", "older modules … remain staged migration candidates") and this
PR makes the tree obey them. No new convention is introduced.

**Offline data prep leaves the package.** The nine `prepare_*.py` CLIs in `unirl/utils` (1314
LOC, 30% of that directory, 9 of its 25 files, and unmentioned in its own README) had no runtime
role — they are one-shot converters. They move to `datasets/<dataset>/`, joining the six
converters already there. Dataset-specific ones get a folder each; the four generic SFT manifest
builders are parameterized by `--dataset` rather than tied to one source, so they share
`datasets/sft_manifests/`.

**The drift had a mechanical cause, and it is fixed here.** `.gitignore` ignored `datasets/*`
behind a per-dataset allowlist, so `git check-ignore` said a new `datasets/<name>/convert.py` —
or the `datasets/tools/` some proposals suggested — was **ignored**. Adding a dataset meant
editing `.gitignore` first; `unirl/utils` was the only zero-friction destination. The rule is now
keyed on **file type** rather than dataset name (@CjhHa1's `7d33840`, replacing a first attempt
of mine that keyed on output-directory convention and left two documented cook targets addable):
code and guides under `datasets/` are tracked, everything else is ignored, the five committed
prompt sets are exempt by name, and `raw/`/`processed/` are re-denied last so the code
re-includes cannot reach inside a snapshot. A new `datasets/<name>/` still needs no entry, and
every currently-tracked file stays tracked.

**The move also improves the failure mode** (corrected — an earlier revision of this description
claimed the old invocation was broken; it was not, thanks to @CjhHa1 for catching it). Run from
the repo root, `python -m unirl.utils.prepare_*` put the repo root on `sys.path`, where a
`datasets/` directory without `__init__.py` is only a PEP 420 namespace *portion*: Python records
it and keeps scanning, and an installed regular package wins. Verified against a synthetic
site-packages with the repo root placed **first** on `sys.path` — the real library still resolves
and `load_dataset` works. So with HF `datasets` installed the old form was fine.

What the old form did produce, when the library was **missing**, was the namespace portion's
confusing error — `ImportError: cannot import name 'load_dataset' from 'datasets' (unknown
location)` — instead of a plain `ModuleNotFoundError: No module named 'datasets'`. The by-path
invocation gives the plain one. That is a real readability win but not a functional fix; the
actual defect was the undeclared dependency, addressed below. The move stands on organizational
grounds regardless.

**`utils/` gets a real boundary.** `misc.py` mixed dynamic import, seeding, logging, memory and
metrics — but four of its six functions were dead repo-wide (`set_seed`, `configure_logger`,
`clear_memory`, `flatten_dict`; the last one's only reference was its own recursion), so they are
deleted rather than resplit. Dropping `set_seed` loses no reproducibility: the tree seeds
explicitly per `torch.Generator` (`sde/noise.py`, `data/data_source.py`,
`models/hunyuan_image3/seed.py`, `train/backend/base_backend.py`), which is the correct pattern
across ranks and actors; the global-state version it superseded was never called. `load_function`
had one consumer and moves into `config/validation.py` (both of its function-local imports become
top-level — the deferral guarded nothing: `misc.py` imports only numpy and torch, `validation.py`
already imports torch at top level, and there is no cycle). `aggregate_numeric_metrics` has five
call sites across three top-level directories, so it stays a leaf under a name that says what it
is (`utils/metrics.py`). `misc.py` is deleted.

`utils/hydra.py` was the only thing making `utils/` depend on `unirl.distributed` — two lines
wrapping `remote()`. All nine importers are trainers (eight `unirl/trainer/*`, one
`experimental/refl`), so it moves to `unirl/trainer/hydra.py`. That removes `utils/`'s only
dependency on `unirl.distributed`. To be precise about what remains (per @CjhHa1): the sole
runtime intra-`unirl` import left in `utils/` is `wandb_metrics.py:92` →
`unirl.types.sampling`, which points downward at a foundation layer; `wandb_logger.py`'s
`unirl.train.stack` import is `TYPE_CHECKING`-only. Whether `unirl.types` counts is a decision
for the eventual `utils` rule in `lint/check_core_dependencies.py`.

`unirl/utils/` goes from 25 modules to 15.

**The converters' dependencies are now declared** (review follow-up). `datasets`, `pandas`,
`pyarrow` and `av` were needed by the converters but present in neither the base deps nor any
extra, so following a cook command in a clean checkout ended at `ModuleNotFoundError`. This is
pre-existing — the same scripts had the same undeclared deps in `unirl/utils` — but this is the
PR that writes the cook commands, so it fixes it: a `dataset-prep` extra, installed with
`pip install -e '.[dataset-prep]'`. Optional rather than base because nothing under `unirl/`
imports any of it; you install it to cook, then train without it. The `av` bound matches the
sglang/vllm extras so a combined install cannot conflict.

## Related Issue

N/A

## Test Plan

- `SKIP=no-commit-to-branch pre-commit run --all-files` → all hooks **Passed** (`ruff check`,
  `ruff format`, `check-recipe-targets`, `check-experimental-boundaries`, `check-docstring-lines`).
  Also clean over the PR range: `pre-commit run --from-ref origin/main --to-ref HEAD`.
- `python lint/check_docstring_lines.py` → `3210 docstrings, all one line.`
- **Import check** — 15 modules touched by the `hydra`/`metrics`/`load_function` moves imported
  clean (all eight trainers, `trainer.hydra`, `config.validation`, `utils.metrics`,
  `train.stack.base`, `train.unified_model_stack`, `algorithms.diffusionnft`,
  `utils.wandb_logger`); `load_function` and `aggregate_numeric_metrics` exercised directly.
- **Hydra compose** on four recipes whose comments changed:
  - `python -m unirl.train_ar --config-name ar/qwen_vl_grpo_geo3k_mc_4x8 --cfg job --resolve`
  - `python -m unirl.train_ar --config-name ar/qwen3_drpo_4b_base_dapo_sglang --cfg job --resolve`
  - `python -m unirl.train_sft --config-name sft/qwen3_sft --cfg job --resolve`
  - `python -m unirl.train_sft --config-name sft/cosmos3_droid100_videopred --cfg job --resolve`
- **Converters** — all nine `python -m py_compile`; `--help` run by the newly documented path for
  `geo3k_mc`, `dapo_math`, `sft_text`, `sft_t2i`.
- **Namespace-portion precedence** tested directly with a synthetic checkout + site-packages: repo
  root first on `sys.path`, a `datasets/` dir with no `__init__.py`, and a regular `datasets`
  package installed → resolves to the installed one and `load_dataset` works. The
  `(unknown location)` error appears only when the library is absent. This is what corrected the
  claim above.
- **`.gitignore`** verified with `git check-ignore` over **24 paths** after @CjhHa1's `7d33840`:
  converter code and guides in a brand-new `datasets/newds/` (including a nested subdir) are
  tracked; the five committed prompt sets stay tracked; and every cook target is ignored,
  including the two documented ones the first inversion had left addable
  (`datasets/video_r1_260k_image/`, `datasets/sft_agent_toolcall_12k/`) plus `*/images/`,
  `*/videos/`, `*/frames/` and the repo-root `data/`. Ordering checked: a `.py` or `.md` inside
  `raw/`/`processed/` stays ignored, so the code re-includes cannot reach into a snapshot.
  `git ls-files datasets | git check-ignore --stdin` → no tracked file became ignored.
- **`dataset-prep` extra** — `uv pip compile --prerelease=allow --extra dataset-prep pyproject.toml`
  resolves against the full base tree (125 packages: `datasets==5.0.1`, `pandas==3.0.5`,
  `pyarrow==25.0.1`, `av==18.1.0`). TOML parses and all four requirement strings are valid
  PEP 508. Dependency set derived by AST-scanning every converter, then corrected by hand for
  `droid100`, which resolves `pandas` and `av` through a string-based `_require()` helper an
  import scan cannot see.
- **Dangling references** — repo-wide grep for `unirl.utils.prepare_*`, `unirl.utils.hydra`,
  `unirl.utils.misc` and the four deleted symbols → zero hits. All 53 relative links in the new
  and edited READMEs resolve.
- Not run: any GPU training or rollout. No runtime code path changes behaviour — the moves are
  import-path only, the 49 recipe/doc edits are comments, and the deleted functions had no callers.

## Compatibility / Risk

- **Import paths (breaking for out-of-tree callers).** `unirl.utils.hydra` →
  `unirl.trainer.hydra`; `unirl.utils.misc.aggregate_numeric_metrics` →
  `unirl.utils.metrics`; `unirl.utils.misc.load_function` → `unirl.config.validation`. No
  compatibility shims, per the direction #366 set — all in-tree call sites are migrated.
- **Command change.** `python -m unirl.utils.prepare_X …` → `python datasets/<name>/prepare_X.py …`.
  The old form no longer exists; it was also the broken one. Converters that import `unirl`
  (`prepare_sft_agent`, and the pre-existing `searchgen`) need the package installed, as before.
- **Packaging.** The nine converters no longer ship in the wheel. They are offline tools invoked
  from a repo checkout; the six already under `datasets/` were never packaged either.
- **New optional extra** `dataset-prep` (`datasets`, `pandas`, `pyarrow`, `av`). Additive — no
  existing install changes, and no `[tool.uv] conflicts` entry is needed since it pins no torch
  or CUDA flavor. Cooking a dataset now requires installing it; previously the requirement was
  undeclared rather than absent.
- No config schema, checkpoint, data format, or resource change. Generated data locations are
  unchanged.

## Reviewer Notes

- **Stacks on #361, which is still open.** It rewrites the same paragraph of
  `unirl/utils/README.md` and the `utils/` row context in `unirl/README.md`, so expect a small
  conflict in those two files whichever lands second; the code changes do not overlap. Reviewing
  #361 first will read better, since it states the rules this PR applies.
- Suggested reading order: `datasets/README.md` (the boundary), `.gitignore` (the root cause),
  then the `utils/` commit. The commits are independent and separately reviewable.
- **`load_function` placement** — @CjhHa1 notes that `from unirl.config.validation import
  load_function` would read oddly for a second consumer, and that `utils/imports.py` fits the
  "stable, domain-agnostic leaf" rule. Kept in `config/` deliberately: that rule's other half is
  "a function used by one owner stays with that owner", and this has exactly one. Putting a
  single-consumer helper in `utils/` is the pattern this PR is unwinding. If a second consumer
  appears, promoting it to `utils/imports.py` is the right move then — and cheap.
- The 49 reference updates across 36 recipes/docs are comment-only and were applied
  mechanically, then verified by the dangling-reference grep and the compose checks above.
- Follow-up left deliberately out of scope: `wandb_logger.py` (758 LOC, one trainer importer),
  `memory_monitor.py`, `profiling.py`, `shard_balance.py` and `scheduler_utils.py` are all
  single-owner modules still sitting in `utils/`; and a `utils` entry in
  `lint/check_core_dependencies.py` is now possible but belongs with #361's guard, not here.
- AI assistance: written with Claude Code; every changed line reviewed, and the claims above
  (dead-code status, importer counts, the shadowing bug, the `check-ignore` behaviour) were each
  verified against the tree rather than assumed. Duplicate-work check: scanned all 42 open PRs
  for overlap on `datasets/`, `unirl/utils/` and `.gitignore` — only #361 touches them, as noted.

## Checklist


CjhHa1:
Reviewed this against the tree. The refactor itself checks out: the four deleted `misc.py` helpers really did have zero callers on `main`, the nine converter moves plus `hydra.py` are pure renames (97–100% similarity, only usage-path lines edited), the 49 recipe edits touch nothing but trailing comments on `data_path` lines, all 78 relative links in the changed READMEs resolve, and `pre-commit run --all-files` is green. Two things needed attention.

**1. The `.gitignore` inversion left documented cook targets writable to the index — pushed a fix as 7d33840.**

`datasets/*` used to blanket-ignore everything non-allowlisted. With only `**/raw/`, `**/processed/` and two jsonl globs ignored, anything a converter cooks elsewhere under `datasets/` became addable. Verified with `git check-ignore` on both branches:

| path | was | became |
|---|---|---|
| `datasets/video_r1_260k_image/` | ignored | addable — documented `--out-dir` at `datasets/video_r1_260k/README.md:107` |
| `datasets/sft_agent_toolcall_12k/` | ignored | addable — the `--out-dir` in `prepare_sft_agent.py`'s own usage block |
| `datasets/{image_edit,geo3k_mc,arxivqa_mc}/images/`, `datasets/droid100/videos/` | ignored | addable |

The Test Plan verified exactly the four patterns the PR adds, so it couldn't surface these. `check-added-large-files` at `--maxkb=2500` is a partial net — a cooked `train.jsonl` trips it, per-sample images don't.

Naming those directories individually would rebuild the allowlist this PR removes, so the fix keys the rule on file type instead: `datasets/**` denied, `!*.py` and `!*.md` re-included, the five committed prompt sets exempt by name, `raw/`/`processed/` re-denied last so the code re-includes don't reach inside a downloaded snapshot. Your invariant holds — a new `datasets/<name>/` with a converter and a README still needs no entry. The two per-dataset jsonl rules are now redundant and are gone. `prepare_sft_agent`'s usage now cooks into `data/`, matching the other converters and the convention `datasets/README.md` states. Checked 33 paths: every leak closed, no tracked file ignored, `git ls-files | git check-ignore --stdin` empty.

**2. The "live bug" is a missing dependency, not shadowing — worth correcting in the description.**

A directory without `__init__.py` is only a namespace *portion*: Python records it and keeps scanning `sys.path`, and a regular package in site-packages wins. So the local `datasets/` folder cannot shadow an installed `datasets`. Three cases:

```
HF datasets installed  + repo root on sys.path -> resolves to the real library, no error
HF datasets missing    + repo root on sys.path -> ImportError: cannot import name 'load_dataset'
                                                  from 'datasets' (unknown location)
HF datasets missing    + by-path invocation    -> ModuleNotFoundError: No module named 'datasets'
```

The second line is the error quoted in the description, and `(unknown location)` is the namespace-package signature — it only appears when the library isn't installed. So `python -m unirl.utils.prepare_*` was not broken; the real defect is the undeclared dependency, which this PR fixes with the `dataset-prep` extra. The move turns a confusing error into a clear one, which is a genuine improvement, just not a functional fix. Flagging it because the description cites it as a reason the move is necessary, and the move stands on organizational grounds without it.

**Smaller, non-blocking:**

- `load_function` in `config/validation.py` reads oddly for a generic dynamic-import helper — a second consumer would `from unirl.config.validation import load_function`. Colocating with the single consumer is defensible, but `utils/imports.py` fits the "stable, domain-agnostic leaf" rule you're writing.
- "`utils/` is a leaf again" holds for `unirl.distributed`, but `utils/wandb_metrics.py:92` still has a runtime function-local `from unirl.types.sampling import ARSamplingParams`. Fine as-is; the planned `check_core_dependencies.py` rule will need to decide whether `unirl.types` counts.

CjhHa1 (approved):
LGTM

PR #377, opened 2026-08-19
refactor(weight-sync): move EP expert-layout policy out of generic transport

haonan3:
## Summary

`unirl/README.md` names this edge as open debt: `distributed/weight_sync/` reaches into
`train/backend/veomni` for EP expert fusing. `FullWeightSync` imported one concrete model's
layout (`ep/models/qwen3_moe`), VeOmni's parallel state, and the `_compat` shim — so the
generic transport could only sync a model family it had been taught by name, and a new EP
model meant editing `distributed/`.

The layout stays at the train-backend seam — where model × backend integration belongs per
`unirl/README.md` ("integration code that must understand both a backend and a model belongs
at that integration seam … not in generic infrastructure") — and the **backend now hands the
transport a ready stream transform**:

- `BaseFSDP2Backend.expert_weight_export_transform()` → `None` (default: shards no fused experts).
- `VeOmniBackend` overrides it, resolving the layout through the new
  `train/backend/veomni/ep/experts.py` registry.
- `FullWeightSync` keeps only "the backend may transform my expert weights for export" — no VeOmni, no Qwen3, no
  parallel state, and no probing of `model._extra_parallel_param_groups`.

**Resolution is keyed by `config.model_type`, not by parameter-name suffix.** The old
`is_fused_expert_param()` matched HunyuanImage3's `.experts.down_proj` as well; HI3 only
failed closed because its *other* suffix (`gate_and_up_proj`) happened not to match — while
its gate/up halves are stored **swapped** relative to Qwen3-MoE's `gate_up`. A suffix match
would otherwise hand one model's packing semantics to another. Unregistered model types now
fail closed with a message naming the model type and the fix.

Both walks were duplicating the merged/raw selection and the `name_remap` loop, so
`_iter_full_tensors_ep` is gone and the single walk applies the `lora_A`/`lora_B` skip to
both branches. Under EP that skip was unreachable anyway: EP + `peft_config` already requires
`lora_merged=True` (unchanged, `base.py`), and `raw_state_dict` emits no adapter keys without
`peft_config`.

Generic infrastructure gets smaller (`full/base.py`: −90 / +16); the replacement policy lives behind an explicit, typed train-backend export contract.

Finally, the now-clean direction is **locked**: `check_core_dependencies` forbids
`distributed → unirl.train`, and `unirl/README.md` drops `distributed/weight_sync/` from the
debt list.

## Related Issue

N/A

## Test Plan

- `SKIP=no-commit-to-branch pre-commit run --files <the 9 changed files>` → **all hooks
  passed** (ruff check, ruff format, docstrings-are-one-line, core-dependency-directions,
  recipe `_target_` paths, experimental-tier boundaries).

- **EP export contract smoke:** synthetic models covered unknown model types, disabled EP state, fused Qwen splitting, and non-expert passthrough — passed.

- **Guard is live, not a no-op.** Re-adding the exact import this PR removes makes it fail:
  ```
  $ printf '\nfrom unirl.train.backend.veomni.ep.models.qwen3_moe import is_fused_expert_param\n' \
      >> unirl/distributed/weight_sync/full/base.py
  $ python3 lint/check_core_dependencies.py
  check-core-dependencies: FAILED
    unirl/distributed/weight_sync/full/base.py:146: distributed imports
    'unirl.train.backend.veomni.ep.models.qwen3_moe.is_fused_expert_param'; depend on a
    lower-level contract/mechanism or move the integration to its owner
  ```

- **Old-vs-new equivalence, 2-rank gloo on CPU** (uncommitted harness per repo policy; stubs
  `veomni.distributed.parallel_state`, replays the pre-refactor `_iter_full_tensors_ep` loop
  verbatim against the new export transform over the same synthetic Qwen3-MoE stream, with a
  non-trivial `name_remap`):
  ```
  $ PYTHONPATH=. python3 -m torch.distributed.run --nproc_per_node=2 verify_ep_equiv.py
  == EP weight-sync walk equivalence (world=2, gloo/CPU) ==
    ep_size=2 (all_gather path): 13 tensors, bit-identical.
    ep_size=1 (no-gather path): 7 tensors, bit-identical.
    expert ordering across EP ranks: rank0 experts 0-1, rank1 experts 2-3 ✓
    unknown model_type fails closed ✓ (EP experts: no fused expert layout is registered
      for model_type='hunyuan...)
    non-EP model -> export transform is None (plain walk) ✓
  ALL EQUIVALENCE CHECKS PASSED
  ```
  This covers the real `dist.all_gather` + expert-ordering path, so a wrong gather order or
  packing would show as a tensor mismatch.

- **Not run: multi-GPU EP training / real weight sync.** Reason: no GPUs on this jumpbox, and
  the live users (`examples/ar/qwen3_moe_grpo_30b_a3b_veomni_ep_sglang.yaml`,
  `examples/ar/qwen3_5_moe_grpo_35b_a3b_*.yaml` — `TensorWeightSync`, `ep_size: 8`) need a
  30B/35B-A3B checkpoint on ≥8 cards. Requesting a smoke run on those recipes before merge.

## Compatibility / Risk

- **No config, checkpoint, or wire-format change.** Emitted tensor names, order, and values
  are identical (verified above); recipes are untouched.
- **New backend method** `expert_weight_export_transform()` on `BaseFSDP2Backend`. Both in-tree
  backends (`FSDPBackend`, `VeOmniBackend` — the only two `_target_`s in `examples/`) inherit
  or override it; every sync handler receives a train backend via
  `remote_hydra(sync_cfg, backend=self.backend, ...)`, out-of-tree duck-typed backends must implement the method (returning `None` when no expert export transform is needed).
- **Behaviour delta, one unreachable corner:** the EP branch previously did *not* skip
  `.lora_A`/`.lora_B` in the raw (unmerged) walk; the unified walk does. As argued in the
  Summary this is unreachable under EP, and pushing adapter tensors under base-weight names
  was wrong where it wasn't.
- **Error-message change:** the "Qwen3-MoE fused expert layout only" `ValueError` is now
  raised from `resolve_expert_weight_export_transform` and names the offending `model_type`. Still a
  construction-time `ValueError` from `FullWeightSync.__init__`, so fail-fast timing is
  unchanged.
- **Additional fail-closed checks:** backend construction now rejects an `ep_size`/parameter-placement
  mismatch, while expert export rejects disabled EP state instead of treating a rank-local block
  as the complete expert set.
- HunyuanImage3's EP recipe syncs via `RemoteLoraWeightSync`, so it never enters this path.

## Reviewer Notes

- **AI assistance:** written with Claude Code; every line reviewed before publishing.
- **Duplicate-work check:** open PRs #369 (delta weight sync), #227, #225 reviewed. #369 also
  edits `full/base.py`, but only adds a `delta_sync` knob + `_resolve_delta_encoder` at the
  top of `__init__` — no overlap with the EP block or `_iter_full_tensors`. Textually
  adjacent, so a trivial conflict is possible; semantically independent.
- **Deliberately out of scope** (so the diff stays reviewable and to avoid touching the EP
  weight-*load* path, which needs multi-GPU verification):
  - `train/backend/sharded_load.py` still imports `build_local_fused_block` from
    `ep/models/qwen3_moe` directly. That is `train → train`, already at the seam, and legal
    under the guard.
  - `models/{hunyuan_image3,qwen3_5,qwen3_moe} → train/backend/veomni` remains, so
    `models → train` is still unlockable. `unirl/README.md` keeps that half of the debt named.
- Start with `unirl/train/backend/veomni/ep/experts.py` (the whole contract) and the
  `unirl/train/readme.md` Gotcha explaining why the registry is keyed by model identity.

## Checklist


CjhHa1 (approved):
LGTM

PR #378, opened 2026-08-19
feat(minimax-h3): vLLM-Omni rollout backend for H3 FlowGRPO at 32-GPU scale

nussejzz:
## Summary

vLLM-Omni rollout backend for MiniMax-H3 T2VA flow-GRPO, plus the 32-GPU
deployment it runs on. H3 is a 33B dense omni-modal transformer that denoises
video and stereo audio in one packed sequence, so it does not fit the existing
single-latent rollout path.

What this adds:

- **Rollout adapter and RL pipeline** for H3 under vLLM-Omni: grouped four-GPU
  replicas (DiT TP2 x Ulysses2, text-encoder TP4), sparse FP32 trajectory
  capture, and behaviour-policy log-probs emitted per SDE transition.
- **Staged LoRA publication.** `LocalLoraWeightSync` splits into
  `extract()` / `push()` so the adapter is extracted while the trainer is
  resident and pushed after trainer offload and rollout wake-up, which is what
  makes three roles time-share the same GPUs.
- **LoRA transport verification.** `add_lora()` only registers CPU-side logical
  weights, so a name or shape mismatch against the fused vllm DiT is silent:
  the engine keeps serving base weights and the only symptom is a flat reward
  curve. The Diffusers-to-serving translation now lives in one place
  (renames, the GEGLU half-swap Diffusers stores as `[up, gate]`, synthesized
  `rope.inv_freq`), every expected block/slice must carry both A and B, and the
  install fails closed if the adapter wrapped no physical layers. That last
  guard is what caught the namespace bug described under Test Plan.
- **Row-level reward dispatch** and a resident reward service at DP32. Reward
  now receives a frontier where each generated row is its own root, so its DP
  divisibility rule follows generated rows while rollout keeps the prompt-tree
  rule.
- **Numerical fixes.** Global CUDA autocast ran the fp32-pinned projections,
  timestep MLP and rope in bf16 despite their storage dtype; H3 conditions on
  `t = 1 - sigma`, where bf16 spacing near 1.0 is ~0.008, so this was real
  precision loss at the low-noise end.
- **Rollout contract fixes** found at 32-GPU scale: variable-length prompt
  embeddings are padded with a mask and trimmed before the packed forward
  (concatenating them directly raises on any batch whose prompts tokenize to
  different lengths, hit at 32 prompts spanning 9-58 tokens); SDE exploration
  keys come from the logical sample id rather than the engine request UUID, so a
  retried or resumed sample replays the trajectory it actually generated; sparse
  replay always stores the terminal latent for decode.
- **Continuation guarantees.** Resuming advanced the data source twice when both
  a checkpoint and `data_source.start_batch` were set; a checkpoint save that
  wrote no model state returned successfully.

## Related Issue

Builds on #347 (merged). `sync.verify` additionally needs
vllm-project/vllm-omni#6351; see Compatibility / Risk.

## Test Plan

Static, on this branch rebased onto `main` (`cdaf237`):

- `ruff check` + `ruff format --check` on all 22 changed Python files — clean.
- `python lint/check_docstring_lines.py`, `lint/check_recipe_targets.py`,
  `lint/check_core_dependencies.py`, `lint/check_experimental_boundaries.py` —
  all pass.
- Hydra composition of the changed recipes, e.g.
  `python -m unirl.train_diffusion --config-name diffusion/minimax_h3/minimax_h3_t2va_vllmomni_32c_8x8 --cfg job`.

On hardware — 4 nodes x 8 x H20 96GB, real MiniMax-H3 33B, LoRA rank 32/64,
256x448x107.

Against the `vllm-omni==0.27.0rc1` release this branch now pins (the
`vllm_omni-0.27.0rc1` wheel contents, no engine patches), the 8x8 gate recipe
completes both rollouts through the full lifecycle — LoRA sync, rollout,
reward, trainer replay/step, sleep — and exits 0:

```
rollout 1/2  reward=0.7673  loss=0.0000   gn=0.0000  ratio=1.0000±0.0000  |dlogp|=0.00e+00
rollout 2/2  reward=0.7342  loss=-0.0000  gn=0.0000  ratio=1.0000±0.0000  |dlogp|=0.00e+00
```

Zero loss and grad-norm are expected here rather than suspicious: GRPO
normalizes advantages within each prompt group, so at `ratio=1` the
policy-gradient term is identically zero. The reward difference between the two
rollouts is the data, not the update.

Reaching that took three fixes that only the pinned release exposes, each found
by running the gate rather than by inspection:

1. `boot()` read `intent["stage_yaml"]` unconditionally, but H3 boots from
   direct diffusion-stage arguments and never supplies that key.
2. The LoRA `target_modules` pattern was anchored one namespace level below the
   tensor keys it had to match. The adapter registered all 350 logical modules
   and wrapped zero physical layers, so the rollout kept serving base weights —
   exactly the silent failure the coverage guard exists for, and it fired.
   A development engine had hidden this by also trying the component-relative
   name.
3. `collective_rpc()` no longer takes `unique_reply_rank`, which broke the
   checksum readback used by `sync.verify` (default off).

Earlier evidence on the previous engine build, kept because it covers ground the
2-step gate does not: an 11-rollout run at 32 prompts x 8 samples with CLAP +
ImageBind held rollout/replay parity at `|dlogp| ~1e-5` against a `1e-3` gate,
and LoRA verification reported 350 modules per replica. Rollout topology at K8
(8 prompts x 8 samples): HSDP4+UP4 median generate 112.697s / 66,876 MiB peak;
TP2 x UP2 91.087s / 77,436 MiB, parity drift 3.73e-5 and 3.94e-5. A TP4 arm
raced a stale GPU keep-alive and OOM'd, so that number was discarded and
TP2 x UP2 is the fastest *verified* arm, not a proven optimum.

Not run:

- Regression of the existing BAGEL / HI3 / Qwen3-Omni vLLM-Omni recipes. The
  0.27.0rc1 / CUDA 13 floor arrived with #397 rather than here, but these
  recipes share the backend this PR edits, and I do not have their checkpoints.
  This is the main outstanding risk and needs an owner with those recipes.
- `sync.verify=true` end-to-end, which is blocked on the engine (below).

## Compatibility / Risk

- Dependencies are unchanged from `main`: this branch inherits the
  `vllm==0.27.0` / `vllm-omni==0.27.0rc1` / `torch==2.13.0+cu130` pins that
  #397 introduced, and adds no fork or git pins.
- **`sync.verify` cannot pass on the pinned engine.** Per-rank LoRA checksums
  need the engine to forward each rank's RPC reply; stock 0.27.0rc1 routes
  worker RPCs through a status-collecting path that keeps only rank 0's envelope
  and discards return values, so there is no way to read the shards back through
  the public API. It now fails closed with a message naming
  vllm-project/vllm-omni#6351 and the opt-out, instead of blaming the adapter.
  Recipes default to `verify: false`; training does not use this path.
- `LocalLoraWeightSync.sync()` is preserved as `extract()` + `push()`, so
  existing callers are unaffected.
- H3 geometry rejects anything off the released 768-short-edge distribution
  unless a recipe opts in with `sampler_kwargs.allow_nonstandard_canvas: true`.
- Reward DP divisibility now follows generated rows instead of prompt trees,
  because this PR makes reward score each row independently. Rollout keeps
  `main`'s prompt-tree rule, since rollout DP_SCATTER must preserve each
  prompt's whole subtree. The row rule is also strictly more permissive, so no
  existing recipe can start failing.

## Reviewer Notes

**Reward convergence is not solved and this PR does not claim it.** The
topology, lifecycle, LoRA sync and parity gates pass, and the throughput
numbers above are reproducible, but no H3 recipe here produces a sustained rise
in held-out visual reward. A Wan2.1 positive control on the same trainer,
PickScore and GRPO path does rise (`0.6985 -> 0.7201` over 20 rollouts), which
is what localizes the open problem to H3/reward rather than to the shared
training stack. Treat this as backend and deployment support, not as a tuned
recipe.

The flow-GRPO algorithm knobs and the experiment recipes are deliberately *not*
here; they are stacked on this branch in a follow-up so this PR stays reviewable
as backend support. A DiffusionNFT counterpart is planned separately.

AI assistance was used for this change.

## Checklist


MinhaoLi0318:
Reviewing this against current main and the vLLM-Omni side; will post a first pass shortly. If a split into smaller PRs (adapter/pipeline, LoRA transport, reward dispatch, trainer fixes) would help it land, I can help carry the rebase.

Issue #386, opened 2026-08-20
Reward collapse when using vLLM-omni on Hunyuanvideo1.5+Pickscore


yuxiaooye:
Hi UniRL team, 

I compared vLLM-omni and trainside on Hunyuanvideo1.5+Pickscore, and found that vLLM-omni leads to reward collapse (accompanied with blurry), but trainside is able to train steadily:

<img width="2584" height="1106" alt="Image" src="https://github.com/user-attachments/assets/c96177f1-67ec-4752-92e1-f58d23e64fd9" />

Could you please help checking what's the root cause? Does this issue relate to train-inference mismatch between vLLM-omni and FSDP engine? 

Also, I have noticed https://github.com/Tencent-Hunyuan/UniRL/pull/371, seems that SGlang is more stable than vLLM-omni?

The recipe of vLLM-omni are as follows:
```
#!/usr/bin/env bash
set -euo pipefail

# Run from the root of a current UniRL checkout, or set UNIRL_ROOT explicitly.
UNIRL_ROOT="${UNIRL_ROOT:-$PWD}"
LAUNCHER="${UNIRL_ROOT}/examples/run_experiment_single_node.sh"
if [[ ! -f "${LAUNCHER}" ]]; then
    echo "error: ${UNIRL_ROOT} is not a UniRL checkout (launcher not found)" >&2
    exit 2
fi

export PRETRAINED_MODEL="${PRETRAINED_MODEL:-hunyuanvideo-community/HunyuanVideo-1.5-Diffusers-480p_t2v}"
export DATA_PATH="${DATA_PATH:-${UNIRL_ROOT}/datasets/pickscore/train.txt}"
export EVAL_DATA_PATH="${EVAL_DATA_PATH:-${UNIRL_ROOT}/datasets/pickscore/test.txt}"

# Set WANDB_API_KEY in the environment (or run `wandb login`) before enabling W&B.
export WANDB_PROJECT="${WANDB_PROJECT:-unirl-reward-collapse-repro}"
export WANDB_RUN_NAME="${WANDB_RUN_NAME:-hyv15_pickscore_vllm-omniomni_collapse_repro}"
REPORT_TO_WANDB="${REPORT_TO_WANDB:-true}"

export TORCH_NCCL_BLOCKING_WAIT="${TORCH_NCCL_BLOCKING_WAIT:-1}"
[...]

PR #371, opened 2026-08-15
fix(hunyuan-video): align SGLang and trainside rollouts

leviking98z-rgb:
## Summary

HunyuanVideo's trainside and SGLang recipes used the same visible sampling values but did not execute the same numerical path:

- SGLang filtered rollout intent to `ServerArgs` fields before construction, so Hunyuan's pipeline-only `embedded_cfg_scale` and `flow_shift` were dropped and silently fell back to 6/7 instead of the recipe's 1/5. Route intent through SGLang's `ServerArgs.from_dict` so model `PipelineConfig` fields are preserved.
- Trainside sampling consumed rank-local SDE noise while SGLang derived it per sample and denoising step. Pass the same sample IDs and base seed through Hunyuan's trainside diffusion stage.
- Both bf16-loaded recipes now keep trainable LoRA masters in fp32, while forward compute remains bf16.

The existing `TensorWeightSync + lora_merged` path and `old_logp_source: rollout` recipe are intentionally preserved.

## Related Issue

N/A

## Test Plan

- `SKIP=no-commit-to-branch pre-commit run --files examples/diffusion/hunyuan_video/hunyuan_video_t2v_sglang.yaml examples/diffusion/hunyuan_video/hunyuan_video_t2v_trainside.yaml unirl/models/hunyuan_video/diffusion.py unirl/models/hunyuan_video/pipeline.py unirl/rollout/engine/sglang_diffusion/backends/native.py` — passed.
- `python lint/check_docstring_lines.py` and `python lint/check_recipe_targets.py` — passed.
- Hydra config composition on commit `0b7b0f0`:
  - `python -m unirl.train_diffusion --config-name=diffusion/hunyuan_video/hunyuan_video_t2v_sglang --cfg job --resolve` — passed; resolved `master_dtype=fp32`, `embedded_cfg_scale=1.0`, `flow_shift=5.0`, and `old_logp_source=rollout`.
  - `python -m unirl.train_diffusion --config-name=diffusion/hunyuan_video/hunyuan_video_t2v_trainside --cfg job --resolve` — passed; resolved `master_dtype=fp32`.
- Temporary CPU contract harness on commit `0b7b0f0` — passed. It covers generator forwarding through `step`/`step_with_logp`, per-sample/per-step trainside SDE seeds, seed-count validation, SGLang `PipelineConfig` routing, and the older-runtime fallback. The one-off harness is not committed per repository policy.
[...]

PR #220, opened 2026-07-18
fix(hunyuan-video): align text conditioning and unstick trainside GRPO

leviking98z-rgb:
## Summary

Consolidates the HunyuanVideo-1.0 work from #220 and the now-closed #215:

1. **Use canonical LLaMA conditioning.** The trainside encoder previously used `outputs.last_hidden_state`; official HunyuanVideo, Diffusers, and the SGLang path use the third-from-last state, `hidden_states[-3]`. This adds `hidden_state_skip_layer` (default `2`), threads it through the pipeline, rejects negative/out-of-range values, and keeps `0` available for the legacy baseline.
2. **Update the canonical trainside recipe in place.** The default recipe now uses the validated climbing-PickScore settings: LoRA rank 64 / alpha 256, batch 8, 8 samples per prompt, 10 denoising steps, standard landscape 1280x720x5 output, and 4 SDE steps. It also restores two PPO updates and the complete FSDP block list.
3. **Add the matching SGLang recipe.** `hunyuan_video_t2v_sglang.yaml` uses the same 1280x720x5 training/sampling defaults and a local TP=1 SGLang engine on each of 8 GPUs. It trains LoRA with FSDP and sends merged transformer weights through `TensorWeightSync`, avoiding the known online-SGLang-LoRA delta mismatch.
4. **Make HunyuanVideo a real video rollout adapter.** The adapter now returns packed video frames for `VideoPickScore` and preserves SGLang's two text streams as `text_llama` and `pooled_clip`, including variable-length LLaMA padding and attention masks required by trainer-side replay.
5. **Make 720p execution reliable.** CUDA text encoding disables the unstable cuDNN SDPA backend while retaining PyTorch's other SDPA backends. Both 720p recipes enable Diffusers VAE spatial tiling to bound fp32 decode activation memory.

## Related Issue

N/A. Supersedes the closed #215, whose recipe work is consolidated here.

## Test Plan

- Repository checks:
  `SKIP=no-commit-to-branch pre-commit run --all-files --show-diff-on-failure`
  — all hooks passed, including YAML, Python AST, ruff, private-key detection, and recipe `_target_` resolution.
- NVIDIA H20 CUDA/bf16 hidden-state smoke with a real four-layer `transformers.LlamaModel`: `skip=0` exactly matched `hidden_states[-1]`, and `skip=2` exactly matched `hidden_states[-3]` (zero tolerance).
[...]

PR #390, opened 2026-08-25
fix(hunyuan-video15): align vLLM-Omni rollout policy

leviking98z-rgb:
## Summary

Align HunyuanVideo-1.5 trainside and vLLM-Omni at the input, stochastic
transition, and online-LoRA boundaries:

- preserve the exact upstream MLLM system-prompt whitespace because
  `mllm_crop_start=108` is tied to that tokenized prefix;
- keep fp32 LoRA master parameters on both matched recipes while retaining
  BF16 forward parameters;
- derive SDE noise from the same `(seed, step, sample_id)` tuple on both
  engines, preserving per-sample independence across vLLM request splitting;
- form the transformer timestep as fp32 `sigma * 1000` before the BF16 cast,
  matching Diffusers/vLLM;
- pin the trainer's SDPA backend in the vLLM-Omni stage configuration;
- expose the main blocks' packed QKV LoRA mapping and derive token-refiner
  targets directly from the PEFT config, wrapping every matching `nn.Linear`;
  and
- keep `algorithm.old_logp_source: replay` until final-HEAD rollout/replay
  parity is validated within the recipe's `1e-4` clip range.

The cleanup deliberately excludes the pin-specific exact-arithmetic
monkey-patches for autocast, Q/K RMSNorm, SDPA mask construction, and RoPE, as
well as the rejected merged-weight and raw-VAE-reward experiments.

## Related Issue

Related to #386.

## Test Plan

- `python -m py_compile` on all changed Python modules — passed.
- `ruff check` and `ruff format --check` on all changed Python modules —
  passed.
- `pre-commit run --files <all changed files>` — all applicable hooks passed,
  including YAML, AST, Ruff, formatting, recipe target, docstring,
  experimental-boundary, and dependency-direction checks.
- Static AST comparison confirms the trainside prompt template is byte-equal
  to both pinned Diffusers and vLLM-Omni (`477` characters).
- YAML assertions confirm the recipes resolve `master_dtype: fp32` and
  `old_logp_source: replay`, the stage resolves
  `diffusion_attention_backend: TORCH_SDPA`, and the separate NCCL sync constructs
  with `lora_merged: true` plus the current `name_remap` interface.
- The prompt-only matched run changed first-rollout reward delta from the
  unpatched `+0.0370` to `-0.0007`, confirming that prompt whitespace is an
  initial-conditioning requirement rather than a cosmetic edit.
- The scheduler contract produced bit-identical emitted transitions and log
  probabilities on both engines while assigning different SDE noise to
  different samples.
- The timestep contract found the old trainside ordering produced `976`
  versus `980` and `684` versus `680` at two steps of the matched schedule;
  fp32 scaling makes all ten transformer timesteps bit-identical.
- Live LoRA coverage increased from `324` to `444` active wrappers. In the
  loader-only versus full-coverage comparison at rollout 75, correlation
  improved from `0.964188` to `0.992005` and MAE from `0.00398356` to
  `0.00226971`.
- The accepted full-coverage arm remained aligned past the suspected failure
  point: correlation/MAE were `0.999703 / 0.00046770` through rollout 10 and
  `0.991211 / 0.00354971` through rollout 125. At rollout 140 correlation was
  `0.990625` with mean bias `+0.00309`; no ratio or reward-curve collapse
  occurred:
  - [vLLM-Omni run](https://wandb.ai/leviking98z-zhejiang-university/tmp/runs/hv15p390lc10000826a)
  - [trainside reference](https://wandb.ai/leviking98z-zhejiang-university/tmp/runs/hv15p390tra10000826a)

That live curve included the exact-arithmetic probes now excluded from the
final diff. The selected input/weight/LoRA contracts were validated within
that run, but the narrowed final HEAD has not been rerun on GPU; the recipes
therefore retain the trainer-side replay anchor.

## Compatibility / Risk

- The runtime LoRA patches are gated to the pinned HunyuanVideo-1.5
  vLLM-Omni model. Existing checkpoint loading remains unchanged.
- The sample-keyed SDE path is opt-in: it is used only when the request carries
  `denoise_seed_keys`; existing callers without those keys keep their prior RNG
  behavior.
- Token-refiner adapters use two ordinary BF16 linear operations instead of a
  Punica wrapper, adding a small refiner-forward cost.
- The targeted refiner wrapper supports adapter slot 0, matching the recipe's
  single-online-adapter setup, and rejects an unexpected slot explicitly.
- No checkpoint schema, dataset format, or trajectory wire-format migration is
  introduced.

## Reviewer Notes

The exploratory history is intentionally retained. Commit `89e29069` removes
the broad candidate set additively; commit `c95c7768` restores only the
validated input/training contracts and small backend selection path. The final runtime contains exactly two HunyuanVideo-1.5 monkey-patches, both
for LoRA coverage.

Duplicate-work check: #220 covers HunyuanVideo-1.0 conditioning/merged rollout
weights, while #371 covers the SGLang pipeline/SDE path and supports both
sample-keyed noise and the rollout-logp anchor choice. Neither applies all
HunyuanVideo-1.5 vLLM-Omni packed/refiner LoRA targets. AI assistance was used;
the final diff and reported validation results were reviewed before
publication.

## Checklist

## EXP
截屏2026-08-29 04 40 36


CjhHa1 (approved):
LGTM

PR #397, opened 2026-08-26
feat(rollout): migrate the vllm-omni engine to 0.27.0rc1 on CUDA 13

celve:
## Summary

Moves the vLLM-Omni rollout engine from 0.20.0 to 0.27.0rc1 — seven upstream
releases — and with it the environment from CUDA 12.9 to CUDA 13. The five RL
diffusion pipelines (SD3, Qwen-Image, HunyuanVideo 1.5, HunyuanImage3, BAGEL)
are adapted to the contracts that changed along the way.

A 20-rollout SD3 FlowGRPO run on 8×H20 lifted mean PickScore from **0.7606 to
0.8547** (+0.094). The pre-rebase curve was 0.7605 → 0.8513; the two track each
other within noise.

## What changed upstream, and how this PR adapts

- **`forward` became batched.** It now takes a `DiffusionRequestBatch` and
  returns a list. These pipelines saturate the GPU inside a single request via
  `num_outputs_per_prompt` and never relied on request batching, so SD3 and
  Qwen-Image — the two that opt in upstream — now declare
  `supports_request_batch = False`. All five unwrap through one
  `single_request` helper that asserts the batch holds exactly one. BAGEL's
  packed GRPO path inspects that unwrapped request, not the batch. Stage configs
  set `max_num_seqs: 1`, replacing deprecated `runtime.max_batch_size`.

- **`DiffusionOutput.custom_output` was deleted** (vllm-omni #4922). Pipelines
  now attach RL-only data to a private `_unirl_captures` bag on
  `DiffusionOutput` while leaving generated media raw for upstream postprocess.
  A formatter hook copies the bag into `metadata["unirl"]` only after
  postprocess, and driver-side adapters read it from
  `OmniRequestOutput.multimodal_output`. Parent and inline engines install the
  hook through `vllm_omni.general_plugins`; `StageDiffusionProc` does not load
  that entry point, so the existing spawn-child hijack installs the same
  idempotent hook in multi-stage HI3 processes. Both `output_formatter` and
  `diffusion_engine` are rebound because the latter holds a from-imported
  formatter reference.

- **Output envelopes are unwrapped without losing data.** HI3/BAGEL metadata is
  merged into the private capture bag before image/video postprocess. SDE
  trajectory attributes take precedence over BAGEL's upstream payload
  trajectory; if the SDE scheduler produced no trajectory, the upstream payload
  trajectory is retained as the fallback instead.

- **Three signature-level breaks are handled.** SD3's `prepare_latents` dropped
  `dtype` and `device`; HI3's `build_prompt_tokens` now returns a
  `PromptTokensResult`; BAGEL's `_merge_naive_caches` became `NaiveCache.merge`.
  Latent injection binds the target's actual signature instead of relying on
  fixed positional slots.

- **Upstream deleted `model_executor/stage_configs/`** in #3172. The two HI3
  comprehension YAMLs sourced from there are vendored at their last upstream
  revision. Stage YAMLs are now always local, so the dead `stage_yaml_source`
  knob and its unreachable `"upstream"` branch are removed.

- **`MASTER_PORT` resolution changed.** vllm-omni ranks the environment above
  the constructor argument and strips `master_port` from single-engine args.
  Engine construction now temporarily sets `MASTER_PORT` to UniRL's reserved
  port and restores the caller's previous value afterward.

- **Expired compatibility patches are removed.** The deleted
  `ImageKVCacheManager` made `patch_hi3_flow_alignment` a silent no-op, and
  vllm-omni 0.27 already provides the Qwen3-Omni Thinker LoRA, audio-length and
  M-RoPE fixes previously carried by UniRL.

- **HI3 AR sets `fully_sharded_loras: true`.** vLLM 0.27's fused-MoE LoRA
  one-shot kernel takes an illegal memory access at this recipe's rank; the flag
  selects the legacy two-kernel path upstream still maintains.

## Environment: cu129 → cu130

vLLM ≥0.26 ships CUDA-13 binaries and requires `torch==2.13.0`, whose wheel is
cu130. A `+cu129` local-version pin would still satisfy that requirement while
linking CUDA-12 torch under CUDA-13 vLLM, so the vLLM extra pins the matching
`torch==2.13.0+cu130` and `torchvision==0.28.0+cu130` wheels and drops the cu129
index. `torchaudio==2.11.0+cu130` is deliberately not included: it is built
against torch 2.11 and no compatible ≥2.12 cu130 wheel exists on the index.

The SGLang extra remains an internally matched torch/vision/audio 2.11 cu130
stack. The two engine extras cannot share a virtual environment.

On the driver-535 fleet, CUDA 13 requires NVIDIA's forward-compat layer on
`LD_LIBRARY_PATH`; INSTALL.md documents `.cuda-compat-13/` and
`CUDA_COMPAT_DIR` setup.

## Test Plan

Static checks on the current head pass: pre-push hooks, `ruff check`, `ruff
format`, stage-config resolution, dependency-direction guards and recipe target
validation.

Dynamic runs recorded during the migration, all on 8×H20 with torch
2.13.0+cu130 / vllm 0.27.0 / vllm-omni 0.27.0rc1:

| Surface | Result |
|---|---|
| SD3.5 FlowGRPO, 20 rollouts | 0.7606 → 0.8547 mean PickScore, clean exit |
| HunyuanVideo 1.5 | 2 rollouts clean |
| HunyuanImage-3 | full training step; PPO ratio 0.9975±0.0655, abs Δlogp ~4e-2 |
| BAGEL t2i | 2 rollouts clean, σ guard green |
| BAGEL it2i | 2 rollouts clean, σ guard green (wiring only; not edit quality) |
| Qwen3-Omni (AR engine) | 2 rollouts clean |

Post-rebase SD3 curve:

| 1–5 | 6–10 | 11–15 | 16–20 |
|---|---|---|---|
| 0.7606 | 0.7991 | 0.8235 | 0.8501 |
| 0.7550 | 0.7998 | 0.8381 | 0.8404 |
| 0.7410 | 0.8196 | 0.8333 | 0.8459 |
| 0.7685 | 0.8355 | 0.8348 | 0.8400 |
| 0.7906 | 0.8177 | 0.8332 | 0.8547 |

## Not covered

- **Latest capture-channel cleanup on GPU.** The current head's private-bag /
  formatter-flush implementation, including explicit installation in HI3's
  multi-stage `StageDiffusionProc`, has static and focused unit coverage but
  still needs an end-to-end GPU regression run.
- **Qwen-Image.** Base t2i weights were not staged.
- **HunyuanImage-3 after the first training step.** A later illegal memory
  access surfaces in `_rotate_half` after CuMemAllocator sleep/wake. HI3 long
  runs are not claimed stable.
- **Strict numerical parity.** No seed-matched A/B against 0.20.0.
- **HI3 denoise alignment.** This remains a numerical-parity item after removal
  of the dead compatibility patch.
- **Multi-node.**
- **Release stability.** vllm-omni 0.27.0rc1 is a release candidate.


celve:
### Verified against the HunyuanImage-3 recipe

`examples/unified_model/hi3_vllmomni.yaml` is the heaviest consumer of this engine code — two-engine AR→DiT, an 80B MoE backbone, LoRA synced into both — so it exercises paths the SD3 run does not. Run on 8× H20 with torch 2.13.0+cu130 / vllm 0.27.0 / vllm-omni 0.27.0rc1, on the rebased branch.

**Working on 0.27:**

- Eight HI3 contract checks pass — `PromptTokensResult` unwrapping (B5), all six stage YAMLs resolving locally plus the two vendored comprehension configs (B3), conditions off the capture envelope (B1), and the postprocess shim registered and signature-transparent.
- Both engines boot (AR on GPUs 0-3, DiT on 4-7); FSDP meta-init materializes the 80B across 8 ranks with all 64 LoRA base-layer params finite.
- AR generates recaptions, DiT renders 1024×1024 over 16 steps, sleep/wake cycles correctly, and both weight-sync worker extensions inject.
- At the recipe's own `batch_size: 2` the rollout runs start to finish.

**Two blockers stop a full training step, neither from this PR:**

1. `ValueError: batch_size=2 not divisible by dp_size=8` — `RewardService` shards over `num_devices` and refuses a remainder, so rollout 0 aborts *after* generation. This PR touches nothing under `unirl/reward/`, `unirl/distributed/` or `unirl/trainer/`, and does not modify the recipe.

2. Raising the batch to clear (1) hits an illegal memory access in **core vLLM 0.27.0**'s `_fused_moe_lora_one_shot_kernel`, reached from HI3's MoE layer with LoRA attached. `_fused_moe_lora` dispatches on `M_pairs * max_lora_rank <= 1024`; at the recipe's rank 8 that threshold is 128 token-expert pairs, so `batch_size=2` takes a safe small-batch kernel and `batch_size=8` takes the faulting one. Ruled out memory pressure (reproduced with zero allocator warnings), UniRL's LoRA tensors (the small-batch kernel reads the same ones), and CUDA-graph replay (engine runs eager).

The two pull against each other — (1) pushes batch size up, (2) caps it — so lowering `num_devices` looks like the direction that satisfies both. Neither is a regression from the upgrade.

celve:
Follow-up to the HunyuanImage-3 verification above: **a complete HI3 training step now runs on this branch.**

```
rollout 1/2  reward=0.8160  ar[loss=0.0088 gn=0.0278 lr=1.00e-04
             ratio=0.9975±0.0655 clip=0.39 |Δlogp|=3.97e-02]
```

A first-step PPO ratio of 0.9975±0.0655 and `|Δlogp|` of 4e-02 mean rollout and training agree on the log-probabilities, which is the property the migration most needed to preserve.

**What changed in this branch.** One line: `fully_sharded_loras: true` on the HI3 AR stage. vLLM 0.27's fused-MoE LoRA one-shot kernel takes an illegal memory access; the flag gates above the kernel choice, so it falls back to the legacy two-kernel path upstream already maintains. Run at `batch_size=8` with `AR_TOKENS=1024` — byte-for-byte the configuration that previously faulted in rollout 0, differing only in the flag — rollouts 0 and 1 now complete with zero illegal memory accesses, `fused_moe` kernels JIT-compiling as expected.

`vllm==0.27.1` is not an alternative: it is four commits over 0.27.0, three of them CI, and touches neither `vllm/lora/` nor `csrc/moe/`.

**Two things reviewers should know, neither of which is this branch's.**

The reward-sharding error reported earlier was not the binding constraint. Reward scoring carries no gradient and had no business enforcing a gradient invariant, so I fixed that separately — but the `batch_size=2` run then failed at `stack.train_track` with the same error from the same guard, where it is correct. Eight ranks cannot take an equal share of two prompt-groups. **HI3 needs `batch_size` to be a multiple of `num_devices`**, and no dispatch change alters that. The reward fix is on a separate branch and deliberately not in this PR.

Rollout 2 then hit an illegal memory access at a *different* site — `hunyuan_image3.py:1427` in `_rotate_half`, rotary embedding, not the LoRA kernel — immediately after the CuMemAllocator sleep/wake cycle that follows each training step. CUDA reports asynchronously, so that stack is where it surfaced, not necessarily where it originated. Two candidates: sleep/wake not restoring state correctly, or UniRL's `HI3ARWeightSyncExtension` writing into the layout `fully_sharded_loras` changed. Distinguishing them is one run with weight sync disabled; I have not done it. Nothing before this work ran long enough to reach that code path.

celve:
All five vllm-omni pipelines have now been exercised on real GPUs (8×H20), and the last one that failed is fixed in `efb18479`.

**BAGEL was discarding UniRL's entire SDE trajectory.** It surfaced as the σ guard firing with `got (14,), sent (15,)`, which looked like an off-by-one in the schedule. Instrumenting the drain showed the worker was building the correct 15-point trajectory and the driver was receiving 14 — so nothing was miscounted; the right trajectory was being dropped in transit.

The cause is this upgrade's own surface. 0.27's BAGEL pipeline writes its trajectory into `output["payload"]["trajectory"]`, and `output_formatter` resolves from that payload first, consulting the `DiffusionOutput` attributes only as a fallback:

```python
for key, value in fallback_fields:
    if key not in trajectory and value is not None:
        trajectory[key] = value
```

`drain_trajectory_into` set only the attributes, which is what 0.20 read. This is the same `custom_output` → payload/metadata migration handled elsewhere in the branch, missed for the trajectory.

Worth flagging that the shape was the lucky part: the same payload-first path also resolves `log_probs`, so BAGEL would have trained on upstream's non-SDE capture. Had the two conventions agreed on length, this would have produced plausible-looking runs indefinitely.

The fix writes the payload as well as the attributes, and is inert for the other four pipelines — BAGEL is the only upstream pipeline that builds that payload. SD3 re-ran with the fix in place and is unaffected.

**Per-model GPU results**

| Model | Result |
|---|---|
| SD3.5 | 4 rollouts clean |
| HunyuanVideo 1.5 | 2 rollouts clean |
| HunyuanImage-3 | full training step, caveats tracked separately |
| BAGEL | 2 rollouts clean, σ guard green (values, not just shape) |
| Qwen-Image | untested — base t2i weights not staged |

**One earlier flag withdrawn.** The `gn=0.0000` reported for SD3 was `%.4f` rounding: at full precision it is 2–4e-5 across all four rollouts, and the near-zero loss is arithmetic, since `_grpo_clip_loss` at ratio ≈ 1 reduces to `-mean(advantage)` and GRPO mean-centres advantages within a group. No gradient problem. Unrelated to this PR, but `wandb_logger`'s `%.4f` on grad norm cannot represent a healthy value on these recipes.

celve:
Three verification gaps are now closed, all on the rebased tree at branch HEAD.

**The reward curve rises again after the re-port.** This was the gap that mattered: the original 20-rollout SD3 FlowGRPO run predated the rebase onto this repository, so it exercised code the re-port rewrote. Repeated through the `sd3-vllm-omni` verification profile at branch HEAD:

| 1–5 | 6–10 | 11–15 | 16–20 |
|---|---|---|---|
| 0.7606 | 0.7991 | 0.8235 | 0.8501 |
| 0.7550 | 0.7998 | 0.8381 | 0.8404 |
| 0.7410 | 0.8196 | 0.8333 | 0.8459 |
| 0.7685 | 0.8355 | 0.8348 | 0.8400 |
| 0.7906 | 0.8177 | 0.8332 | 0.8547 |

0.7606 → 0.8547, **+0.094 mean PickScore**, clean exit. The pre-rebase run went 0.7605 → 0.8513 for +0.09, and the two track each other closely: both start at 0.760, both dip at rollout 3, both cross 0.82 at rollout 8, both land above 0.85. The re-port did not perturb the learning dynamics measurably.

**Qwen3-Omni was a sixth surface the earlier verification missed.** It reaches vllm-omni through the AR engine rather than a diffusion pipeline, so it has no entry in the static checks, but it has its own adapter and compat patch that this upgrade touched. Two rollouts, clean:

```
rollout 1/2  reward=0.6875  gn=0.0947  ratio=0.9964  k3=6.86e-03
rollout 2/2  reward=0.5312  gn=0.0898  ratio=0.9917  k3=5.68e-03
```

Unambiguously live numbers, unlike the DiT recipes.

**BAGEL it2i runs clean too**, with the σ guard green. Worth calling out because it was the least-tested code and the most likely to be wrong: it is the rebase's one *semantic* conflict, where upstream added its own it2i conditioning where UniRL injects its own, resolved by hand. Two substitutions keep this a test of the conditioning and nothing else — EditReward swapped for PickScore, and generated source images — so it says the wiring survives 0.27 and says nothing about edit quality.

**Still not covered**, for the record: Qwen-Image (no weights staged), HunyuanImage-3's two residual faults, numerical parity against 0.20 as a per-tensor comparison, and multi-node.

CjhHa1 (approved):
LGTM

PR #401, opened 2026-08-27
fix(rollout): separate transfer and policy versions

### LINKED ISSUE #324: [Bug] Agentic rollouts stamp Parts with a weight-update call count, not the policy version
## Summary

`set_policy_version` is the only sanctioned way to tell a rollout engine which train policy it currently holds, and it has exactly two references in the repo: its definition in `unirl/rollout/engine/synchronous.py` and one call from the **batch** path in `unirl/trainer/async_policy.py`. The agentic path never calls it. Its Parts are still stamped, though — so in an agentic run `Part.weight_version` carries whatever the engine's `_weight_version` happens to hold, which is the count of weight-update API calls the worker has received.

That is precisely the value `set_policy_version` documents as invalid:

```python
def set_policy_version(self, train_version: int) -> None:
    """Assign the optimizer-update version of the currently loaded weights.

    Weight transports may receive one model through several buckets, so
    receiver-local API-call counters are not valid policy provenance. The
    trainer calls this only after a full weight publication succeeds.
    """
```

## Why the field ends up wrong

Three writers share one attribute, `SyncRolloutEngine._weight_version`, and one reader stamps it onto Parts (`_stamp_weight_version` → `Part.fill(weight_version=...)`):

1. `set_policy_version` sets it to `train_version` (committed optimizer updates) — valid provenance.
2. Every concrete engine bumps it by one inside its weight-update handlers — `sglang/engine.py`, `sglang_diffusion/engine.py`, `vllm_omni/engine.py`, `fastvideo/engine.py` all do `self._weight_version += 1` in `update_weights_from_tensor` / `update_weights_from_distributed`. This is the receiver-local counter the docstring rules out, and its magnitude depends on bucket count and transport, not on policy identity.
3. `AsyncAgenticRolloutEngine.sync_weights` bumps a *different* `_weight_version` — a driver-side publication tally on the wrapper object — and never forwards anything to the workers.

So:

- **Batch path is correct, but only by ordering.** `AsyncBatchControl.sync_rollout` runs `weight_sync.sync()` and then `rollout.set_policy_version(self.train_version)` on the next line, so the bumps from (2) are overwritten before anything is stamped. Nothing enforces that order.
- **Agentic path is wrong.** `sync_weights` does (3) only, so the workers' `_weight_version` keeps whatever (2) left it at. Parts are then stamped with a bucket-call count.

## What it breaks

`unirl/trainer/agentic_async.py` logs two different clocks two lines apart, both named after the same concept:

```python
"async/publication_count": self._engine.weight_version,   # driver tally, advances per sync
"async/version_span": (max(versions) - min(versions)),    # from gp.weight_version, a bucket count
```

`version_span` is meant to show how far a carried multi-turn trajectory spans policies — that is the whole justification for per-Part stamping (see the `asynchronous.py` module docstring: "the per-turn version spread inside a carried trajectory is corrected per-token by each gen Part's own `weight_version`"). As written it reports a difference of API-call counts.

The per-token **correction** itself is unaffected: it uses the rollout's own logprobs, not the version field. What is broken is the provenance bookkeeping and the metric built on it.

## Suggested fix

1. Give the receiver-local tally its own attribute (e.g. `_weight_update_calls`) so it cannot land on the provenance field. This also removes the batch path's silent dependency on `set_policy_version` running last.
2. Have the agentic sync path publish a train version: `AsyncAgenticRolloutEngine.sync_weights` should accept the trainer's train version and call `set_policy_version` after `weight_sync.sync()`, the same shape the batch path already uses.
3. Consider making `_stamp_weight_version` fail loudly (or stamp `None`) when no version has ever been published, so a future path that forgets step 2 does not silently produce a plausible-looking integer.

## Naming makes this hard to see

Three unrelated clocks are all called `weight_version`: the engine's policy provenance, `AsyncAgenticRolloutEngine`'s publication tally, and `WeightSync.weight_version` (a sender-side counter used only for deterministic checkpoint filenames, documented as such in `unirl/distributed/weight_sync/README.md`). A rename giving each clock one name — e.g. `behavior_version` / `publication_count` / `transfer_seq` — would have made this a type error rather than a plausible number. Worth doing alongside the fix.

## Verified / not verified

- Verified by reading code: the two `set_policy_version` references, the four engine bump sites, `sync_weights` not forwarding a version, and the stamping path.
- Not verified on a run: the actual values `Part.weight_version` takes in an agentic job, and therefore whether `async/version_span` is nonzero-but-meaningless or coincidentally zero. Worth confirming with one short ALFWorld or deep-research async run before picking between fix options 2 and 3.

quinnarnold:
## Summary

Keep receiver-local transfer counts separate from the trainer policy version stamped on rollout output. Successful weight transfers increment `_weight_update_calls`, leaving `_version` to the publication contract.

Rebased onto `c8af5d6` and included the checkpoint-engine IPC receiver added in #225. Its successful update previously changed published version 7 to 8; it now preserves 7 and counts one transfer.

## Related Issue

Fixes #324.

## Test Plan

- `python3 ../review-followup-2026-09-20/check_unirl_receivers.py .` — disposable CPU harness discovers and executes all 15 concrete/forwarding `update_weights_from_*` methods from the checkout with stubbed transport backends. All 30 success/failure cases pass, plus three-bucket counting, non-TP-zero no-ops, and checkpoint-engine poisoned-retry checks. The same harness failed on the rebased code before the one-line IPC fix (`_version=8`, expected 7).
- `SKIP=no-commit-to-branch uvx pre-commit run --files unirl/rollout/engine/fastvideo/engine.py unirl/rollout/engine/sglang/engine.py unirl/rollout/engine/sglang_diffusion/engine.py unirl/rollout/engine/vllm_omni/engine.py --show-diff-on-failure` — all applicable hooks pass.
- No new GPU/model run locally. The reviewer's separate H20/SGLang tensor-path result is not included as locally performed validation.

## Compatibility / Risk

No configuration or public API changes. Existing publication ordering, checkpoint-engine failure poisoning, and native vLLM publication remain unchanged. CPU checks exercise the receiver method bodies, not real transport implementations.

## Reviewer Notes

AI-assisted.

## Checklist

- [x] Full diff checked for unrelated/generated artifacts.
- [x] Relevant validation completed and limitations documented above.

### DIFF
--- a/unirl/rollout/engine/fastvideo/engine.py
+++ b/unirl/rollout/engine/fastvideo/engine.py
@@ -181,6 +181,7 @@ class FastVideoRolloutEngine(BaseRolloutEngine):
         )
 
         self._version = 0
+        self._weight_update_calls = 0
         self._generate_lock = threading.Lock()
         self._shutdown_lock = threading.Lock()
         self._shutdown_requested = False
@@ -620,7 +621,7 @@ class FastVideoRolloutEngine(BaseRolloutEngine):
             require(self._generator is not None, "fastvideo engine is offloaded/not initialized")
             self._generator.update_transformer_weights_from_path(checkpoint_path)
             self._last_weights_path = checkpoint_path
-            self._version += 1
+            self._weight_update_calls += 1
             logger.info("fastvideo transformer weights updated from %s", checkpoint_path)
 
 
--- a/unirl/rollout/engine/sglang/engine.py
+++ b/unirl/rollout/engine/sglang/engine.py
@@ -152,6 +152,7 @@ class SGLangRolloutEngine(BaseRolloutEngine):
         )
 
         self._version = 0
+        self._weight_update_calls = 0
 
     def _prepare_generation(self, sample: Sample) -> Any:
         require(
@@ -282,7 +283,7 @@ class SGLangRolloutEngine(BaseRolloutEngine):
             load_format=load_format,
             flush_cache=flush_cache,
         )
-        self._version += 1
+        self._weight_update_calls += 1
 
     def init_weights_update_group(
         self,
@@ -329,7 +330,7 @@ class SGLangRolloutEngine(BaseRolloutEngine):
             group_name=group_name,
             flush_cache=flush_cache,
         )
-        self._version += 1
+        self._weight_update_calls += 1
 
     def destroy_weights_update_group(
         self,
--- a/unirl/rollout/engine/sglang_diffusion/engine.py
+++ b/unirl/rollout/engine/sglang_diffusion/engine.py
@@ -98,6 +98,7 @@ class SGLangDiffusionRolloutEngine(BaseRolloutEngine):
         self.schedule_policy = self.adapter.schedule_policy()
 
         self._version = 0
+        self._weight_update_calls = 0
         self._generate_lock = threading.Lock()
         self._shutdown_lock = threading.Lock()
         self._shutdown_requested = False
@@ -236,7 +237,7 @@ class SGLangDiffusionRolloutEngine(BaseRolloutEngine):
             load_format=load_format,
             flush_cache=flush_cache,
         )
-        self._version += 1
+        self._weight_update_calls += 1
 
     def init_weights_update_group(
         self,
@@ -279,7 +280,7 @@ class SGLangDiffusionRolloutEngine(BaseRolloutEngine):
             target_modules=target_modules,
             flush_cache=flush_cache,
         )
-        self._version += 1
+        self._weight_update_calls += 1
 
     def destroy_weights_update_group(
         self,
--- a/unirl/rollout/engine/vllm_omni/engine.py
+++ b/unirl/rollout/engine/vllm_omni/engine.py
@@ -38,6 +38,7 @@ class VLLMOmniRolloutEngine(BaseRolloutEngine):
     ) -> None:
         self.cfg = config
         self._version = 0
+        self._weight_update_calls = 0
         self._generate_lock = threading.Lock()
         self._shutdown_lock = threading.Lock()
         self._shutdown_requested = False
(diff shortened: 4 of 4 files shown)


Zcchill (changes requested):
I rebased the PR commit onto current `main` (`a77575a`) and reran the version semantics dynamically, including on an H20 with a real in-process SGLang engine.

**Blocking: cover the checkpoint-engine IPC receiver added after this branch forked.** Current `main` commit `8fc283e` added `SGLangRolloutEngine.update_weights_from_checkpoint_engine_ipc()`, whose success path still does `self._version += 1`. The PR branch predates that method, so the patch applies cleanly but leaves this new transport as another writer of policy provenance.

Concrete repro on the rebased tree: initialize the engine state with `_version=7`, perform a successful checkpoint-engine IPC update, and the result is `_version=8`, `_weight_update_calls=0`. The same runtime harness passed the tensor/NCCL/IPC/path success, failure, three-bucket, manager-publication, and generation-stamping cases before failing specifically on this newer method.

Please rebase and change that success path to increment `_weight_update_calls` instead, ideally with a regression test that enumerates every concrete `update_weights_from_*`/`update_weights_from_path` receiver and asserts that transfers never mutate `_version`.

For positive GPU validation, I booted a tiny Llama model through the real SGLang native backend on an NVIDIA H20, generated once, published version 7, performed a real serialized tensor weight update, and generated again. The observed chain was `initial=0 -> published=7 -> post-transfer=7`, with `_weight_update_calls=1`, so the intended tensor-path behavior is correct.

quinnarnold:
Thanks for the detailed reproduction and H20 check. Rebased onto `c8af5d6` and fixed the checkpoint-engine IPC success path in `585c5ef` to increment `_weight_update_calls`.

A disposable CPU harness now discovers all 15 concrete/forwarding `update_weights_from_*` methods and executes their actual method bodies with stubbed transports. It reproduced version 7 becoming 8 before the fix; afterward, all 30 success/failure cases pass, along with three-bucket counting, non-TP-zero no-ops, and poisoned-retry checks. The harness stays outside the repo per its test-harness policy. All applicable pre-commit hooks pass; I haven't rerun a GPU model locally.

PR #404, opened 2026-08-28
fix(reward): inject nested T2AV scorers and require prompt-video alignment

celve:
## Summary

`T2AVCompositeScorer` rebuilt every inner spec from defaults and forwarded a hardcoded field allow-list. Inner-only fields such as `ImageBindSpec.mode` were unreachable, leaving ImageBind at `audio_video`. Combined with CLAP (prompt↔audio only), that gives a T2AV objective with no prompt↔video term and permits the measured black/silent collapse.

The composite now receives Hydra-instantiated backends under `scorers`, matching `PerDomainRewardScorer`; registry lookup, inner-spec reconstruction, and field forwarding are gone. Prompt-video support is an optional `PromptVideoReward` protocol instead of a method on every reward backend.

The composite rejects mismatched scorer/weight names, non-backend or non-video scorers, nonnumeric/nonfinite outer weights, inner scorer failures, and mixes without a positive-weight prompt-video term. ImageBind now validates `mode` and validates its three finite `mode: all` weights locally.

All three in-tree T2AV recipes use nested scorers. MiniMax-H3 NFT selects `mode: all` and writes ImageBind's current internal weights explicitly so the objective is reviewable.

## Related Issue

Follow-up: #423 tracks capability validation before eager inner-model loading. Interacts with #403: a weights-only ImageBind + CLAP recipe must migrate to nested scorers and select `text_video` or `all`.

## Test Plan

- Full pre-push hook suite — pass (YAML, AST, Ruff, recipe targets, dependency boundaries, one-line docstrings)
- `PYTHONPATH=. python3 /tmp/verify_t2av_nested.py` — pass:
  - nested blending and inherited `preferred_input_kind`
  - absent, zero, negative, and nonfinite prompt-video coverage rejected
  - mismatched keys, non-backend entries, and incompatible media kind rejected
  - inner `successes=False` propagates as composite failure
  - strict ImageBind mode validation and complete all-mode weights
  - positive internal ImageBind/VideoAlign alignment coefficients
- `PYTHONPATH=. python3 /tmp/verify_t2av_hydra.py` — all three T2AV recipes instantiate with model loading stubbed; H3 NFT reaches ImageBind `mode: all`
- Hydra `--cfg job --resolve` composition for all three recipes — pass (`PRETRAINED_MODEL=/tmp/dummy` for H3)
- End-to-end GPU rollout: Not run; this changes reward construction and the H3 NFT objective. A fresh run is required before reusing old `audio_video` measurements as training evidence.

## Compatibility / Risk

Breaking T2AV configuration: weights-only composites no longer construct scorers implicitly; every weight needs a matching nested backend. In-tree recipes are migrated. The constructor still accepts `base_device` for the common backend/factory contract, but inner scorers own device configuration. The now-inert composite `batch_size` field was removed.

H3 NFT ImageBind changes from `audio_video` to `all`, so its old reward measurements do not describe the new recipe. Internal weights remain ImageBind's existing 0.5 / 0.25 / 0.25 defaults; this PR does not invent a new unmeasured mix.

## Reviewer Notes

- Validation is limited to cheap structural/numeric invariants. Scorer-specific mode/weight validation stays in ImageBind; the composite does not encode scorer business rules or impose subjective reward thresholds.
- The guard runs after Hydra constructs inner backends. Because local backends load eagerly, it prevents training but not model-loading cost; #423 tracks a shared two-phase lifecycle.
- Rebased onto current `main`; unrelated diff removed.
- AI-assisted follow-up; collapse measurements in the original report are from celve.

## Checklist

### DIFF
--- a/examples/diffusion/ltx2/ltx2_3_t2av_audioreward_trainside.yaml
+++ b/examples/diffusion/ltx2/ltx2_3_t2av_audioreward_trainside.yaml
@@ -12,8 +12,9 @@
 #   - clap           (audio-text alignment, LAION CLAP; zero extra deps)
 # ImageBind (audio-video alignment) is also available as an inner scorer but is
 # NOT enabled here: it requires the facebookresearch/ImageBind package and is
-# CC-BY-NC-SA 4.0 (NonCommercial). To enable, add `imagebind: <weight>` to the
-# composite weights (and install the package).
+# CC-BY-NC-SA 4.0 (NonCommercial). To enable it, nest an ImageBind scorer with
+# mode: text_video or all; audio_video does not relate the prompt to the video,
+# so the composite rejects that mode unless another scorer supplies this term.
 #
 # With audio_joint_sde=true, audio is part of the SDE policy AND now receives a
 # direct reward signal, so it is optimized toward prompt alignment (not just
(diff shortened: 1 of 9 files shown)


CjhHa1:
Update: the earlier per-name override approach has been replaced. The composite now takes fully Hydra-instantiated nested backends, matching `PerDomainRewardScorer`; registry lookup, inner-spec reconstruction, field forwarding, and override helpers are gone.

The prompt-video guard remains and now requires positive outer and inner alignment coefficients. All in-tree T2AV recipes are migrated; H3 NFT uses ImageBind `mode: all`.

CjhHa1 (approved):
LGTM

PR #405, opened 2026-08-28
fix(meta-init): resolve Hugging Face checkpoint paths

### LINKED ISSUE #400: feat(meta-init): resolve Hugging Face repo IDs before sharded loading
## Problem

Meta-init bundles stash `_transformer_weights_path` for the shared sharded loader. Model config loading accepts either a local checkpoint path or a Hugging Face repo ID, but `load_sharded` only accepts a local directory containing safetensors.

A recipe can therefore resolve successfully with a value such as `org/model`, load expensive auxiliary components, and fail only later during backend construction:

```text
sharded_load: transformer weights dir not found
HF repo IDs are not supported here
```

This affects any meta-init bundle using `_transformer_weights_path`; it is not specific to one model. Large models are especially affected because disabling meta-init introduces a per-rank full-model load spike.

## Desired behavior

Provide a shared checkpoint-path resolver for meta-init bundles that:

- passes local snapshot paths through unchanged;
- resolves Hugging Face repo IDs to a local cached snapshot/component directory;
- avoids every distributed rank independently downloading the same checkpoint;
- supports component subfolders such as `transformer/`;
- fails with an actionable error before expensive auxiliary models are loaded;
- preserves offline/local-cache operation.

## Acceptance criteria

- Existing local-path recipes behave unchanged.
- A meta-init bundle configured with a Hugging Face repo ID reaches `load_sharded` with a real local directory.
- Concurrent ranks reuse one cached download rather than starting duplicate downloads.
- Missing/offline checkpoints fail during bundle setup with the original repo ID and expected component path in the error.
- Coverage includes mocked local paths, Hub resolution, component subfolders, and concurrent/cache reuse behavior.

## Context

This was exposed while reviewing MiniMax-H3 support in #347, whose temporary model-specific mitigation is to require `PRETRAINED_MODEL` to point to a local snapshot.

quinnarnold:
## Summary

Add a shared meta-init checkpoint resolver that keeps local paths unchanged, resolves Hugging Face repo IDs through the standard snapshot cache, validates root or component safetensors before auxiliary models load, and gives actionable errors for missing or offline checkpoints. Wire it into every bundle that supplies `_transformer_weights_path`, replacing MiniMax-H3’s temporary local-only restriction and Cosmos3’s model-specific download path.

## Related Issue

Fixes #400

## Test Plan

- `SKIP=no-commit-to-branch uvx --from pre-commit pre-commit run --all-files --show-diff-on-failure`
- Uncommitted Python resolver harness covering unchanged local root/component paths, missing component validation, mocked Hub resolution, eight concurrent callers converging on one cached download, and offline cache-miss errors.
- `pytest --collect-only -q --ignore=unirl-reward-service` (no first-party pytest tests are present by repository policy)

## Compatibility / Risk

Eager-loading paths are unchanged. Meta-init local checkpoints now validate their safetensors directory during bundle setup, while Hub IDs download only the required safetensors pattern through `snapshot_download`; standard Hub cache and offline behavior is preserved.

## Reviewer Notes

AI assistance was used for implementation and review. The full diff and each changed call site were manually inspected, and a duplicate-PR search found no overlapping open work. No test files were added because the repository requires one-off verification harnesses to remain uncommitted.

## Checklist


CjhHa1 (approved):
LGTM

CjhHa1:
Hi there, I've reviewed your code and fix some logics, thx again for contributing!

PR #406, opened 2026-08-30
feat(train): add shuffle knob for diffusion multi-update

### LINKED ISSUE #323: [Bug] Diffusion multi-update splits mini-batches contiguously; flow_grpo shuffles first
## Summary

Every diffusion recipe with `stack.num_updates_per_batch > 1` partitions the rollout into per-optimizer-step mini-batches by **contiguous index range**. The flow_grpo reference implementation shuffles the whole rollout *before* it splits. Because our samples arrive group-by-group, contiguous splitting hands each optimizer step a set of **complete** prompt groups, while flow_grpo hands each step a random subset that cuts groups apart.

Neither is obviously wrong, but the published flow_grpo curves were produced with the shuffle and we deviate from it silently.

Scope: **72** recipes under `examples/diffusion/`, `examples/unified_model/`, `examples/pe/` run at `num_updates_per_batch >= 2`. The 16 `examples/ar/*` recipes at `>= 2` are **not** affected (see "AR is already at parity").

## What flow_grpo does

`scripts/train_sd3.py` (yifan123/flow_grpo), inside the outer epoch:

```python
for inner_epoch in range(config.train.num_inner_epochs):
    # shuffle samples along batch dimension
    perm = torch.randperm(total_batch_size, device=accelerator.device)
    samples = {k: v[perm] for k, v in samples.items()}

    # rebatch for training
    samples_batched = {
        k: v.reshape(-1, total_batch_size//config.sample.num_batches_per_epoch, *v.shape[1:])
        for k, v in samples.items()
    }
```

Optimizer steps per rollout is `samples_per_epoch // total_train_batch_size` times `num_inner_epochs`. With the `config/base.py` defaults (`sample.num_batches_per_epoch = 2`, `train.batch_size = 1`, `gradient_accumulation_steps = 1`, `num_inner_epochs = 1`) that is 2 — the same 2 our diffusion recipes carry, and it comes from `num_batches_per_epoch`, not from re-epoching. The `randperm` runs before the reshape, so it is what decides *which* samples land in step 1 vs step 2.

## What we do

`unirl/train/stack/planner/types.py`:

```python
def _update_ranges(*, total_size: int, num_updates: int) -> Tuple[Range, ...]:
    ...
    update_size = total // n
    return tuple((i * update_size, (i + 1) * update_size) for i in range(n))
```

Contiguous, no permutation. Both planners route through it: `count.py` directly, and `packed.py` slices `samples[u_start:u_end]` first and only reorders *within* each update for token packing. So update membership is contiguous in original order either way.

## Why the membership matters

Samples reach the stack group-contiguous (`Sample.concat` over per-prompt groups). With 4 prompts x 4 samples and `num_updates_per_batch=2`:

```
index : 0  1  2  3 | 4  5  6  7 | 8  9 10 11 | 12 13 14 15
prompt: A  A  A  A | B  B  B  B | C  C  C  C |  D  D  D  D

ours      : update 1 = A,B complete        update 2 = C,D complete
flow_grpo : update 1 = random 8 of 16      update 2 = the other 8
            (partial groups, all 4 prompts in both)
```

GRPO advantages are group-normalized *before* the split, so a complete group's advantages sum to ~0. Ours therefore gives every optimizer step a balanced within-group contrast, at the cost of each step seeing a disjoint prompt subset. flow_grpo gives every step all prompts but partial groups, so one step can be net-positive or net-negative (e.g. a group's two best samples both land in step 1). Across the N steps that averages out; within a step it does not.

Prompt order entering a rollout is already random on both sides (`run.shuffle` defaults to `True` in `MultimodalRLDataSource`), so at *prompt* granularity the two schemes are statistically equivalent. The remaining difference is group integrity.

## AR is already at parity

verl exposes this as an explicit knob and defaults it off:

```python
# verl/workers/config/actor.py
ppo_epochs: int = 1
shuffle: bool = False
```

Our contiguous split matches verl's default, so the AR recipes are at reference parity as-is. (`data.shuffle: True` in verl is the dataloader-level prompt shuffle — a different layer, which we also do.) This issue is diffusion-only.

## What to decide

1. **Match flow_grpo** — permute the track before `_update_ranges`, seeded so runs stay reproducible. Restores parity for all 72 recipes; changes numerics on every existing diffusion run.
2. **Make it a knob** — a `TrainStack(shuffle_updates=...)` mirroring verl's `actor.shuffle`, default `False` to preserve today's behavior, set `True` only in the recipes that claim flow_grpo parity.
3. **Keep contiguous and document it** — argue that whole-group mini-batches are better conditioned for group-normalized advantages, and record the deviation in `unirl/train/readme.md`.

Option 2 looks best to me: it is the only one that both preserves current numerics and makes a parity run requestable. But it needs an SD3 A/B first to establish whether the difference is measurable at all — if it is not, option 3 with a documented note is cheaper.

## Verified (was: not verified)

Resolved in the comment below. Summary: **DanceGRPO and MixGRPO do not shuffle the batch axis** — they shuffle the *time* axis (per-sample `randperm(num_timesteps)`), so our contiguous split is already at parity with them and their recipes must be excluded from any batch-axis fix. **FlowDPPO** has no external reference implementation and inherits whatever is decided here. This narrows the batch-axis gap to the flow_grpo-parity recipes only, and makes option 2 the clear choice over option 1.

The comment also records a **separate time-axis gap**: our `AllSDEScheduler` draws the trained SDE steps once per rollout and shares them across the whole batch, while DanceGRPO/MixGRPO draw per sample. That needs its own issue.

f2842:
## Summary

- add planner-layer update shuffling, default off
- preserve AR/image lineage alignment for both equal-size and fan-out parts
- seed each permutation from `shuffle_seed + rollout_id` for checkpoint-resume reproducibility
- release pre-shuffle training tensors before device alignment to avoid retaining a second full track during training
- enable shuffling in 50 explicit flow_grpo-parity multi-update recipes; keep DanceGRPO, MixGRPO, FlowDPPO, AR, CPS, and custom UniGRPO recipes unchanged

## Related Issue

Fixes #323

## Test Plan

- full pre-commit suite passed
- `py_compile` passed for all changed Python modules
- verified equal-size and P*N to P*N*M lineage-preserving permutations; malformed equal-size lineage is rejected
- verified the same rollout id reproduces the same permutation independent of prior calls
- verified reordered input storage is released before training; a 256 MiB CPU payload retained only 4.3 MiB RSS overhead after release
- verified exactly 50 target recipes enable `shuffle_updates`, with no DanceGRPO/MixGRPO/FlowDPPO recipe enabled


Ideny42 (commented):
The pre-shuffle tensors are not actually released on the normal remote path.

Worker.call retains both the raw args and hydrated fetched tensors until train_track returns, while the controller also retains localized leases until ray.get completes. Therefore, clearing input_part.segment/conditions/advantages only removes one reference; the transport-owned original tensors remain alive while Part.select() materializes the full shuffled copy.

I verified this with a 1-GPU Ray + colocate_store A/B using an 8 GiB LatentSegment:

shuffle_updates=false: 10.00 GiB peak (nvidia-smi: 10,623 MiB)
shuffle_updates=true: 18.00 GiB peak (nvidia-smi: 18,815 MiB)
after _release_reordered_track_inputs: allocated memory remained 16.00 GiB (8 GiB original + 8 GiB shuffled copy)
checksums matched, so this is a memory-lifetime issue rather than data corruption
The CPU RSS check in the PR body does not exercise this remote GPU ownership path. This extra full-track allocation can OOM large image/video recipes. Could we avoid materializing the complete shuffled Part, for example by carrying the permutation in the update plan and gathering each update/micro-batch lazily?

PR #407, opened 2026-08-31
feat(sensenova): add OCR FlowGRPO recipes (full fine-tune, LoRA, KL-anchored) and fix the OCR reward deps

leviking98z-rgb:
## Summary

Adds trainside FlowGRPO recipes for `sensenova/SenseNova-U1.5-8B-MoT-SFT` on the OCR text-rendering reward, plus the reward-side fixes they need. Three recipes ship because the OCR objective has a trade-off that no single configuration resolves:

| recipe | eval `edit_similarity` | `exact_match` | image content preserved |
|---|---|---|---|
| `..._ocr_trainside_fullft_4x8.yaml` | **0.895** (peak 0.931) | 0.562 | no |
| `..._ocr_trainside_lora_4x8.yaml` | 0.854 (peak 0.862) | 0.531 | no |
| `..._ocr_trainside_lora_kl_4x8.yaml` | 0.527 | 0.188 | **yes** |

All measured over 100 rollouts from the same SFT base on the full `datasets/ocr` splits (19653 train / 1018 test, median target 15 non-space characters).

**Reward side.** `unirl/reward/local/ocr.py` now consistently uses PaddleOCR with a normalized Levenshtein primary score, Unicode-aware normalization, and edit / substring / exact / counter-IoU diagnostic components. `unirl/trainer/diffusion.py` surfaces those component means during evaluation, which is what made the analysis below possible.

**Two optimizer settings carry most of the result.** `adam_epsilon` moves from the usual `1e-8` to `1e-15`: this model's per-parameter gradient RMS is ~5e-9, below a 1e-8 epsilon, so Adam's denominator was eps-dominated and the realised update was ~38% of `lr`. Holding `lr` fixed and changing only epsilon takes full fine-tuning from 0.296 → 0.334 to 0.296 → 0.895, and LoRA from 0.645 to 0.854. `max_grad_norm` drops to 0.003 (LoRA) / 0.01 (full-FT) because gradient norms here are ~1e-4 to 5e-4, so the original 1.0 could never fire and mid-training spikes went unclipped.

**Known limitation, documented in-file.** The reward scores OCR text and nothing else. At `beta=0` there is no anchor to the base policy, so over 100 rollouts the model learns to blank the scene and render high-contrast text on it: dark-pixel fraction 0.451 → 0.926, edge density −60%, colour entropy −54%. The OCR gains are not fabricated (counter-IoU rises independently of the substring rule; the non-exempt subset mean goes 0.2235 → 0.6879) — the reward simply never measured generic image quality. `beta=0.01` prevents the collapse at the cost of 38% of the OCR gain, which is the third recipe.

**Dependencies.** The `eval` extra moves from `easyocr>=1.7` to `paddlepaddle` / `paddleocr` / `python-Levenshtein`. The `easyocr` pin was declared in four places and imported nowhere, while `ocr.py` on `main` already imported paddleocr and Levenshtein, neither of which was declared — so `pip install -e ".[eval]"` installed an unused package and still could not run the OCR reward. PaddlePaddle (~180 MB) is deliberately kept out of `requirements.txt` and lives only in the extra; `INSTALL.md` states the size.

No new dataset files: the recipes train and evaluate on the existing `datasets/ocr/train.txt` and `test.txt`.

## Related Issue

N/A

## Test Plan

- `python3.12 -m compileall -q unirl`
- `SKIP=no-commit-to-branch pre-commit run --files <changed files> --show-diff-on-failure`
- Hydra compose + `instantiate` on all three recipes (compose alone silently accepted a stray LoRA key that only failed inside a Ray worker)
- `yaml.safe_load` on all three recipes after the in-file caveat blocks were added
- Dependency check on the H20 nodes: PaddlePaddle `3.2.2`, PaddleOCR `3.5.0`, python-Levenshtein `0.27.4`
- Five 100-rollout training runs on 4×8 H20 covering the epsilon, learning-rate and KL dimensions; curves, generated samples and image-quality statistics are in the PR comments

## Scope

One 100-rollout run per configuration, no seed sweep, so slopes are directional. The epsilon finding is the robust part: single-variable comparisons against otherwise identical runs, with the mechanism verified against `torch.optim.AdamW` directly rather than inferred from the curve. `beta=0.001` is untested — the two shipped `beta` points bracket the trade-off but do not locate its knee.


leviking98z-rgb:
## Re-validated on the full splits (4x8 H20)

Pushed two follow-up commits: this PR no longer ships any dataset file. `short_text_train.txt` was a 1140-line subset of the existing `train.txt` (verified line-for-line) and `short_text_test.txt` was filtered the same way, so both were dropped and the recipe now points at `datasets/ocr/train.txt` / `datasets/ocr/test.txt`.

That matters for the numbers: the subsets kept only one-or-two ASCII words of 3-8 characters, which made the originally published curve easier than the task it claimed to measure. The full splits have a median target of **15 non-space characters across ~3 words**.

Re-ran the same recipe (LoRA r64/alpha128, lr 3e-4, `samples_per_prompt=6`, PaddleOCR + normalized Levenshtein, 320x320 / 6 denoising steps, 32 GPUs) for 100 rollouts on the full data:

| metric | start | end |
|---|---|---|
| eval `edit_similarity` (training reward) | 0.2963 | **0.6453** (+118%) |
| eval `counter_iou` (independent of the substring rule) | 0.3516 | **0.6427** |
| eval `exact_match` | 0.0625 | 0.1562 |
| training `reward_mean` | 0.302 | 0.574 |

reward curves

`counter_iou` rising by a comparable margin is the load-bearing check here. `edit_similarity` grants a free 1.0 whenever the target appears as a substring of the OCR reading, so it can in principle be gamed by rendering extra text. `counter_iou` is a character-bag IoU with no such exemption, and it tracks `edit_similarity` throughout. Decomposing the reward the other way agrees: the mean score of the samples that did *not* hit the substring exemption goes 0.2235 -> 0.5271.

Samples from the run, first rollout vs last:

samples

Early outputs misspell the target ("Neever Forget"); late ones render multi-word targets legibly. The visible failure mode shifts from wrong glyphs to imperfect letterforms on long strings, which is consistent with `exact_match` staying low (0.156) while `edit_similarity` reaches 0.645 -- getting all 15 characters exactly right is much harder than getting most of them right.

Note on scope: this is a single 100-rollout run per configuration, not a seed sweep, so treat the slope as directional rather than a precise effect size.

leviking98z-rgb:
## Full fine-tune recipe added — and one config value turned out to gate the whole path

Pushed `examples/diffusion/sensenova_u1_5/sensenova_u1_5_ocr_trainside_fullft_4x8.yaml`. On the full splits it reaches **eval edit similarity 0.895 (peak 0.931)** versus the LoRA recipe's 0.645, over the same 100 rollouts from the same SFT base:

| metric | full-FT | LoRA r64 |
|---|---|---|
| eval `edit_similarity` | 0.2963 → **0.8954** (peak 0.9313) | 0.2963 → 0.6453 |
| eval `counter_iou` (independent of the substring rule) | 0.3516 → **0.8728** | 0.3516 → 0.6427 |
| eval `exact_match` | 0.0625 → **0.5625** | 0.0625 → 0.1562 |
| training `reward_mean` | 0.281 → **0.909** | 0.302 → 0.574 |
| slope /rollout | **+0.00811** | +0.00361 |

curves

The grey line is the same recipe with `adam_epsilon` left at the usual `1e-8` — it sits flat at ~0.33 for all 100 rollouts. That is the entire difference between the two full-FT curves.

### Why `adam_epsilon` matters here

This model's **per-parameter gradient RMS is ~5e-9**, which is *below* a `1e-8` epsilon. Adam's denominator `sqrt(v) + eps` becomes eps-dominated, the adaptive scaling never engages, and the realised update collapses to ~38% of `lr`. Measured directly with `torch.optim.AdamW` at that gradient scale: 200 steps move a parameter by `2.27e-3` against an ideal `6.0e-3`; at `eps=1e-15` it moves the full `6.0e-3`.

Holding `lr` at `3e-5` and changing only epsilon takes the curve from `0.296 → 0.334` to `0.296 → 0.895`. Worth noting this is not a full-FT-specific problem: LoRA sits at the same per-parameter gradient scale (6.4e-9) and is throttled identically — it simply runs 10x more `lr`, so 38% of it was still enough to learn. The LoRA path is left unchanged in this PR pending its own validation run.

`max_grad_norm` also drops `1.0 → 0.01`. Gradient norms here are ~5e-4, so a 1.0 threshold can never fire: an earlier `lr=5e-5` attempt reached 0.605 by rollout 70, took an unclipped spike (grad_norm 0.00044 → 0.00286), and settled back at 0.365 without recovering. At 0.01 normal steps are untouched while spikes get truncated — this run clipped twice near rollout 15, dipped to 0.15, and recovered within ~8 rollouts before going on to 0.93.

### Samples

samples

Rollout 1 renders illegible marks; by rollout 100 multi-word targets like "Height Requirement 48in" come out clean across three lines. `exact_match` at 0.562 means 18 of the 32 eval prompts are read back character-for-character correct.

### Reward hacking check

`edit_similarity` grants a free 1.0 whenever the target appears as a substring of the OCR reading, so it is gameable by rendering extra text. Two independent checks say that is not what happened: `counter_iou` (character-bag IoU, no substring exemption) rises by a comparable margin, and decomposing the reward gives a mean score of **0.2235 → 0.7609** over just the samples that did *not* hit the substring exemption. The `substring_match − exact_match` gap stays at +0.016 → +0.034.

### Scope

One 100-rollout run per configuration, no seed sweep — treat slopes as directional. The eps finding is the robust part: it is a single-variable comparison against an otherwise identical run, and the mechanism was verified against the optimizer directly rather than inferred from the curve.

leviking98z-rgb:
## The epsilon fix applies to LoRA too — and the dependency footprint is now opt-in

Two more commits. First, `adam_epsilon` and `max_grad_norm` on the LoRA recipe: LoRA sits at essentially the same per-parameter gradient scale as full-FT (6.4e-9 vs 4.9e-9), so `eps=1e-8` throttled it identically. It just had 10x more `lr` to absorb the loss, which is why it looked fine while full-FT looked broken.

Holding everything else fixed (`lr 3e-4` constant, r64/alpha128, spp6, full splits), changing only epsilon:

| metric | LoRA, eps 1e-8 | LoRA, eps 1e-15 | full-FT, eps 1e-15 |
|---|---|---|---|
| eval `edit_similarity` | 0.2963 → 0.6453 | 0.2963 → **0.8537** (peak 0.8620) | 0.2963 → **0.8954** (peak 0.9313) |
| eval `counter_iou` | → 0.6427 | → **0.8385** | → 0.8728 |
| eval `exact_match` | → 0.1562 | → **0.5312** | → 0.5625 |
| slope /rollout | +0.00361 | **+0.00515** | +0.00811 |

curves

samples

`max_grad_norm` goes to 0.003 rather than the full-FT recipe's 0.01: LoRA's gradient norms are an order of magnitude smaller (p50 1e-4, max 9e-4 against full-FT's 5e-4 / 1.1e-2), so both the original 1.0 and 0.01 would have been unreachable thresholds. At 0.003 it clipped once in 100 rollouts — spikes truncated, normal steps untouched.

### Where the two recipes stand

full-FT still ends higher (0.895 vs 0.854, peak 0.931 vs 0.862) at the cost of a 127 GB checkpoint against LoRA's 2.4 GB, and it takes a much deeper early dip (down to 0.15 around rollout 25, versus LoRA climbing monotonically past rollout 20). Both are in the repo; the trade is checkpoint size and early-phase stability against a few points of final reward.

### One caveat worth stating plainly

`exact_match` on the LoRA run is not monotone: it reaches 0.500 at rollout 60, falls to 0.094 by rollout 85, then recovers to 0.531 at 100. `edit_similarity` and `counter_iou` rise more smoothly through that window. So part of the late-phase gain is the model getting *closer* on characters rather than getting *more strings exactly right*, and a single-seed run cannot separate that from eval noise on 32 prompts. If exact-string correctness is what matters for a downstream use, evaluate the checkpoint on that metric directly rather than reading it off the reward curve.

Reward-hacking checks still pass: `counter_iou` (no substring exemption) rises comparably, and restricted to samples that did *not* hit the substring exemption the mean score goes 0.2235 → 0.6879.

### Dependency footprint

`paddlepaddle` and `paddleocr` are out of `requirements.txt` and live only in the `eval` extra. `paddlepaddle` is a ~180 MB second deep-learning framework needed solely for the PP-OCRv5 detection/recognition models behind `unirl.reward.local.ocr`, so a default install no longer pays for it; `INSTALL.md` states the size. `python-Levenshtein` stays in the core set at ~45 KB.

For context on the extra changing at all: the `easyocr>=1.7` it replaced was declared in four places and imported nowhere, while `ocr.py` on main already imported `paddleocr` and `Levenshtein`, neither of which was declared. `pip install -e ".[eval]"` on main therefore installed an unused package and still could not run the OCR reward. Keeping PaddleOCR rather than switching to easyocr was deliberate: swapping the OCR engine swaps the reward's judge, which would invalidate every curve in this PR and re-bias what the model is rewarded for.

leviking98z-rgb:
## Correction: the reward degrades image quality outside what it measures

I need to correct an earlier comment in this PR. I wrote that the reward-hacking checks passed — those checks were sound, but their **scope was wrong**. I only verified that the *text* score wasn't inflated by the substring exemption. I never checked anything else in the image.

Someone looking at the samples pointed out that late-training outputs are "just text, the background is gone." That is correct, and it is measurable. On the fixed eval prompts, over 100 rollouts of the LoRA recipe with `beta=0`:

| image metric | rollout 0 | rollout 100 |
|---|---|---|
| dark-pixel fraction (<0.15) | 0.451 | **0.926** |
| edge density (structure) | 0.01183 | 0.00479 (**−60%**) |
| colour entropy (content) | 4.037 | 1.849 (**−54%**) |
| brightness | 0.2710 | 0.0719 (−73%) |

Monotone across all 21 eval points, not a single-sample artifact. The mechanism is straightforward: `edit_similarity(OCR(image), target)` scores text and nothing else, `beta=0` means there is no anchor to the base policy, and blanking the scene raises text contrast. The model is optimising the objective exactly as written.

To be clear about what still holds: the OCR gains are not fabricated. `counter_iou` (no substring exemption) rises independently, and restricted to samples that never hit the exemption the mean goes 0.2235 → 0.6879. This is reward misspecification, not reward hacking — every monitored metric looks good, which is what makes it easy to miss.

### A KL anchor prevents it, at a real cost

`beta=0.01` (KL against the LoRA-disabled base policy), everything else identical:

| | no KL (`beta=0`) | KL `beta=0.01` |
|---|---|---|
| eval `edit_similarity` | **0.8537** | 0.5268 (peak 0.5520) |
| eval `exact_match` | **0.531** | 0.188 |
| eval `counter_iou` | **0.839** | 0.601 |
| dark-pixel fraction | 0.451 → 0.926 (**+105%**) | 0.451 → 0.395 (**−12%**) |
| colour entropy | 4.037 → 1.849 (**−54%**) | 4.037 → 3.930 (**−3%**) |
| edge density | −60% | −24% |

tradeoff

samples

The anchor works — scenes stay intact, entropy barely moves — but it costs 38% of the OCR gain. `beta=0.01` still beats the base policy by +78% on edit similarity (0.296 → 0.527), so it is a usable operating point, just a different one.

Both recipes now carry this caveat in-file rather than only in a comment thread. `beta` stays at 0 by default because the recipes' stated purpose is the OCR curve, and the caveat points at three options depending on what you need: raise `beta`, take an earlier checkpoint, or add an image-quality reward term. Note `beta>0` requires LoRA — the reference policy is the adapter-disabled model, so the full-FT recipe raises if you set it.

`beta=0.001` is untested; the two points here bracket the trade-off but do not locate its knee.

KemingWu (approved):
LGTm

PR #408, opened 2026-08-31
feat(minimax-h3): rework the vLLM-Omni H3 rollout onto 0.27.0rc1

celve:
## What this is

A rework of #378 (MiniMax-H3 vLLM-Omni rollout) onto current `main`, which has
moved under it: #397 migrated the engine to `vllm-omni==0.27.0rc1` and
introduced the capture channel, stage-YAML topology and `patches/` package that
#378 predates. This branch keeps that PR's H3 work and drops everything #397
now supplies.

Targets `vllm-omni==0.27.0rc1` deliberately — the 0.28.0rc1 migration is
separate, and the one patch here that overlaps it is written to self-retire
(see below).

For scale: #378 is 2500+/156- across 36 files; this is 1834+/51- across 30.
The difference is infrastructure that no longer has to ship with the model.

## Fixes that are not H3-specific

These stand on their own and are the first four commits, so they can be split
out if preferred:

- **`lora.optimize()` is pinned to one thread while loading an adapter.** Every
  sequence-parallel worker runs this CPU-bound call at once; unpinned, the ranks
  oversubscribe the box and the load stalls.
- **`CUDA_VISIBLE_DEVICES` is cleared before the first vllm import**, not after.
  vLLM's platform layer snapshots the visible-device set on first import, so the
  old ordering left the snapshot already taken.
- **The colocated LoRA handler gets the `extract()`/`push()` split** that
  `RemoteLoraWeightSync` already had, so a colocated run can offload the trainer
  between extracting adapter tensors and pushing them.
- **H3's float32-pinned transformer regions are kept out of autocast.** Under a
  global autocast the rope, projections and time embedder were silently cast to
  bfloat16.

## H3-specific

- **Fused qkv and SwiGLU are declared to the diffusion LoRA manager.**
  `DiffusionLoRAManager` binds adapters by walking `stacked_params_mapping`;
  `v0.27.0rc1` declares no rows for `MiniMaxH3DiTModel`, so a trained adapter
  bound against **zero** layers — no error, rollout just served the base model.
  The patch extends the tuple and skips rows already present, so on 0.28 (where
  vllm-omni#6476 added the three qkv rows) the qkv half becomes a no-op and only
  the `mlp.fc1` half still does work. The 0.28 migration should not need to
  touch it.
- **The adapter's `target_modules` regex is re-anchored** to
  `^transformer\.blocks`. The manager matches component-prefixed names, so
  #378's `^blocks` anchor matched nothing.
- **A file-backed LoRA transport**, opt-in per adapter via `lora_file_transport`
  and `transport: file` in the recipe. H3's adapter is ~166M parameters, and
  base64 in the control message would put a ~440MB string on the RPC. Other
  models keep the existing handle and copy transports untouched.
- **Pipeline and adapter use the #397 capture channel** (`stamp_capture` /
  `read_captures`) instead of overloading `DiffusionOutput.trajectory_*`, and
  the pipeline is a package with profiling split out.
- **Topology lives in a stage YAML** (`minimax_h3_t2va_rl.yaml`) rather than
  engine-config fields, matching how every other model on `main` does it. The
  recipes carry only the per-variant `omni_extra` overrides their throughput
  comparison needs.

## Verification status

Lint is clean: all four `lint/` checks pass (2552 recipe `_target_` paths
resolve), `ruff check` and `ruff format` clean on the touched trees.

**No GPU run happened on this branch.** The throughput and parity numbers in
`examples/diffusion/minimax_h3/README.md` are #378's measurements, carried over
as the record of what that deployment achieved; they have not been reproduced
here. The topology, LoRA-binding and transport changes above all want a 32-GPU
run before this merges.

## Known gap

`quality100` was qualified against uniform sampling over 9 frames, but
`T2AVCompositeSpec` on `main` has no multi-frame aggregation and its scorer
rejects any `frame_selection` outside `first`/`middle`. The recipe scores the
middle frame alone and the README records the gap — the multi-frame reward work
is a separate change.

Reward convergence on H3 remains open, as #378 already documented; this branch
changes the infrastructure it runs on, not that result.


celve:
## Verified on GPU

Pod `unirl-4x8-gz-1-1787583608` (H20 96 GB, GZ), `vllm-omni==0.27.0rc1` / `vllm==0.27.0` / `torch==2.13.0+cu130`. A full Flow-GRPO iteration completes on `minimax_h3_t2va_vllmomni_2x4_timeshare` — one 8-GPU node, two four-rank Ulysses replicas, 8 prompts x 4 siblings, 10 steps at 768x768x124:

```
rollout 1/1  reward=0.8467  loss=-0.0000 gn=0.0000 lr=3.00e-04
             ratio=1.0000±0.0000 clip=0.00 |Δlogp|=2.98e-07
exit 0
```

`|Δlogp|=2.98e-07` is the number worth looking at: the rollout-vs-replay log-prob gap, i.e. the vLLM-Omni engine's trajectory against the trainer's own recomputation of it. The two agree to fp32 noise, four orders below the 1e-3 gate the quality100 recipe sets.

Getting here needed four fixes beyond the original port, each a commit on this branch:

1. **Recipe layering** — every `*_vllmomni_*` recipe descended from `minimax_h3_t2va_trainside`, so Hydra merged `stage_attrs` / `forward_batch_size` into the vLLM-Omni engine, which takes neither. Split out `minimax_h3_t2va_base.yaml`; trainside is now a thin leaf. PR 378 absorbed these as ignored kwargs instead, which would hide a future typo.
2. **Multi-GPU replicas** — an engine could not span more than one GPU (`Stage 0 requires 4 device(s) ... 1 available`). `main` already has the mechanism, but it is opt-in via `_accepts_rollout_tp_kwargs` and only SGLang set it. The engine now boots on the replica head with its group's device list; the other three ranks answer the same RPCs as no-ops. This is also why `replica_size` / `tp_size` belong in `rollout.config` — an earlier pass deleted them as unknown fields.
3. **Checkpoint completeness** — the staged snapshot was missing `processor`, `tokenizer`, `scheduler`, `audio_scheduler`.
4. **Two checkpoint layouts** — vLLM-Omni reads the `FL2VA/` partition, the trainer reads the flat diffusers tree. Both point at the same `model_path`, so that directory has to carry both (~269 GiB node-local).

`gn=0.0000` is the recipe, not a defect: `sde_indices: [6]` injects exploration noise at one of ten steps and `init_same_noise: true` shares x_T, so a group's four siblings are near-identical and the centred advantages are tiny. Per-sibling seeding was checked directly and is correct. `[6]` is a memory decision — one replayed transition already peaks at 92,188 MiB of a 97,871-MiB H20.

Full write-up, including the two environment findings (HF-through-proxy is fast, `uv`-through-proxy is not; `uv`'s global cache lock turns one stall into a permanent one): https://github.com/foshstl/LIN-980/blob/main/docs/h3-vllm-omni-027-engine-test.md

celve:
## One more finding: training stops at rollout 2

The single-rollout test above passes, but a 30-rollout run dies waking the engine for rollout 2:

```
12:27:06 [Worker 3] Memory usage before sleep: 39.14 GiB.
12:32:22 [Worker 3] Responding to Wake-up Task: 18454af8-...
12:32:23 ERROR Wake-up failed on Rank 3: CUDA Error: out of memory at csrc/cumem_allocator.cpp:163
12:32:23 ERROR Wake-up failed on Rank 1: ...  Wake-up failed on Rank 2: ...
12:32:24 INFO  [Worker 0] Wake-up complete.
12:32:28 ERROR CUDA error: an illegal memory access was encountered
```

Ranks 1-3 could not restore the 39.14 GiB the allocator backed up to CPU; rank 0 could. The replica then entered its wake barrier with one rank holding weights and three holding nothing, and the next collective read freed memory — the illegal access is the consequence, not the cause.

What differs between the wake that worked and the one that did not is the optimizer step between them. Adam allocates `exp_avg` / `exp_avg_sq` lazily on its first `step()`, so from rollout 2 the trainer holds two extra tensors per LoRA parameter that were absent when the engine first sat down. `_generate_with_residency` wakes into that:

```python
self.rollout.wake_up()                 # engine restores 39.14 GiB/rank here
if sync_weights and self.weight_sync is not None:
    self.weight_sync.sync()
if should_offload_train:
    self.backend.offload()             # train state leaves the GPU only now
```

`backend.offload()` already moves optimizer state to CPU — it just runs after the allocation that needs the room. The ordering is deliberate, since weight sync must read the trainer's parameters while the engine is awake. But Adam's moments are not part of that contract: nothing in sync reads them, so moving only the optimizer state to CPU before `wake_up()` and back after would cut the overlap without changing what sync sees.

This is not specific to this branch — nothing in the rework touches sleep/wake, and any colocated run that trains more than one step should hit it. Flagging rather than fixing here, since `_generate_with_residency` is shared by every diffusion recipe and the fix wants its own change. Happy to take it if you'd like it on this PR.

Detail: https://github.com/foshstl/LIN-980/blob/main/docs/h3-vllm-omni-027-engine-test.md

celve:
## Fixed and verified: four consecutive rollouts on one 8-GPU node

`6fa7472d` fixes the rollout-2 wake OOM. Verified on `unirl-4x8-gz-1-1787583608`, single node, 8 GPUs, `minimax_h3_t2va_vllmomni_2x4_timeshare`:

```
13:07:07  pre-wake cache release: returned  0.00-2.14  GB across 8 rank(s)   <- first wake
13:22:44  rollout 1/4  reward=0.8467  |Δlogp|=6.56e-07
13:22:45  pre-wake cache release: returned 46.21-47.76 GB across 8 rank(s)   <- the wake that used to die
13:37:16  rollout 2/4  reward=0.8334  |Δlogp|=7.45e-07
13:37:17  pre-wake cache release: returned 48.12-48.20 GB across 8 rank(s)
13:51:48  rollout 3/4  reward=0.8420  |Δlogp|=1.01e-06
14:06:19  rollout 4/4  reward=0.8424  |Δlogp|=6.85e-07
```

Steady 14.5 minutes per rollout, no wake failure, no traceback.

### What the fix is

`backend.offload()` is the only step in the rollout loop that calls `empty_cache()`, and it runs *after* `wake_up()`. So the engine asked the driver for 39.14 GiB of physical pages while each train rank was still caching the blocks its last backward had reserved — the recipe puts that peak at 92,188 MiB of a 97,871-MiB card. Freed is not returned.

The overlap between wake and the trainer's parameters is required, since weight sync has to read them. Releasing *cached* blocks is not: it moves no tensor, so it can happen before the wake without changing anything sync sees. New `BaseBackend.release_cache()` does that per rank, and `_generate_with_residency` calls it first. It loops `gc.collect()` with `empty_cache()` rather than calling the latter once, because the backward leaves autograd graphs in reference cycles.

### One correction to my earlier comment

I blamed Adam's moments. The instrumentation says otherwise — about 48 GB per rank comes back at every post-training wake, and a rank-32 LoRA's optimizer state is on the order of a gigabyte. The retained allocator pool was the cause; the optimizer state was a rounding error next to it. The fix targets the pool.

Also worth noting for anyone reading logs from this recipe: the `EXIT=` marker the launcher prints is not trustworthy — the earlier crashed run printed `EXIT=0` too. Judge by the metrics line and the absence of a traceback.

Detail: https://github.com/foshstl/LIN-980/blob/main/docs/h3-vllm-omni-027-engine-test.md

Jayce-Ping (approved):
Looks good to me. The e2e gpu tests show the good training-inference consistency. There should be no problem.

PR #410, opened 2026-09-01
feat(train): add configurable HSDP and H3 2x8 recipe

CjhHa1:
## Summary

- Make the native FSDP2 hybrid shard degree configurable through `training.fsdp.hsdp_shard_size`, initialize the default process group before mesh construction, and fail fast on invalid or unapplied HSDP topology.
- Add a MiniMax-H3 trainside 2-node × 8-GPU HSDP recipe and reject statically known DP-geometry mismatches before constructing the 33B trainable model and frozen 32B conditioner.
- Reduce H3 conditioning cost by caching frozen prompt embeddings on CPU and serializing transient conditioner GPU residency per node; also allow local PickScore and CLAP checkpoints for network-isolated jobs.

## Related Issue

N/A

## Test Plan

- [x] `pre-commit run --all-files`
- [x] Hydra-composed `diffusion/minimax_h3/minimax_h3_t2va_trainside_hsdp_2x8` with `num_devices=16`, `batch_size=16`, `backend.fsdp_cfg.hsdp_shard_size=8`, and local reward checkpoint overrides.
- [x] Ran focused Python checks for HSDP mesh shape, DP geometry, prompt caching, and conditioner residency failure paths.
- [x] Completed a Taiji end-to-end topology-gate run on 2 nodes × 8 H20 GPUs (16 GPUs total), using MiniMax-H3 from the local `PRETRAINED_MODEL` snapshot, `datasets/pickscore/train.txt`, and local PickScore/CLAP checkpoints. Taiji instance: `8b1d8917a047494701a05fb3705e1fd3`.
- [x] The topology gate completed rollout, offline reward, backward, reduce-scatter, replica all-reduce, and optimizer step. Observed reward `0.3746`, gradient norm `0.0001`, and ratio `1.0000`.
- No unit-test files are added: the changed HSDP behavior requires an initialized multi-GPU process group and is covered by the 2×8 H20 topology gate above.

## Compatibility / Risk

- `hsdp_shard_size` defaults to `8`, preserving existing hybrid-mode behavior. `full` and single-process `no_shard` behavior remain unchanged.
- Hybrid mode now fails early when the process group is unavailable or uninitialized, the shard size does not divide `world_size`, or only one shard group would be created. Configurations that previously fell back silently must select a valid hybrid topology or use `fsdp_mode: full`.
- `text_encoder_onload_for_embed` defaults to `false`; transient VAE/conditioner swapping is enabled only by the new H20 recipe. The per-node lock uses Linux `fcntl`, matching the project's supported runtime.
- No checkpoint or dataset format changes are introduced. The new reward checkpoint fields retain their existing Hugging Face IDs as defaults.
- The two-step, one-transition H3 recipe is a distributed correctness and memory gate, not a production-quality training schedule or throughput benchmark.

## Reviewer Notes

- Please focus review on HSDP mesh construction/post-wrap verification in `unirl/train/backend/fsdp/wrap.py` and the conditioner residency/cache lifecycle in `unirl/models/minimax_h3/text_embed.py`.
- The new 2×8 recipe intentionally minimizes retained replay state while still exercising backward and both HSDP communication dimensions.
- AI assistance was used for review and validation cleanup; the author reviewed the resulting changes and validation evidence.

## Checklist

### DIFF
--- a/.claude/skills/development/add-model-bundle/SKILL.md
+++ b/.claude/skills/development/add-model-bundle/SKILL.md
@@ -60,6 +60,7 @@ Single-transformer bundles (the common case) branch in `from_config`:
 
 ```python
 if config.meta_init_transformer:
+    weights_path = resolve_meta_init_weights(path, component="transformer")
     transformer_config = <Class>.load_config(path, subfolder="transformer")   # diffusers
     transformer, meta_init_state = build_meta_init_transformer(               # unirl.models.types.meta_init
         lambda: <Class>.from_config(transformer_config),
@@ -70,14 +71,15 @@ else:
 ...
 bundle = cls(...)
 if config.meta_init_transformer:
-    bundle._transformer_weights_path = os.path.join(path, "transformer")      # diffusion layout
+    bundle._transformer_weights_path = weights_path
     bundle._meta_init_state = meta_init_state
 return bundle
 ```
 
+- `resolve_meta_init_weights` passes local snapshots through, resolves Hub IDs into the shared cache, and validates complete root/component safetensors shards.
 - `build_meta_init_transformer` builds under `accelerate.init_empty_weights(include_buffers=False)`, captures init-computed non-persistent buffers/plain tensor attributes, dtype-casts (metadata-only on meta), and stamps `init_weights` to a no-op. Stash its returned state on `bundle._meta_init_state`; the backend restores it after the sharded weight load, including across Ray actor serialization.
-- Stash `_transformer_weights_path` = the safetensors dir the backend reads via `load_sharded` (`unirl/train/backend/sharded_load.py`): `<ckpt>/transformer` for diffusers-layout models; the **checkpoint root** (`path`) for AR/VL models loaded through `AutoModelForCausalLM` (no subfolder).
-- AR/VL bundles use the same helper with an `AutoModelForCausalLM.from_config(...)` (qwen3) or `ModelClass(cfg)` (qwen_vl) factory. Structural setup that does not touch weights (`gradient_checkpointing_enable`, `requires_grad_(False)` for a frozen vision tower) runs on both builds and persists through `to_empty` + load.
+- Stash `_transformer_weights_path` from `resolve_meta_init_weights`; it is the local safetensors directory consumed by `load_sharded` (`unirl/train/backend/sharded_load.py`).
+- AR/VL checkpoint-root layouts call `resolve_meta_init_weights(path)` without `component`, then use an `AutoModelForCausalLM.from_config(...)` (qwen3) or `ModelClass(cfg)` (qwen_vl) factory. Structural setup that does not touch weights (`gradient_checkpointing_enable`, `requires_grad_(False)` for a frozen vision tower) runs on both builds and persists through `to_empty` + load.
 
 Per-architecture init-computed state that `to_empty` destroys is handled by the shared capture/restore path. `build_meta_init_transformer` deliberately keeps buffers and `__dict__` tensors real on CPU while parameters are meta, and raises if a captured tensor is unexpectedly still meta. This covers plain-tensor rope tables (Qwen-Image `QwenEmbedRope.pos_freqs`) and non-persistent sincos buffers (SD3 `PatchEmbed.pos_embed`) without bespoke rebuild/deferred-stamp helpers.
 - params the checkpoint omits (FLUX.2-klein guidance embedder): zero-init them post-load via a deferred op keyed on checkpoint-absent names — `to_empty` leaves them as garbage (not meta), so an `is_meta`-gated fix won't catch them.
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -1,0 +1,107 @@
+# Adapted from verl-project/verl-omni (originally Hugging Face transformers).
+name: Bug report
+description: Crash, wrong result, hang, or install failure
+labels: ["bug"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thanks for the report. Search existing issues first — duplicates are closed.
+
+        For install and recipe questions that are not a bug, use a blank issue or the WeChat group, and link back here if it later turns into a real defect.
+
+        Paste the output of this snippet from the venv you actually used (`vllm` or `sglang` extra, not system Python):
+
+        ```bash
+        git rev-parse HEAD
+        python -c "import sys, platform; print(sys.version); print(platform.platform())"
+        python -c "import torch; print('torch', torch.__version__, 'cuda', torch.version.cuda); print(torch.cuda.get_device_name(0) if torch.cuda.is_available() else 'no cuda')"
+        python -c "import importlib.metadata as m
+        for pkg in ('unirl', 'vllm', 'vllm-omni', 'sglang', 'torch'):
+            try:
+                print(pkg, m.version(pkg))
+            except m.PackageNotFoundError:
+                print(pkg, 'not installed')"
+        ```
+
+  - type: textarea
+    id: system-info
+    attributes:
+      label: System info
+      description: Paste the command output here. Do not leave the snippet un-run.
+      placeholder: UniRL SHA, Python, torch/CUDA, GPU name, package versions
+    validations:
+      required: true
+
+  - type: dropdown
+    id: engine
+    attributes:
+      label: Rollout / inference engine
+      options:
+        - vLLM-Omni
+        - SGLang
+        - FastVideo
+        - Train-side only
+        - Install / docs / CI (no engine)
+        - Other / mixed
+    validations:
+      required: true
+
+  - type: dropdown
+    id: domain
+    attributes:
+      label: Domain
+      options:
+        - diffusion (train_diffusion)
+        - diffusion async (train_async_diffusion)
+        - ar (train_ar)
+        - ar async (train_async_ar)
+        - sft (train_sft)
+        - pe (train_pe)
+        - unified_model (train_unified_model)
+        - agentic (train_agentic, examples/deep_research)
+        - reward service
+        - docs / install / CI
+        - other
+    validations:
+      required: true
+
+  - type: input
+    id: recipe
+    attributes:
+      label: Recipe / config-name
+      description: Hydra `--config-name`, or "N/A".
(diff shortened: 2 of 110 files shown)


leviking98z-rgb:
## Review — measured on 8×H20, with fixes pushed to the branch

**Verdict: the HSDP work is correct, and it fixes a real bug.** I reproduced the silent degradation on `main` and confirmed the PR's new gating actually takes effect. I also found four issues and pushed fixes to the branch (`a8d7dc8`, `52cee0b`, `57a5c4e`, `0f311e8`).

### Measured evidence (8×H20, torch 2.7.1, running this branch's real code — not a reimplementation)

**1. `fsdp_mode: hybrid` on `main` really does silently degrade to flat FSDP — this PR fixes a real problem**

Given `main`'s ordering, `_create_device_mesh` runs *before* the first `fully_shard()` call, and the latter is what lazily brings up the default PG. So the `return None` at `wrap.py:259` is guaranteed to hit:

```
[t=0] dist.is_initialized() = False
[main's _create_device_mesh] mesh=None because: PG NOT INITIALIZED -> return None
[after first fully_shard] dist.is_initialized() = True
[result] param mesh dim_names=None shape=(1,) ndim=1
[verdict] hybrid was SILENTLY DEGRADED TO FLAT
```

One correction while I'm here: the commit message for `fix(examples): keep bagel 2x8 on its validated HSDP topology` says that recipe "was 8-way sharded and 2-way replicated, exactly as its comment said". Measurement says it was actually 16-way flat sharded all along. The conclusion (keep hybrid, state the degree outright) is right; the stated reason needs fixing.

**2. This PR's HSDP genuinely takes effect under `root_wrap=false` + mixed dtype, with both collectives live**

Running this branch's `fsdp_wrap` on 8 GPUs (shard=4):

```
[e2e] resolve_fsdp_mesh_shape(hybrid, world=8, shard=4) = (2, 4)
[e2e] wrapped OK: mesh dim_names=('dp_replicate','dp_shard') shape=(2, 4) ndim=2
[e2e] grad placements=(Replicate(), Shard(dim=0)) dtype=torch.float32
[e2e] max |grad - rank0 grad| across all 8 ranks = 0.000e+00   -> PASS
```

I checked `_validate_hsdp_mesh`'s comparison separately: it returns `(2,4)` under `root_wrap=false` too, so no false positives. The `root_wrap=false` stray-param guard does bite in practice — it caught the trainable LayerNorm in my first test model.

**3. Things I was worried about that measurement ruled out** (to save other reviewers the time)

- The bare `init_process_group()` in `ensure_dist_initialized()` does **not** collide with the `dist.TCPStore` that `DevicePool._setup_nccl` already builds: I measured building the TCPStore first and then `init_process_group` on the same addr/port, and it works; the socket that `Worker._reserve_port` still holds doesn't block rendezvous either, thanks to `SO_REUSEADDR`. And this isn't new behaviour anyway — on `main`, torch already makes the same bare env:// call from inside `fully_shard`; this PR only hoists it and makes it explicit.
- `monitored_barrier(wait_all_ranks=True)` on a gloo subgroup under an nccl default PG is **valid**; one rank arriving 8s late still converges; a genuinely asymmetric call count produces a clean error rather than a silent hang.
- Per-node serialized residency does **not** fight the 30-minute barrier budget: measured H2D at 7.0 GiB/s (pageable) → 64GiB × 8 ranks serialized is about 1.2 minutes, so there's plenty of headroom.
- `embed()` call counts are identical across ranks on both the train and eval paths (`pytree_chunk` forces divisibility for DP_SCATTER; every eval chunk goes through `_validate_prompt_tree_dp_geometry`), so the barrier is symmetric.

---

### Issues fixed

**(1) `a8d7dc8` — the preflight's reward dp_size disagrees with the runtime check it stands in for (substantive: the gate is dead)**

`_preflight_trainside_geometry` exists to statically reject what `_validate_dp_geometry` would reject later. But the runtime reads `self.reward.dp_size`, and a reward Handle carries no sp/tp/pp, so its `dp_size` is its slab width — for a colocated reward, all shared devices. After `a4ac7fe` pinned it to 1, `batch_size % 1` is always 0, so **the reward arm of the gate became a no-op**.

Replaying all 55 trainside recipes through real Hydra composition shows the gap is live:

```
GAP  diffusion/wan21/wan21_t2v_full: nd=16 bs=24 reward_dp runtime=16 pr=1
     runtime_says=['rollout_dp=16 !| batch=24', 'reward_dp=16 !| batch=24']
     pr_says=['rollout_dp=16 !| batch=24']            <- reward arm missed
```

Changed it to take the slab width (1 only when there's no `reward:` block at all). Same replay: **55 recipes, 0 disagreements with the runtime**, and no new false rejections; this PR's own new recipe (16 devices / batch 16 / colocated reward) still passes.

**(2) `52cee0b` — `import ray` was hoisted out of the try, swallowing the actionable error the PR itself promises**

`_ray_job_id()` puts `import ray` outside the `try`, so an environment without ray raises `ModuleNotFoundError` instead of the `RuntimeError` that names the `UNIRL_RUN_ID` escape hatch (on `main` it was inside the try). The H3 residency lock now shares this path, so a missing ray blows up while naming a lock file, with no hint. Moved it back inside; verified that without ray it raises `RuntimeError` naming `UNIRL_RUN_ID`, and with the env var set ray is never imported.

**(3) `57a5c4e` — the T2AV per-scorer id forwarding is superseded by upstream #404, and conflicts**

This branch is 7 commits behind `main`, and #404 (`fix(reward): inject nested T2AV scorers`) rewrote `T2AVCompositeSpec` to hold `scorers: Dict[str, RewardBackend]`, each with its own `_target_`. A recipe can now set an offline checkpoint id **directly on the inner spec**, so the three forwarding fields plus the `hasattr` guard loop have nothing left to do. `git merge-tree` shows them conflicting with `main` in both the module and the two recipes.

Dropped that block (module + the corresponding recipe fields), keeping only the `replace` import hoist, which stands on its own. After the revert, `check-recipe-targets` is green (2548 `_target_` paths resolve).

> This one also illustrates CLAUDE.md §5: **when behind `main`, cherry-pick onto latest `main` first to check for duplicate work** — upstream had already solved offline reward checkpoints in a better way.

---

### Still worth the author's confirmation (I did not change these)

- `hsdp_shard_size` defaulting to 8, combined with the new fail-fast, turns `fsdp_mode: hybrid` on world ∈ {1,2,4,8} from "silently degrade to flat" into "hard error". That's a breaking change for existing single-node 8-GPU hybrid configs. The Compatibility section says "defaults to 8, preserving existing hybrid-mode behavior", which isn't accurate for world=8 (old behaviour: runs flat; new behaviour: refuses to start). I think erroring is the right call — silent degradation is worse — but it's worth spelling out in Compatibility.
- The `reward 0.3746 / grad norm 0.0001 / ratio 1.0000` numbers in the Test Plan are from a 2-step, 1-transition topology gate and are not convergence evidence — the PR already states this plainly, which is good.

Ran locally: `ruff check` / `ruff format --check` / `check-docstring-lines` / `check-core-dependencies` / `check-recipe-targets` all green across the 14 changed files. Lint CI won't re-run on my push without a maintainer approving the fork PR workflow.

AI assistance was used for this review and the fixes above; every claim is backed by a measurement or a code location.

leviking98z-rgb:
## Follow-up: found and fixed one more shared-gloo-process-group hazard (`0f311e8`)

I'd skipped checking the other callers of `init_gloo_group()` in the review above. Having done that, there's a substantive hazard:

`init_gloo_group()` (`unirl/utils/distributed_utils.py:46-51`) memoizes a **single process-wide** gloo group. `BaseFSDP2Backend._save_dcp` (`base_backend.py:414-421`) hands **that same group** to `dcp.async_save(..., process_group=pg)`'s background thread and holds it for the save's lifetime. This PR points the per-embed `monitored_barrier` at the same memo, so the main thread's barrier and the background save thread can **drive one process group concurrently** — undefined behaviour.

Measured on 4 ranks (nccl default PG, plus a background thread continuously all-reducing on the memoized group):

```
[iso] both gloo groups created
[iso] DEDICATED barrier OK x5 while bg thread drove the memoized group
[iso] bg thread errors on memoized group: ['RuntimeError: ... Connection closed by peer']
```

That is: **a dedicated group is unaffected (5/5 pass), while the shared group itself errors under concurrency.** Changed it to build its own group via `dist.new_group(backend="gloo")`.

Triggering it requires `checkpoint_format: dcp` + `checkpoint_async: true` + `text_encoder_onload_for_embed: true`. The first two default to `"torch"` / `False`, and **no shipped recipe turns them on together**, so this is latent rather than active — but this barrier is the memo's first hot user (it runs on every `embed()`), so it's worth isolating now.

---

### Two other findings I don't think need changes (FYI)

**Barrier symmetry holds, but only incidentally — nothing enforces it.** `embed()` runs `gen.batch_size` = `(batch_size/rollout_dp_size) × samples_per_prompt` times per rank per rollout (4 for both H3 recipes). On the sync path it can't diverge: `pytree_chunk` forces divisibility for DP_SCATTER, and a non-divisible batch raises driver-side *before* any RPC, so no rank ever enters the barrier alone. The eval path is likewise a full-rank collective per chunk.

But nothing in `text_embed.py` declares "my caller must be collective". The async/agentic paths dispatch via `slot.launch("generate_on_slot", ...)`, which targets a **single worker**; under that dispatch, per-rank `embed()` counts diverge immediately and the barrier deadlocks for 30 minutes. The only thing preventing it today is the indirect guard in `_build_rollout` that refuses the trainside engine on a separate slab (`diffusion.py:588-600`). Worth a comment on `_synchronize_embedding_ranks` recording that precondition, so nobody relaxes that guard and trips over this.

**The 30-minute barrier timeout coupling is worth noting, but the budget is comfortable.** The barrier is per-`embed()`, so the whole world waits on the slowest rank, while `_serialize_node_residency` deliberately serializes 8 × ~64GB onloads per node. Measured H2D on this node: 7.0 GiB/s pageable, 52.5 GiB/s pinned → 64GiB × 8 serialized is about **1.2 minutes** (pageable figure), well inside the 30-minute budget. The PR comment notes concurrent onloads went "hour-scale", which is exactly what motivated the lock; serialized, the measured bandwidth is safe.

One side effect worth knowing: with this barrier in place, any rank-local failure (OOM, the batch-1 `require` at `diffusion.py:151-156`) turns a fast crash into "every other rank waits 30 minutes". Acceptable, but easy to misdiagnose as a hang.

leviking98z-rgb (approved):

PR #411, opened 2026-09-02
feat(diffusionnft): implement the reference-model KL penalty

nussejzz:
## Summary

`DiffusionNFTConfig.kl_coef` has existed since the algorithm landed, but
`__init__` rejected any positive value:

```
DiffusionNFT: kl_coef > 0 not supported (KL penalty against base model not
implemented in this revision).
```

This implements it. Nothing in the framework was missing —
`_resolve_reference_model` in `algorithms/base.py` is what FlowDPPO already
uses, `adapters_disabled` in `train/lora.py` already routes a PEFT model
through its frozen base weights, and the trainer already injects `backend`
because DiffusionNFT declares `requires_ema_rollout`. Only the code connecting
them was absent, and the reason was cost: the reference is a third forward per
timestep, alongside the trainable and shadow ones.

Two choices worth flagging for review:

- **The reference is the LoRA-disabled base policy, not the EMA shadow.** The
  shadow follows the policy by construction, so anchoring to it bounds nothing.
- **The penalty is measured on the prediction, not on the reconstructed `x0`.**
  Both predictions are means of Gaussians sharing a variance at that timestep,
  so the KL reduces to their squared difference; taking it on the prediction
  keeps the penalty independent of `t`, which `xt - t * pred` is not.

Default is unchanged at `0.0`, so no existing recipe changes behaviour and the
extra forward is only paid when a recipe asks for it.

## Related Issue

N/A

## Test Plan

Environment: MiniMax-H3 t2va (33B, LoRA rank 64, `beta` 0.1, `adv_clip_max` 5),
4 nodes x 8 H20, 8 rollout replicas at DiT TP2 x Ulysses2, 8 prompts x 16
samples, reward = CLAP 1.0 + ImageBind 1.0 (`audio_video`) with **no**
visual-quality term — i.e. deliberately the reward that is easiest to push
off-distribution.

- Full `pre-commit` suite on the changed files — clean. The first revision passed
  `ruff` but failed `check-docstring-lines`; that is fixed.
- `kl_coef=0`: `_ref_model` resolves to `None`, no reference forward runs, and
  `ref_prediction_deviation` is absent from the metrics. Behaviour is
  bit-identical to before this change.
- `kl_coef=0` full run (112 rollouts): video collapsed into saturated noise by
  roughly rollout 20 while the reward climbed 0.13 -> 0.41 — entirely on
  ImageBind (0.157 -> 0.34) while CLAP *fell* 0.11 -> 0.06. ImageBind in
  `audio_video` mode is a bare cosine between two streams the policy generates,
  with the prompt absent (`need_text` is False), so it can be raised by moving
  both off-distribution together.
- `kl_coef=1e-4` (94 rollouts so far, ongoing): video intact throughout, and the
  gain moved onto the prompt-anchored component — CLAP +0.043 against ImageBind
  +0.021. `ref_prediction_deviation` rises 0.008 -> 0.057, so the term is active
  rather than inert. An independent VideoPickScore probe, weighted `1e-6` so it contributes
  ~3e-7 of a total near 0.3 and cannot influence the gradient, stayed flat at
  0.78 across all 94 rollouts.
- Timing: train phase +24%, not the +50% a third forward suggests, because the
  reference runs under `no_grad` and needs no backward.

Not run: no unit test is included. `tests/` was deliberately removed from this
repo in #267 and `pyproject.toml` carries no `[tool.pytest]` section, so adding
one back for a single feature seemed wrong; I have local coverage (reference
resolution at `kl_coef=0` vs `>0`, the negative-coefficient and missing-backend
rejections, and that the penalty is strictly positive when the adapter is not a
no-op) and can attach it in whatever form you prefer.

## Compatibility / Risk

- Default `kl_coef=0.0` short-circuits before any reference work, so existing
  DiffusionNFT recipes (SD3, Qwen-Image, Qwen-Image-Edit-Plus) are unaffected.
- `kl_coef > 0` requires a LoRA recipe and an injected `backend`;
  `_resolve_reference_model` already raises with an explanatory message for both,
  and a negative coefficient is now rejected explicitly.
- Adds one forward per trained timestep when enabled. Under
  `train_timestep_mode: all` that is once per timestep in the K-loop, so the
  cost scales with the schedule length.
- `ref_prediction_deviation` and `kl_coef` are new metric keys, present only when
  the penalty is enabled.

## Reviewer Notes

- The measurement above is the reason I think this is worth having beyond
  feature parity: with this reward the collapse is not a tuning artefact, it is
  what the reward's own optimum looks like, and the penalty was what moved the
  optimiser onto the prompt-anchored term instead.
- It does not make a reward unhackable. It taxes the distance to the exploit
  rather than removing it, and verl-omni's own H3 DiffusionNFT recipe runs
  `ref_kl_coef=1e-4` with its `ref_kl_loss` climbing 0.02 -> 0.22 over 60 steps,
  so a policy can still drift a long way under this term.
- A forward-process objective has a mismatch this term does not address: it
  trains on `xt = (1-t)*x0 + t*noise`, while inference at the same `t` sees a
  latent that walked the denoising trajectory. Those distributions differ most
  at low `t`, and separately measured, moving the trained timesteps down (mode
  `random`, mean t 0.885 -> 0.514) made a visual-quality probe drift down 0.8
  sigma while `ref_prediction_deviation` rose 8x.
- Written with AI assistance (Cursor).

## Review follow-ups (pushed by @CjhHa1 via maintainer edits)

Three commits on top of the original, all confined to this PR's own surface:

- `282dd75b` — one-line the `_reference_deviation` docstring and move its rationale
  into `unirl/algorithms/README.md`; `check-docstring-lines` has no allowlist, so
  the 12-line version was a hard CI failure. Also threads a `coef_name` argument
  through `_resolve_reference_model` so DiffusionNFT's errors name `kl_coef` rather
  than `beta`: DiffusionNFT has its own `beta` that must be `> 0`, so the old text
  told a non-LoRA user to "set beta=0", which then raised.
- `f67d2bbc` — emit `kl_coef` beside the penalty metric, as FlowGRPO and FlowDPPO
  emit `beta` beside `kl_ref_mean`; move `_resolve_reference_model` below the scalar
  checks, so a config wrong in two places reports the cheap failure first and a 33B
  model is not walked before the rest is known sane; record in the README that the
  cost is not only the +24% wall-clock, since the penalty competes with the reward
  gradient and settles at a lower proxy reward for the same step count.
- `88118b80` — rename the metric to `ref_prediction_deviation` and the method to
  `_reference_deviation`. It carries no `/(2σ²)`, so it is not FlowGRPO/FlowDPPO's
  `kl_ref_mean`, and `_reference_kl` sat one word from `base.py`'s shared
  `_reference_kl_loss`, which *is* the σ-normalized one. It is the same formula as
  the neighbouring `prediction_deviation` with the anchor swapped from the EMA shadow
  to the LoRA-disabled base, so the two now share a readable scale. `kl_coef` itself
  is untouched — it predates this work and six recipes pin it. Note the dashboards
  behind the numbers above were logged under the old `ref_kl_loss` key.

Three review points were documented in `unirl/algorithms/README.md` rather than
changed, since each would alter what `kl_coef` means for a run already in flight:
the missing σ normalization, the `kl_coef / adv_clip_max` coupling, and the choice
of a static anchor. The last one is the most interesting as follow-up work — a
periodically refreshed reference is better supported than a fixed one — but the
anchor identity lives in `_resolve_reference_model`, which FlowGRPO and FlowDPPO
share, so it belongs in its own PR.

## Checklist

Made with [Cursor](https://cursor.com)

### DIFF
--- a/examples/diffusion/minimax_h3/minimax_h3_t2va_nft.yaml
+++ b/examples/diffusion/minimax_h3/minimax_h3_t2va_nft.yaml
@@ -154,7 +154,7 @@ algorithm:
     path: unirl.models.minimax_h3.conditions.MiniMaxH3Conditions
   params: ${sampling}
   beta: 0.1
-  adv_sat_std: 5.0
+  adv_std_saturate: 5.0
   adv_mode: raw
   use_adaptive_weight: true
   # Read the rollout's own 10-transition schedule and train on all of it; Flow-GRPO
--- a/examples/diffusion/minimax_h3/minimax_h3_t2va_nft.yaml
+++ b/examples/diffusion/minimax_h3/minimax_h3_t2va_nft.yaml
@@ -163,8 +163,8 @@ algorithm:
   shuffle_train_timesteps: true
   apply_time_shift_in_loss: false
   training_timestep_fraction: 1.0
-  # A reference-model KL term would cost a third forward per timestep and is not
-  # implemented; the algorithm rejects kl_coef > 0.
+  # kl_coef > 0 anchors the policy to the LoRA-disabled base with a third no_grad
+  # forward per trained timestep; see unirl/algorithms/README.md. Default off.
   kl_coef: 0.0
 
 stack:
--- a/examples/diffusion/minimax_h3/minimax_h3_t2va_nft.yaml
+++ b/examples/diffusion/minimax_h3/minimax_h3_t2va_nft.yaml
@@ -163,9 +163,9 @@ algorithm:
   shuffle_train_timesteps: true
   apply_time_shift_in_loss: false
   training_timestep_fraction: 1.0
-  # kl_coef > 0 anchors the policy to the LoRA-disabled base with a third no_grad
+  # ref_deviation_coef > 0 anchors the policy to the LoRA-disabled base with a third no_grad
   # forward per trained timestep; see unirl/algorithms/README.md. Default off.
-  kl_coef: 0.0
+  ref_deviation_coef: 0.0
 
 stack:
   _target_: unirl.train.stack.TrainStack
--- a/examples/diffusion/minimax_h3/minimax_h3_t2va_nft.yaml
+++ b/examples/diffusion/minimax_h3/minimax_h3_t2va_nft.yaml
@@ -154,7 +154,7 @@ algorithm:
     path: unirl.models.minimax_h3.conditions.MiniMaxH3Conditions
   params: ${sampling}
   beta: 0.1
-  adv_clip_max: 5.0
+  adv_sat_std: 5.0
   adv_mode: raw
   use_adaptive_weight: true
   # Read the rollout's own 10-transition schedule and train on all of it; Flow-GRPO
--- a/examples/diffusion/qwen_image/qwen_image_nft.yaml
+++ b/examples/diffusion/qwen_image/qwen_image_nft.yaml
@@ -119,7 +119,7 @@ algorithm:
   params: ${sampling}
   # DiffusionNFT loss hyperparameters (v1 algorithm.algorithms.diffusion block).
   beta: 1.0
-  adv_sat_std: 5.0
+  adv_std_saturate: 5.0
   adv_mode: raw
   use_adaptive_weight: true
   train_timestep_mode: all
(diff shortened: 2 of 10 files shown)


CjhHa1 (approved):
LGTM

PR #415, opened 2026-09-03
feat(rollout): extend FastVideo runtime integration to WAN 2.2

Zcchill:
## Summary

This PR supersedes [#308](https://github.com/Tencent-Hunyuan/UniRL/pull/308).

It preserves the WAN 2.1 FastVideo runtime-patch integration from that draft and extends it to support WAN 2.2 A14B T2V, including its dual-expert architecture and substantially larger model lifecycle requirements.

The main review focus is the WAN 2.2 delta described below.

## What remains unchanged from #308

The following WAN 2.1 functionality is preserved:

- pinned upstream FastVideo integration through UniRL-owned runtime patches;
- native and replay old-log-probability modes;
- Flow and Dance SDE trajectory handling;
- fail-closed FastVideo compatibility checks;
- runtime patch propagation to spawned workers;
- WAN 2.1 single-transformer model adaptation.

No intentional numerical change is made to the validated WAN 2.1 DanceGRPO path.

### 1. WAN 2.2 model adapter

A dedicated WAN 2.2 adapter handles model-specific behavior without adding WAN 2.2 branches to the generic engine:

- validates `boundary_ratio` and `num_train_timesteps`;
- forwards `guidance_scale_2`;
- supports both `transformer` and `transformer_2`;
- builds and validates WAN 2.2 FastVideo requests;
- preserves generic engine behavior for WAN 2.1.

### 2. Dual-expert weight synchronization

WAN 2.2 A14B contains two transformer experts selected by the denoising boundary.

Weight synchronization now:

- separates high-noise and low-noise expert weights;
- routes them to `transformer` and `transformer_2`;
- updates layerwise CPU parameter caches directly;
- retains the existing single-transformer path for WAN 2.1.

### 3. A14B loading and offload support

Loading two 14B experts temporarily exceeded GPU memory even with upstream layerwise offload enabled.

The added offload patch:

- loads layerwise-offloaded modules on CPU initially;
- prevents temporary full-model GPU staging;
- releases transient CUDA cache between large components;
- keeps only active layers on GPU during execution.

These changes are activated by the relevant offload configuration and do not alter sampling equations.

### 4. Large-model IPC handling

Passing an A14B state dictionary through multiprocessing pipes exhausted `/dev/shm`.

The weight synchronization path now sends only the checkpoint path. Each FastVideo worker loads its assigned weights directly from that path.

Worker response payloads are also converted to CPU before IPC, avoiding GPU tensor ownership and CUDA IPC lifetime issues.

### 5. Colocated lifecycle stabilization

WAN 2.2 exposed lifecycle races that were not visible in shorter WAN 2.1 tests:

- FastVideo workers are joined and forcibly terminated if necessary;
- CUDA cache and IPC handles are collected during engine sleep;
- trainer startup waits for rollout-worker memory release;
- TCPStore bind collisions are retried;
- frozen trajectory dataclasses are copied without mutating their fields.

These changes affect resource ownership and process lifecycle, not reward or log-probability calculations.

### 6. WAN 2.2 recipe and model loading

The WAN 2.2 FastVideo recipe inherits the official trainside DanceGRPO configuration and changes only rollout-engine-specific settings.

Trainer-side VAE loading can be disabled when FastVideo owns decoding, reducing colocated memory use without changing the training model.

## Files reviewers should focus on

WAN 2.2 model behavior:

- `unirl/rollout/engine/fastvideo/adapters/wan22.py`
- `examples/diffusion/wan22/wan22_t2v_14b_dancegrpo_fastvideo.yaml`
- `tests/rollout/test_fastvideo_wan22.py`

Dual-expert synchronization and memory handling:

- `unirl/rollout/engine/fastvideo/_patches/weights.py`
- `unirl/rollout/engine/fastvideo/_patches/offload.py`
- `unirl/rollout/engine/fastvideo/_patches/conditions.py`

Shared lifecycle changes:

- `unirl/rollout/engine/fastvideo/backends/native.py`
- `unirl/rollout/engine/fastvideo/_patches/multiproc.py`

## Validation

Focused runtime-patch, WAN 2.2 adapter, and UniPC tests:

- 26 focused tests passed after rebasing onto the latest target branch.
- Python compilation and diff checks passed.

Formal aligned WAN 2.2 A14B experiments on 8xH20:

Clipboard_Screenshot_1788770734

Clipboard_Screenshot_1788770800

## WAN 2.1 compatibility

WAN 2.2-specific behavior is selected through the model adapter and dual-expert capability checks.

For WAN 2.1:

- no `transformer_2` routing is enabled;
- no WAN 2.2 boundary logic is used;
- the existing single-transformer weight path remains active;
- shared IPC and lifecycle changes preserve tensor values.

A WAN 2.1 native/replay smoke step will be rerun to verify that the shared infrastructure changes introduced no regression.

## Test Plan

- [x] Runtime patch unit tests
- [x] WAN 2.2 adapter tests
- [x] UniPC strategy tests
- [x] Three-way aligned WAN 2.2 comparison
- [x] Trainside baseline: 88 rollouts
- [x] FastVideo native: 31 rollouts
- [x] FastVideo replay: 26 rollouts
- [x] Create the cleaned and rebased PR branch
- [x] Rerun focused tests after rebase
- [ ] Run one WAN 2.1 native and replay training step after rebase


leviking98z-rgb:
Rebased this branch's WAN 2.2 work onto current `main` and pushed it here. (An earlier version of this comment pointed at #425 — that PR is now closed and everything lives in this one, with you as author.)

**Why the branch needed updating:** it forked at `50256793` (2026-07-22) and `main` has moved 105 commits since. In that window `main` gained the canonical UniPC solver (#251, hardened by #374), `unirl/types/rollout_req|resp` was replaced by `Sample`/`Part` (#214), `utils/scheduler_utils.py` moved to `unirl/sde/index_schedule.py` (#376), and three new pre-commit guards landed (#316 et al).

**How, and what it did not do:** `main` is *merged* in, not rebased over. Your 13 commits are untouched and still in the history; the follow-up work is stacked on top as ordinary commits, and the push was a fast-forward. `backup/pr415-original-20260909` pins the pre-merge state if a rollback is ever wanted. Conflicts were resolved toward `main` throughout — an AST and numeric comparison showed this branch's UniPC copy and `main`'s produce bit-identical trajectories, so `main`'s (with #374's frozen `UniPCSpec` and checkpoint-verified solver SSOT) is kept rather than landing a second solver.

**The part of your design I initially misjudged, and want to be explicit about.** I first read `_patches/contracts.py`, `denoising.py` and `conditions.py` as duplicating what `main` already had, and dropped them. That was wrong three times over. Verified against the pinned upstream `hao-ai-lab/FastVideo@2095477e`:

- `sde_step_with_logprob` takes 9 parameters there — no `eta`, no `sde_type`.
- `ForwardBatch.RLData` has neither `sde_step_indices` nor `sde_type`.
- There is no weight-update API at all on `Worker` / `MultiprocExecutor` / `VideoGenerator`.
- `MultiprocExecutor.execute_forward` rebuilds a bare `ForwardBatch` and drops `rl_data`, the trajectory, and the prompt embeddings.
- Its stock transition uses a third diffusion law (`sigma_min + (sigma_max - sigma_min) * sigma`, `std=0.6400` at `sigma=0.8`) matching neither UniRL Dance (`eta`, `0.7000`) nor Flow.

So `main`'s current `_unipc.py` only works against the fork, and UniRL owning those contracts — your approach — is what makes unmodified upstream possible. All three patches are re-landed here against the current interfaces.

**Two bugs fixed on the way, both measured on hardware:**

1. `_patches/weights.py` wrote `state.cpu_named_parameters` directly. Upstream's layerwise hooks are circularly linked, so the last block's `pre_forward` prefetches block 0 onto GPU and `wait_and_replace_params` reuses that cached copy, leaving block 0 stale for one forward. On a real `WanTransformer3DModel`: `stale_after_sync=44` (all of `blocks.0` in both experts) versus `0` through upstream's `mutate_params_scope()`. Currently masked because the colocate loop tears the generator down every step. Worth noting the naive fix is also wrong — holding every block's scope open at once stages the whole expert on GPU (measured peak 40 blocks vs 1).
2. `Wan22FastVideoAdapter.validate` only checked `model_index.json` `if model_index.is_file()`, so a missing file passed. Now required.

**Dropped:** the 30 s unconditional `time.sleep` in `backends/native.py`. `trainer/diffusion.py` calls `sleep()` after every `generate`, so that was 30 s of stall per training step. Measured on both WAN 2.1 and A14B, sleep returns the memory in full and wake succeeds with no grace period (`SLEEP_OK in 14.1s`, free VRAM back to `9.58G`, `WAKE_OK in 64.3s`).

**A14B end-to-end now passes** — boot, rollout, 2190-key dual-expert sync, sleep/wake, rollout again, on unmodified upstream. The pair costs ~6 GiB of CUDA (`9.58 → 3.61 GiB` free); without the offload patch upstream stages 106.5 GiB. Full numbers are in the PR body. One caveat on evidence: trajectory diffs cannot show a weight sync worked, because upstream hard-codes `generator=None` in the RL branch and a same-weights control differs by `d_traj=2.75e+00`; the sync is instead evidenced structurally (`2190 -> 1095/1095` split, `coverage=1.0`, `stale_after_sync=0`).

The merge also surfaced two pieces of drift now fixed: `TIMESTEP_SCALE` ended up declared twice in `WAN21DiffusionStep`, and `wan21_t2v_fastvideo_replay.yaml` still pointed at `unirl.utils.scheduler_utils.AllSDEScheduler`, which `check-recipe-targets` rejects after #376.

**Still open:** the pip pin plus retiring the fork from the compatibility allowlist. That changes pin policy for existing WAN 2.1 users, so it seems better as its own PR. Note the code's `_PINNED_FORK` now names upstream, so anyone still pointing `$FASTVIDEO_PATH` at `Zcchill/FastVideo@7fe1d7db` will be rejected by the fingerprints at init.

Happy to revert any of this if you would rather take a different route — the backup branch makes that a one-liner.

leviking98z-rgb (approved):

leviking98z-rgb:
Restored this PR's description to your original text — I had replaced it while stacking the follow-up commits, which dropped your two 8×H20 A14B curve screenshots. That was my mistake; the description is yours again and everything below stays in comments from here on.

Two things in it are stale after the merge, so here is the mapping rather than an edit.

### Your "files reviewers should focus on", at current paths

`main`'s `fastvideo/` layout won the merge (its `engine.py` speaks `Sample`/`Part` after #214, whereas this branch's still spoke `RolloutReq`/`RolloutResp`), so `adapters/` and `backends/` are gone as trees — but no capability is:

| your body says | now |
| --- | --- |
| `adapters/wan22.py` | `engine.py` — `_verify_dual_expert_checkpoint`, `_align_dual_expert_args`, `_low_noise_guidance`, `_model_timestep_scale`, all gated behind `_is_dual_expert` |
| `backends/native.py` | `engine.py` — the lifecycle is `main`'s, minus the 30 s unconditional `time.sleep` (see below) |
| `tests/rollout/test_fastvideo_wan22.py` | not committed — CLAUDE.md records the `tests/` tree was removed by #99/#267; harnesses are run and quoted instead |
| `_patches/weights.py`, `offload.py`, `conditions.py`, `multiproc.py` | **unchanged paths** — see next section |

### `_patches/` is back

I had flattened your subpackage into the engine directory during the merge. That was path dependency, not a decision: `main`'s `fastvideo/` held a single `_unipc.py` only because it had exactly one patch, and porting onto it collapsed four concerns into a file named after one of them. Both other engines quarantine upstream monkey-patches in a subpackage (`sglang_diffusion/_patches/`, 27 modules; `vllm_omni/patches/`), and so did you. Restored in `35cbf399`:

```
_patches/hijack.py      installer + install order
_patches/compat.py      stock-surface fingerprints, taken before any patch rewrites them
_patches/multiproc.py   spawn-child propagation
_patches/unipc.py       UniPC dispatch + the canonical-sigma set_timesteps wrapper
_patches/contracts.py   RLData field extension
_patches/denoising.py   the UniRL transition kernel
_patches/conditions.py  _patches/weights.py  _patches/offload.py
```

`timesteps.py` is the one module not restored: it fixed FastVideo's int64 timestep truncation at `TimestepPreparationStage`, and this branch fixes the same thing inside the patched `set_timesteps` (float32 `833.333` where upstream gives `833`, verified on a real scheduler), so a second seam would be redundant. `verify_fastvideo_used_sigmas` sits in `engine.py` rather than `_patches/` — it checks an engine response and patches nothing, which is also where `sglang_diffusion` and `vllm_omni` keep theirs.

That commit is a pure restructure: an AST comparison across the move reports 52/57 top-level symbols bit-identical after renaming, the 4 differing ones are exactly the installer/entrypoint functions the split reshaped, and the 5th differs only by a module name inside an error string.

### Verification, consolidated

Since the description is yours again, the evidence that was living in it goes here. Static gates on this branch:

```
python lint/check_docstring_lines.py           # 3379 docstrings, all one line
python lint/check_recipe_targets.py            # 2683 recipe _target_ paths resolve
python lint/check_core_dependencies.py         # ok
python lint/check_experimental_boundaries.py   # ok
ruff check unirl/rollout/engine/fastvideo/ unirl/models/wan22/   # All checks passed
ruff format --check unirl/rollout/engine/fastvideo/             # 14 files already formatted
```

Hardware — 8×H20 (driver 535), `torch 2.11+cu129`, **unmodified** `hao-ai-lab/FastVideo@2095477e`:

1. **WAN 2.2 A14B end-to-end, `ALL_DONE`.** Boot → rollout → 2190-key dual-expert sync → rollout → sleep → wake → re-apply → rollout. `1095` tensors per expert, `WEIGHT_SYNC_OK`, `SLEEP_OK in 14.1s` (free VRAM back to `9.58G`), `WAKE_OK in 64.3s` with no grace period, `REAPPLY_AFTER_WAKE_OK`, third rollout clean. The two experts cost ≈ 6 GiB of CUDA (`9.58 → 3.61 GiB` free); without the offload patch upstream stages `106.5 GiB`, impossible on a 97 GiB card. (A temporary version of the description said this was not run — it completed after that was written.)
2. **WAN 2.1 end-to-end**, same shape, 825 tensors, `ALL_DONE`.
3. **Transition math bit-exact against `unirl/sde/kernels.py`**: Dance and Flow × `eta` 0.7/0.3 × all 4 steps → `d_mean=0.0`, `d_logp=0.0` (16/16).
4. **UniPC dispatch equals FastVideo's own `scheduler.step()`**: `max_abs_diff=0.0` over 10 steps on a real `FlowUniPCMultistepScheduler` from the A14B `scheduler_config.json`. Control: an Euler tail differs by `6.79e-01`, so the comparison is discriminating.
5. **Patched `set_timesteps`**: sigmas consumed bitwise-verbatim, `index_for_timestep` round-trips 8/8, int-truncated echo rejected, duplicate / terminal-zero / out-of-range schedules all rejected.
6. **Dual-expert sync on a real `WanTransformer3DModel`** (real 18-regex `param_names_mapping`, real HF keys, real layerwise hooks): `coverage=1.0`, CUDA peak `1.01` blocks, `stale_after_sync=0`.

One caveat worth stating plainly: a trajectory diff cannot show a weight sync worked here, because upstream hard-codes `generator=None` in the RL branch, so SDE noise is unseeded — a same-weights control rollout differs by `d_traj=2.75e+00`, as much as a post-sync one. Item 6 is what establishes the swap reaches the model.

Still open, and I'd rather not fold it in here: the pip pin plus retiring the fork from the compatibility allowlist, which changes pin policy for existing WAN 2.1 users. Note `_PINNED_FORK` in the code already names upstream, so anyone still pointing `$FASTVIDEO_PATH` at `Zcchill/FastVideo@7fe1d7db` gets rejected by the fingerprints at init.

Issue #419, opened 2026-09-06
reward: no way to score a component without feeding it into the gradient (monitor-only rewards)


CjhHa1:
## What

There is no way to score a reward component without also feeding it into training. A
`RewardService` remote backend runs a panel of reward models and aggregates all of them into the
single reward the trainer z-scores into the advantage, so every component that gets scored also
shapes the gradient. Searching `unirl/reward/` for `monitor`, `probe`, `log_only`, or any
zero-weight escape hatch returns nothing.

## Why that's a gap

Detecting reward hacking requires at least one signal the optimiser is *not* pushing on. With no
support for that, the pattern gets approximated with a very small weight — #411 ran an independent
VideoPickScore probe at `weight: 1e-6` specifically so it would be logged while, in the author's
words, being unable to influence the gradient.

That approximation is close, but not exact, and the reasons are worth writing down.

`weighted_sum` normalizes by the weight total:

```python
if self.aggregation_method == "weighted_sum":
    total_w = sum(weights)
    return sum(s * w for s, w in zip(scores, weights)) / total_w if total_w > 0 else 0.0
```

`unirl/reward/remote.py:549`

So with CLAP 1.0 + ImageBind 1.0 + probe 1e-6 the probe's effective coefficient is
`1e-6 / 2.000001 ~= 5e-7`. Negligible in practice, but a real term in the objective rather than an
absent one, and it perturbs every other component's share on the way through.

The sharper problem is that the small-weight trick is silently unsafe under the other aggregation
methods, which ignore weights entirely:

```python
if self.aggregation_method == "mean":
    return sum(scores) / len(scores)
if self.aggregation_method == "min":
    return min(scores)
return max(scores)
```

`unirl/reward/remote.py:552`

Under `min`, a component weighted `1e-6` is a full participant, and becomes *the* reward outright
whenever it happens to score lowest. Anyone who adds a probe under `weighted_sum` and later
switches the panel to `min` — a reasonable thing to do, since `min` resists the trade-one-component-
for-another failure in #411 — would hand the objective to their monitor without any warning.

## Proposal

A `monitor: true` flag on a reward component spec: scored as usual, reported in the per-model
breakdown, excluded from `_aggregate_scores` and from `total_w`.

That makes the held-out-probe pattern exact instead of approximate, removes the `min`/`max` footgun,
and is the prerequisite for anything that wants to *act* on an unoptimised signal — for instance
[...]

Issue #418, opened 2026-09-06
reward(imagebind): the default mode scores no prompt, and it is the mode that collapses video under RL


CjhHa1:
## What

`ImageBindRewardScorer` defaults to `mode="audio_video"` in both places that set it:

- `unirl/reward/local/imagebind.py:48` — `DEFAULT_MODE = "audio_video"`
- `unirl/reward/local/imagebind.py:244` — `ImageBindSpec.mode: str = "audio_video"`

In that mode the prompt never reaches the model:

```python
need_text = self._mode in ("text_audio", "text_video", "all")   # :195
...
if self._mode == "audio_video":
    return cos(embeddings[ModalityType.AUDIO], embeddings[ModalityType.VISION])   # :223
```

So the score is a bare cosine between two streams the policy itself generates. Nothing in it is
anchored to what was asked for, which means it can be raised by moving audio and video
off-distribution *together*, with no prompt adherence required.

## Why it matters

#411 measured this. MiniMax-H3 t2va, reward = CLAP 1.0 + ImageBind 1.0 (`audio_video`) with no
visual-quality term, 112 rollouts:

- Total reward climbed 0.13 -> 0.41.
- The gain was entirely ImageBind: 0.157 -> 0.34.
- CLAP, the prompt-anchored component, *fell*: 0.11 -> 0.06.
- Video collapsed into saturated noise by roughly rollout 20.

The reward tripled while the output was destroyed, and the component doing the work was the
prompt-blind one. That is not a tuning artefact — it is what this reward's own optimum looks like.

## The safe mode already exists, it just isn't the default

`mode="all"` mixes the text terms in, at `{"audio_video": 0.5, "text_audio": 0.25, "text_video": 0.25}`
(`:52`). A recipe that writes `imagebind` without naming a mode silently gets the hackable
configuration rather than this one.

## Options

1. Flip the default to `all`. Cleanest, but changes behaviour for anyone relying on the current default.
2. Keep the default, but reject or warn when `audio_video` is used as a *training* reward while still
   allowing it for eval. `audio_video` is a legitimate way to *measure* AV sync; it is only dangerous
   as an optimisation target.
3. Documentation only — a warning in `unirl/reward/README.md`.

I lean 1 or 2, and am happy to send the PR once someone picks.

## Not the same as #386 / #412

Those report reward collapse that depends on the rollout engine — the same recipe trains steadily
[...]

PR #420, opened 2026-09-07
feat(minimax-h3): DiffusionNFT on the vLLM-Omni rollout backend at 32-GPU scale

nussejzz:
## Summary

Runs the DiffusionNFT objective for MiniMax-H3 t2va on the colocated vLLM-Omni
serving stack, where it had only ever run trainside.

DiffusionNFT is a forward-process objective: it never walks an SDE, it noises
the rollout's clean final latent to a chosen level and takes one MSE there. That
breaks an assumption the rollout path had baked in — that a non-empty
`sde_indices` is what distinguishes "this rollout feeds an algorithm" from "this
is an evaluation pass". With no SDE steps the two look identical, so the engine
returned decoded media only and the algorithm got a segment of `None`. The
request now says which it is; SDE steps still imply it, so Flow-GRPO recipes are
untouched and eval still infers the opposite from the same empty schedule.

**This PR is stacked on #378 and its diff contains #378's commits**, because a
PR here can only target `main`. Of the 21 commits and 39 files shown, 6 commits
and 8 files are new -- everything from `fix(train): make EMA LoRA training work
with a colocated external rollout` onward. Reviewing it against #378's head
rather than `main` shows only those. It is not mergeable before #378.

The forward-process stage API (`predict_noise_at_step`, `nft_clean_latents`,
dual-stream packing) is already upstream via #347, so it is not repeated here.

- **Train side.** `EmaLoraConfig` gained `module_prefix`, because unscoped
  suffixes adapted the token refiner too — 364 modules where the serving engine
  has slots for the denoising stack's 350, and weight sync rejected the payload.
  Scoping beats dropping the extras at sync time: an off-policy objective whose
  trainable set is wider than the set that generated its samples is not the
  objective the recipe asked for. Separately, `offload()` was the only thing that
  ever called `empty_cache()`, so a run keeping the train state resident — which
  an EMA algorithm must, since `_validate_residency_config` forbids offloading
  here — starved the rollout's physical allocation on the second wake.
- **Rollout contract.** The terminal clean latent is always captured, old-policy
  log-probs became optional, and a no-SDE schedule that carries them anyway is
  now an error. The algorithm also names the contract when a rollout hands it no
  segment at all, instead of dying on `'NoneType' has no attribute 'latents'`
  several frames from the cause.
- **Recipe.** 8 rollout replicas at DiT TP2 x Ulysses2, resident reward, FSDP32
  trainer. Standalone rather than layered on the Flow-GRPO chain, whose
  `old_logp_source` / `clip_range` keys Hydra would merge into this `algorithm`
  block and hand DiffusionNFT arguments it rejects.
- **Recipe defaults corrected** after the arms below showed the originals
  reproduce a known failure rather than run the documented objective. See
  Reviewer Notes.

## Related Issue

Depends on #378. The reference-anchor penalty landed separately in #411, which
renamed it on the way in: `kl_coef` is now `ref_deviation_coef`, `adv_clip_max`
is `adv_std_saturate`, and the metric `ref_kl_loss` is
`ref_prediction_deviation`. This recipe uses the new names and defaults the
penalty off, so nothing here requires it. The arms below that enabled it ran
before the rename and are labelled with the new one.

The arms below are also direct evidence for two issues opened out of #411's
review: #418 (ImageBind's default mode scores no prompt, and is the mode that
collapses video under RL) and #419 (no way to score a component without feeding
it into the gradient). The `1e-6` observer weight used below is exactly the
workaround #419 describes.

## Test Plan

Environment: 4 nodes x 8 H20 (32 GPUs), MiniMax-H3 33B t2va, 256x448 / 107
frames / 10 steps, prompts from the wan_dancegrpo train split.

Static:

- `ruff check` + `ruff format --check` on the changed files — clean.
- `python lint/check_docstring_lines.py` (3344 docstrings),
  `lint/check_recipe_targets.py` (2572 `_target_` paths resolve),
  `lint/check_core_dependencies.py`, `lint/check_experimental_boundaries.py` —
  all pass.
- `python -m unirl.train_diffusion --config-name=diffusion/minimax_h3/minimax_h3_t2va_vllmomni_32c_nft --cfg job --resolve`
  — resolves, exit 0.

Against the stock `vllm-omni==0.27.0rc1` wheel that `main` pins, the equivalent
Flow-GRPO gate on the parent branch (#378) completes its full lifecycle and
exits 0. **The forward-process gate has not run on that engine yet**: the head
node of the 32-GPU cluster has six reward workers wedged in uninterruptible
`lock_page` on the shared FUSE mount, so it cannot host a run until that clears.
Everything below was measured on the previous engine build.

Composing the recipe did surface three things that only the pinned release
rejects, all now fixed and statically verified:

- `num_frames: 107` is 4.46s at 24 fps, below the 5s floor the geometry
  validator enforces. It had only ever resolved through an
  `allow_nonstandard_canvas` opt-in that no longer exists. Every geometry block
  now resolves without one, at the same 124 frames the trainside recipes use.
  **All nine arms below therefore generated at a duration H3 does not support**,
  which is a candidate explanation for the flat visual reward in its own right.
- `autocast_precision` and `text_encoder_hidden_layer` are no longer fields on
  `MiniMaxH3PipelineConfig`; the layer is now a checkpoint-fixed constant.

Nine arms, roughly 700 rollouts, varying one thing at a time. `delta` compares
the mean of the first and last five rollouts of each arm:

| arm | rollouts | reward start -> end | delta | varied |
|---|---|---|---|---|
| formal200 | 112 | 0.2496 -> 0.3705 | +0.1209 | CLAP+ImageBind only, EMA warmup/flat=75 |
| visualgate30 | 30 | 0.9095 -> 0.8973 | -0.0122 | + PickScore 1.0 visual anchor |
| emaarm60 | 60 | 0.9039 -> 0.9660 | +0.0621 | EMA constant 0.9, flat=0 (contrast live) |
| cont80 | 20 | 0.9488 -> 0.9477 | -0.0011 | emaarm continued to 80 |
| ps200 | 95 | 0.7794 -> 0.7835 | +0.0041 | PickScore only |
| klarm60 | 60 | 0.2487 -> 0.3716 | +0.1228 | CLAP+ImageBind + ref_deviation_coef 1e-4 |
| pskl80 | 42 | 0.7794 -> 0.7856 | +0.0062 | PickScore + ref_deviation_coef 1e-4 |
| randt150 | 150 | 0.7791 -> 0.7752 | -0.0039 | PickScore+KL, random timesteps |
| klformal200 | 140 | 0.3452 -> 0.3756 | +0.0304 | klarm resumed to 200 |

The plumbing claims are backed by the NFT diagnostics rather than by the reward:

- `prediction_deviation` is the mean squared gap between the trainable and
  shadow predictions, so zero means the contrastive term contributes nothing.
  Under the original `ema_flat_steps: 75` it was **exactly 0.0** for all 30
  rollouts of `visualgate30` and 4e-05 at the end of `formal200` — NFT was
  running as advantage-weighted regression. With constant decay from step 0 it
  measures 0.006 (`emaarm60`, and varying) and 0.049 (`ps200`). The
  contrastive objective is live only after that change.
- `ref_prediction_deviation` appears and rises in every arm that enabled the
  penalty (`klarm60` 0.058, `klformal200` 0.066), so it is doing real work
  rather than being silently zero. These dashboards were logged under the old
  `ref_kl_loss` key, before #411 renamed it. Note it carries no `/(2σ²)`, so it
  is not on the same scale as FlowGRPO's `kl_ref_mean` — and per #411's own
  analysis the policy term carries an `adv_std_saturate / beta` gain, so this
  coefficient is not comparable to verl-omni's `ref_kl_coef` either.
- `x0_norm` stays in 0.96-1.39 across all arms, so the packed clean latent that
  `nft_clean_latents` builds is not drifting or exploding.

Not run:

- Regression of the trainside DiffusionNFT recipe, which shares
  `unirl/algorithms/diffusionnft.py` with this one. The only change to that file
  here is an added `None` guard ahead of an existing `None` guard.
- `sync.verify=true`, which #378 documents as blocked on
  vllm-project/vllm-omni#6351.

## Compatibility / Risk

- Behaviour is unchanged for any rollout that has SDE steps, which is every
  existing Flow-GRPO recipe: `record_trajectory` only adds a second way to ask
  for the trajectory, and the terminal position is one extra stored latent.
- `EmaLoraConfig.module_prefix` defaults to `""`, so existing EMA recipes keep
  their current unscoped behaviour.
- `release_cached_memory()` is a new backend method on the BROADCAST path; it
  fires only for a colocated external rollout that is not offloading, so no
  existing configuration reaches it.
- The added guard in `diffusionnft.py` can only fire where the previous code
  would have raised `AttributeError` on the next line.
- ImageBind is CC-BY-NC-SA; the recipe is NonCommercial and says so.
- This recipe's reward weights changed, which changes the numbers any future run
  of it produces. Nothing else consumes them.

## Reviewer Notes

**What these arms establish, and what they do not.** The implementation runs and
its levers do what they claim: the contrastive term activates and moves, the KL
penalty registers, the packed latent stays well-scaled, and 700 rollouts
completed without a contract failure. What none of it does is raise visual
reward.

The decisive number is not in the table above. `reward_videopickscore_mean` —
the only component that looks at the video without also looking at the audio it
was generated with — ends at 0.735, 0.774, 0.772, 0.772, 0.772, 0.768, 0.781 and
0.758 across the eight arms that measure it. A 0.046 spread, across arms that
differ in EMA schedule, KL on/off, PickScore-only versus composite, and schedule
versus random timesteps. Whatever is limiting H3 here is upstream of every lever
this PR adds.

One candidate for what that is came out of composing the recipe against current
`main`, and I would rather flag it than bury it: every one of those arms
generated 107 frames, which is 4.46s and below the 5s duration H3 supports. The
old opt-in that let the recipe past the validator made it easy not to notice.
Whether the flat visual reward is a consequence of training off-distribution is
untested — the arms would have to be rerun at 124 frames to say — but it is the
most concrete lead the nine arms produced, and it is not a lead about the NFT
implementation.

The two arms that rise substantially (`formal200` +0.121, `klarm60` +0.123) are
the two whose reward is CLAP plus ImageBind with no visual term. ImageBind in
`audio_video` mode scores a generation against *itself*, so that pair can be
climbed by degrading both streams in agreement — which is what the video did.
`emaarm60` also rose (+0.062) with a visual anchor present, but its PickScore
component ended at 0.774, i.e. the composite moved because the gameable
components moved.

That is why this PR changes the recipe defaults rather than only documenting the
trap. As committed, the recipe shipped `ema_flat_steps: 75` (contrast provably
inert) and `videopickscore: 0.0` (no visual term): a reader running it
reproduces the collapse. The new defaults exercise the objective the recipe
documents. They do not make the reward rise, and this PR does not claim they do.

Treat this as "DiffusionNFT runs correctly on this backend, and here is the
evidence that the remaining problem is not in it".

Written with AI assistance (Cursor).

## Checklist


Ideny42:
I addressed several MiniMax-H3 configuration and rollout wiring issues:

Require a local checkpoint for meta-init instead of passing an HF repo ID to the local-only sharded loader.
Remove the unsupported seed argument from quality100’s AllSDEScheduler.
Set the trainside NFT recipe’s ema_flat_steps to 0, preventing old and new from remaining identical for the first 75 refreshes.
Forward reward_num_frames so non-default YAML values reach the worker.
Keep sde_means=None for an empty SDE schedule instead of calling torch.stack([]).
Remove H3’s unused max_sequence_length config, which was consumed by neither the tokenizer nor the serving backend.
These changes are in f6aff3f, 157d0d2, and 521c217. All relevant pre-commit hooks pass. The FSDP input-precision concern is intentionally excluded and described separately below.

Ideny42:
I also noticed a potential FSDP input-precision issue in the new 32-GPU NFT recipe. I have not committed a fix for this yet.

MiniMax-H3 uses a mixed-dtype checkpoint: time_embedder is fp32, while the transformer blocks and AdaLN projections are bf16. The model forward applies SiLU to the fp32 temb first, then casts the activation to the projection dtype.

FSDPConfig.cast_forward_inputs defaults to true, so each fully-sharded block casts temb to bf16 before entering the block. This changes the AdaLN activation to bf16 across all 50 blocks, rather than preserving the model’s intended fp32 SiLU → bf16 projection ordering.

Both existing H3 trainside recipes explicitly set cast_forward_inputs: false. Should the new NFT recipe do the same? The corresponding hunk remains only in my local working tree pending discussion.

Ideny42:
Follow-up: I ran a controlled A/B where the only difference was cast_forward_inputs. As shown in the plot, the default true setting produces stronger reward growth, while false does not show comparable improvement.

Although false appears to preserve fp32 temb processing, it is not behavior-preserving in training. I will therefore not commit this change and recommend keeping the default true. Note that this is CLAP + ImageBind reward, so higher reward does not necessarily mean better visual quality.

Purple shows the default cast_forward_inputs=true; green shows false, with all other settings held constant.
Clipboard_Screenshot_1789634424

Ideny42 (commented):
Could you please confirm the intended cast_forward_inputs setting? Although false preserves the fp32 temb path, our controlled A/B shows significantly stronger reward growth with the default true.

PR #421, opened 2026-09-08
fix(reward): nest T2AV composite scorers and require a prompt-video term

CjhHa1:
## Summary

`T2AVCompositeScorer` rebuilt every inner spec from dataclass defaults and forwarded a hardcoded allow-list (`device`, `batch_size`, `frame_selection`). Inner-only fields such as `ImageBindSpec.mode` were unreachable from any recipe, so `imagebind` inside the composite was always `cos(audio, video)`. Combined with `clap` (prompt↔audio only), a T2AV objective can have **no term that relates the prompt to the video**.

This keeps the existing recipe shape: inner scorers are still constructed from `weights` names. The allow-list is gone. Overlapping composite fields (`device`, `batch_size`, `frame_selection`) still copy onto inners that declare them, so current YAML does not need to move. Additional inner fields go on an optional per-name map, not onto the composite spec:

```yaml
weights:
  imagebind: 1.0
  clap: 1.0
scorers:
  imagebind:
    mode: all
```

The composite refuses to start unless some positive-weight inner scorer returns `covers_prompt_video()`. `clap` and `imagebind` with `mode: audio_video` do not; `videopickscore` / `videoalign` / `videoclipdelta` do; `imagebind` does only for `text_video` or `all` with a nonzero `text_video` mix weight.

## Related Issue

N/A. Alternative to #404 (which only adds `mode` to the allow-list). Interacts with #403: that `imagebind`+`clap` recipe cannot start until it sets `scorers.imagebind.mode` to `text_video` or `all`.

## Test Plan

- `python3 lint/check_recipe_targets.py` — pass (pre-push)
- pre-push hooks (yaml, ruff, one-line docstrings, dependency directions) — pass
- One-off harness `/tmp/verify_t2av_inplace.py` (not committed; `tests/` removed by #99/#267):
  - default weights-only `videopickscore`+`clap` still starts
  - `frame_selection: middle` still forwards in-place to videopickscore only
  - `imagebind`+`clap` at default mode is refused
  - `scorers.imagebind.mode: all` starts without rewriting the recipe into nested `_target_` backends
  - unknown inner field is refused
  - `scorers` key outside `weights` is refused
  Result: `ALL PASSED`
- End-to-end T2AV run: Not run; reason: init/config plumbing plus a start-up guard; the in-tree LTX-2 recipe is still weights-only `videopickscore`+`clap`.

## Compatibility / Risk

Existing weights-only recipes keep working, including `frame_selection` on the composite. New: `imagebind`+`clap` without a prompt-video term fails at init. Downstream #403 must set `scorers.imagebind.mode`. `VideoRewardScorer` still uses a small allow-list; out of scope.

## Reviewer Notes

- Duplicate-work check: #404 is the two-line allow-list hotfix for the same hole. This PR does not extend that list.
- Cannot push onto celve's #404 branch; this is the in-place version on a new PR.
- AI-assisted implementation; the submitting human should still review the diff.

## Checklist

### DIFF
--- a/examples/diffusion/ltx2/ltx2_3_t2av_audioreward_trainside.yaml
+++ b/examples/diffusion/ltx2/ltx2_3_t2av_audioreward_trainside.yaml
@@ -12,8 +12,8 @@
 #   - clap           (audio-text alignment, LAION CLAP; zero extra deps)
 # ImageBind (audio-video alignment) is also available as an inner scorer but is
 # NOT enabled here: it requires the facebookresearch/ImageBind package and is
-# CC-BY-NC-SA 4.0 (NonCommercial). To enable, nest an ImageBindRewardScorer
-# under `scorers` with mode: text_video or all (audio_video does not relate
+# CC-BY-NC-SA 4.0 (NonCommercial). To enable, add `imagebind: <weight>` and set
+# `scorers.imagebind.mode` to text_video or all (audio_video does not relate
 # the prompt to the video and the composite will refuse to start).
 #
 # With audio_joint_sde=true, audio is part of the SDE policy AND now receives a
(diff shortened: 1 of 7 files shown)


CjhHa1:
Superseded: the in-place composite change was pushed onto #404 instead of a parallel PR.

PR #425, opened 2026-09-09
feat(rollout): run the FastVideo engine on unmodified upstream and add WAN 2.2 A14B

leviking98z-rgb:
## Summary

Supersedes the WAN 2.2 work in #415, rebased onto current `main` instead of the 7-week-old `engine/fastvideo` branch, and extends the FastVideo rollout engine to WAN 2.2 A14B T2V.

The load-bearing finding is that **`main`'s FastVideo integration cannot run on unmodified upstream FastVideo** — it only works against the `Zcchill/FastVideo` fork. Verified directly against the pinned upstream snapshot `hao-ai-lab/FastVideo@2095477e`:

- `sde_step_with_logprob` takes 9 parameters upstream; there is no `eta` and no `sde_type`.
- `ForwardBatch.RLData` has neither `sde_step_indices` nor `sde_type`.
- Upstream exposes **no** weight-update API at all (`Worker` / `MultiprocExecutor` / `VideoGenerator` all lack `update_transformer_weights*`).
- `MultiprocExecutor.execute_forward` rebuilds a bare `ForwardBatch` keeping only `output` / `logging_info` / `extra`, discarding `rl_data`, the latent trajectory, and the prompt embeddings the trainer replays against.
- Most importantly, upstream's stock transition uses a **third diffusion law** (`sigma_min + (sigma_max - sigma_min) * sigma`, i.e. `std=0.6400` at `sigma=0.8`) matching neither UniRL Dance (`eta`, `0.7000`) nor Flow (`sqrt(sigma/(1-sigma)) * eta`). Running unpatched means rollout and train-side replay score different SDEs, so the GRPO ratio is wrong at the root.

So UniRL now owns those contracts (`_contracts.py`, `_conditions.py`, `_weights.py`) and the fingerprints are taken on the **stock** surface before any patch rewrites it. `unirl/sde/unipc.py` and the UniPC dispatch are reused from `main` unchanged — this PR adds no second solver.

WAN 2.2 delta on top: dual-expert boundary routing (`boundary_ratio` pinned onto FastVideo's pipeline/dit configs and every request batch, checkpoint verified to declare both experts), `high_noise.*` / `low_noise.*` weight-sync routing, and an offload patch that keeps the two 14B experts off CUDA during materialization. WAN 2.1 behavior is unchanged: every WAN 2.2 path is gated behind `_is_dual_expert`, and `examples/diffusion/wan21/` is untouched apart from a stale dependency comment.

## Related Issue

N/A — supersedes the WAN 2.2 portion of #415.

## Test Plan

Static gates (all four repo pre-commit guards, on this branch):

```
python lint/check_docstring_lines.py       # 3371 docstrings, all one line
python lint/check_recipe_targets.py        # 2663 recipe _target_ paths resolve
python lint/check_core_dependencies.py     # ok
python lint/check_experimental_boundaries.py  # ok
ruff check unirl/rollout/engine/fastvideo/ unirl/models/wan22/   # All checks passed
ruff format --check unirl/rollout/engine/fastvideo/             # 9 files already formatted
```

Real-hardware validation — 8×H20 (driver 535), `torch 2.11+cu129`, **unmodified** `hao-ai-lab/FastVideo@2095477e`, `Wan2.1-T2V-1.3B-Diffusers` and `Wan2.2-T2V-A14B-Diffusers`:

1. **WAN 2.1 end-to-end rollout → weight sync → rollout** (`ALL_DONE`):
   ```
   BOOT_OK
   sigmas=[1.0,0.9722,0.9375,0.8929,0.8333,0.75,0.625,0.4167,0.0]
   sde_indices=(0,1,2,3)  unipc_indices=(4,5,6,7)
   [rollout-1] traj=(1,9,16,2,16,16) logp=(1,8) decoded=(1,3,5,128,128) sigma_roundtrip=OK
   loaded 825 transformer tensors -> perturbed ckpt (x1.01) -> WEIGHT_SYNC_OK
   [rollout-2] traj=(1,9,16,2,16,16) logp=(1,8) sigma_roundtrip=OK
   ```
   What this does and does not show: the sync applies all 825 tensors with full key
   coverage and the next rollout runs cleanly on the swapped weights. It does **not**
   show the trajectory moved *because of* the new weights. A same-weights control
   rollout differs by `d_traj=2.75e+00`, i.e. as much as the post-sync one, because
   upstream's RL branch hard-codes `generator=None` in `sde_step_with_logprob`, so SDE
   noise is unseeded and no caller can make a rollout reproducible. That the swap
   reaches the model is instead established by the offline check in item 5
   (`coverage=1.0`, `stale_after_sync=0` on a real `WanTransformer3DModel`).
2. **Transition math is bit-exact against `unirl/sde/kernels.py`**: Dance and Flow × `eta` 0.7/0.3 × all 4 steps → `d_mean=0.0`, `d_logp=0.0` (16/16).
3. **UniPC dispatch equals FastVideo's own `scheduler.step()`**: `max_abs_diff=0.0` over 10 steps on a real `FlowUniPCMultistepScheduler` built from the A14B `scheduler_config.json`. Control: an Euler tail differs by `6.79e-01`, so the comparison is discriminating.
4. **Patched `set_timesteps` on a real scheduler**: sigmas consumed bitwise-verbatim, `index_for_timestep` round-trips 8/8, float32 timesteps (`833.333`, vs upstream's `int64` `833`), int-truncated echo rejected, and duplicate / terminal-zero / out-of-range sigma schedules all rejected.
5. **Dual-expert weight sync on a real `WanTransformer3DModel`** (real 18-regex `param_names_mapping`, real HF key names from the A14B index, real layerwise hooks): coverage `1.0`, CUDA peak `1.01` blocks, `stale_after_sync=0`.
6. **A14B structure and offload budget** (safetensors headers + live observation): one expert `53.23 GiB` = 40 × `1340 MiB` + `0.87 GiB` stem/head. With the offload patch the two experts need ≈ `4.35 GiB` of CUDA; without it upstream stages `106.5 GiB`, impossible on a 97 GiB card. Observed live: worker at `105 GiB` host RSS holding `0 MiB` GPU.

Harnesses were run and discarded per CLAUDE.md (`tests/` tree removed by #99/#267); commands and results are quoted above rather than committed.

7. **Sleep/wake needs no CUDA grace period.** Full teardown then rebuild on WAN 2.1 with
   `UNIRL_FASTVIDEO_CUDA_RELEASE_GRACE_S` unset: `SLEEP_OK in 4.9s` (free VRAM back to
   `9.58G`), `WAKE_OK in 32.7s`, weights re-applied and a third rollout ran clean. This
   is why the 30 s unconditional `time.sleep` from the earlier draft is not carried here.
   Caveat: measured on WAN 2.1, not on A14B.

Not run: WAN 2.2 A14B end-to-end rollout / weight sync / sleep-wake. Reason: on the test node forced-spin held ~85 GiB per card, and loading A14B off CephFS ran at 5–14 MB/s because FastVideo's loader uses `safe_open` + per-tensor `get_tensor` (random 4 KB page faults over FUSE; `dd` reads the same file at 4.8 GB/s). The boot did not complete inside the allocation window. This is an environment/upstream-loader property, not a code path in this PR, but it does mean A14B is verified structurally, not end-to-end.

## Compatibility / Risk

- **Pin changes from the fork to unmodified upstream.** `_PINNED_FORK` is now `hao-ai-lab/FastVideo@2095477e`; the README and the WAN 2.1 recipe comment were still telling readers to check out `Zcchill/FastVideo@7fe1d7db`, which would now be rejected by the fingerprints at init. Anyone with the fork checked out must repoint `$FASTVIDEO_PATH`. No pip pin is added here — that belongs in a separate PR together with retiring the fork allowlist.
- **`unirl/sde/` is untouched.** The UniPC solver, `UniPCSpec` frozen-ness, and the model-config solver SSOT from #251/#374 are all reused as-is.
- **WAN 2.1 numerics.** The deterministic non-SDE tail already ran UniPC rather than the trainside `eta=0` Euler on `main`; this PR does not change that, and the 6.79e-01 Euler control above quantifies the gap for anyone comparing engines.
- New config surface: `model_family: wan2.2`, and `WAN22PipelineConfig` inherits the existing `unipc_*` fields (spelled out explicitly in the new recipe).
- `_weights.py` loads the sync checkpoint with `weights_only=True`. Every worker reads the whole file, so host RAM — not `/dev/shm` — bounds `sp_size`.

## Reviewer Notes

AI-assisted (Claude Code); every changed line was reviewed and the validation above was run and inspected rather than assumed.

Duplicate-work check: this supersedes the WAN 2.2 portion of #415 and deliberately does **not** re-land the UniPC solver that #251/#374 already put on `main` — an AST and numeric comparison showed #415's copy and `main`'s produce bit-identical trajectories, so `main`'s (with #374's hardening) is kept.

Worth inspecting first:

1. `_contracts.py` — the diffusion-law divergence is the correctness argument for this whole approach.
2. `_weights.py::_load_transformer_state` — it mutates layerwise-offloaded blocks through upstream's `mutate_params_scope()`, **one block at a time**. Writing `state.cpu_named_parameters` directly (as the earlier draft did) leaves block 0's prefetched GPU copy stale: measured `stale_after_sync=44` on a real DiT, all of `blocks.0` in both experts. Holding every block's scope open at once instead stages the whole expert on GPU (measured peak 40 blocks vs 1).
3. `_unipc.py::_verify_stock_surface` and the `_*_PARAMS` tuples — these must be updated together with any patch edit.

Follow-ups deliberately left out: the pip pin plus removing the fork from the compatibility allowlist (changes the pin policy for existing WAN 2.1 users, so it deserves its own PR), and the A14B end-to-end run once a node without forced-spin is available.

### DIFF
--- a/examples/diffusion/wan21/wan21_t2v_dancegrpo_fastvideo.yaml
+++ b/examples/diffusion/wan21/wan21_t2v_dancegrpo_fastvideo.yaml
@@ -18,9 +18,10 @@
 #      weights and pushes the merged full model (FastVideo serves merged weights;
 #      it has no separate adapter path here).
 #
-# FastVideo dependency: https://github.com/Zcchill/FastVideo at or after
-# 7fe1d7db9a0b8aebb46679e7924f597431f23665. Set $FASTVIDEO_PATH to that
-# checkout; FastVideo is lazy-imported by the rollout engine only.
+# FastVideo dependency: unmodified upstream https://github.com/hao-ai-lab/FastVideo
+# at 2095477eac7e289c7a7ab13acb367ca60687c304. Set $FASTVIDEO_PATH to that
+# checkout; FastVideo is lazy-imported by the rollout engine only. UniRL installs
+# the RL contracts itself (rollout/engine/fastvideo/README.md: The pin).
 #
 # This recipe defaults to replay log-probs. Native mode is enabled by setting
 # rollout.config.native_logprob=true and algorithm.old_logp_source=rollout;
--- a/examples/diffusion/wan22/wan22_t2v_14b_dancegrpo_fastvideo.yaml
+++ b/examples/diffusion/wan22/wan22_t2v_14b_dancegrpo_fastvideo.yaml
@@ -1,0 +1,100 @@
+# @package _global_
+# DanceGRPO WAN 2.2 T2V A14B with the FastVideo rollout engine, colocate, v2 trainer, 1x8.
+#
+# This is wan22_t2v_14b_dancegrpo.yaml with rollout moved from the in-process
+# trainside engine to the dedicated FastVideo VideoGenerator engine (+ the
+# weight-sync block a dedicated engine requires). Reward, LoRA targets, Dance
+# SDE kernel, replay log-prob, sigma schedule, and SDE-index window remain
+# aligned with the official recipe. The deterministic non-SDE tail intentionally
+# uses UniPC instead of the trainside recipe's eta=0 Euler steps, so generated
+# trajectories can differ (unirl/sde/README.md).
+#
+# WAN 2.2 A14B delta over the WAN 2.1 FastVideo recipe:
+#   1. `model_family: wan2.2` selects boundary-routed dual-expert rollout. The
+#      engine pins `boundary_ratio` onto FastVideo's pipeline/dit configs and
+#      verifies the checkpoint declares both `transformer` and `transformer_2`.
+#   2. Two 14B experts do not fit alongside the trainer, so the rollout runs
+#      with layerwise DiT offload plus CPU text-encoder/VAE offload.
+#   3. The weight sync ships a checkpoint path; each worker splits it on the
+#      `high_noise.*` / `low_noise.*` prefixes and loads both experts.
+#
+# FastVideo dependency: see unirl/rollout/engine/fastvideo/README.md for the
+# pinned surface. Set $FASTVIDEO_PATH to that checkout; FastVideo is
+# lazy-imported by the rollout engine only.
+#
+# This recipe defaults to replay log-probs. Native mode is enabled by setting
+# rollout.config.native_logprob=true and algorithm.old_logp_source=rollout;
+# both modes use the same SDE sampling window.
+
+defaults:
+  - wan22_t2v_14b_dancegrpo
+  - _self_
+
+weight_sync_interval: 1      # push freshly-trained weights into FastVideo every rollout (on-policy)
+
+# Shared nodes (lifted from the trainside recipe so the dedicated FastVideo
+# engine can be injected with the same model config + SDE kernel).
+model_config:
+  _target_: unirl.models.wan22.config.WAN22PipelineConfig
+  pretrained_model_ckpt_path: ${oc.env:PRETRAINED_MODEL,Wan-AI/Wan2.2-T2V-A14B-Diffusers}
+  model_precision: bf16
+  shift: 5.0
+  max_sequence_length: 512
+  # Sigma boundary routing the high-noise / low-noise experts; the engine pins
+  # the same value onto FastVideo's dit_config so both engines agree.
+  boundary_ratio: 0.875
+  num_train_timesteps: 1000
+  # Deterministic-solver SSOT: the engine verifies these against the
+  # checkpoint's scheduler/scheduler_config.json at init.
+  unipc_solver_order: 2
+  unipc_solver_type: bh2
+  unipc_lower_order_final: true
+  # FastVideo owns video decoding; the trainer-side pipeline only replays the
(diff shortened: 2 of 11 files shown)


leviking98z-rgb:
Closed in favour of #415, which now carries this same work with @Zcchill as author.

#415 was 105 commits behind `main`. Rather than rebasing it (which would have rewritten his history), `main` was merged into that branch: his 13 commits are untouched, this work is stacked on top, and the push was a fast-forward. Its base is now `main` and it includes the WAN 2.2 A14B end-to-end validation.

Branch `engine/fastvideo-wan22-onmain` has been deleted. `backup/pr415-original-20260909` pins the pre-merge state of #415 should a rollback ever be needed.

PR #426, opened 2026-09-09
perf(vllm-omni): restore cuDNN SDPA for colocated training

yuxiaooye:
## Summary

Importing vLLM disables cuDNN SDPA process-wide. In colocated mode, that import-side state can leak into the training actor and reroute train-side diffusion attention from cuDNN flash to the slower PyTorch memory-efficient CUTLASS backend.

During `VLLMOmniBackend.boot`, preserve the actor's existing cuDNN SDPA setting around UniRL patch installation and vLLM/vLLM-Omni runtime imports, then restore the exact original value. Spawned engine workers initialize their own process-local backend state.

Stable-step timing: vLLM-Omni `train_time_s` improved from ~170s to ~137s; the trainside baseline was ~130.5s. `generate_time_s` remained ~36s versus ~77s for trainside, and **end-to-end `step_time_s` improved from ~240s to ~205s.**

eb6336fb6cfbb8d686a3cd64a499e843

## Related Issue

N/A

## Test Plan

- Environment: HunyuanVideo-1.5, 8×H800, identical recipe, data, seed, batch size, and Diffusers native PyTorch SDPA backend across timing runs. Recipe: [`hunyuan_video15_t2v_vllmomni_colocate`](https://github.com/Tencent-Hunyuan/UniRL/blob/main/examples/diffusion/hunyuan_video15/hunyuan_video15_t2v_vllmomni_colocate.yaml).
- In the CUDA training environment, verified that importing `vllm.platforms.cuda` changes `torch.backends.cuda.cudnn_sdp_enabled()` from `True` to `False`, and the scoped restoration changes it back to the original value.
- Profiler: train attention forward and backward changed from PyTorchMemEffAttention/CUTLASS (~31.2s per update) to cuDNN flash (~16.2s per update).
- Local focused behavioral test covers initial `True`/`False` states and restoration on both success and injected exceptions: 4 cases passed. The test is intentionally not included in this focused PR.
- `python3 -m py_compile unirl/rollout/engine/vllm_omni/backends/native.py`
- `SKIP=no-commit-to-branch pre-commit run --files unirl/rollout/engine/vllm_omni/backends/native.py --show-diff-on-failure` — all applicable hooks passed.

## Compatibility / Risk

No API, configuration, checkpoint, or data-format changes. The code restores the exact pre-import cuDNN SDPA setting rather than forcing it on, so actors that already require cuDNN SDPA disabled remain disabled. Model-specific safeguards that disable it at their owning execution boundary remain unchanged.

## Reviewer Notes

Only `unirl/rollout/engine/vllm_omni/backends/native.py` is changed in the final diff. The preservation scope includes patch installation because importing the patch bundle also imports vLLM modules. Profiling artifacts and the focused local test are intentionally excluded.

## Checklist

### DIFF
--- a/unirl/rollout/engine/vllm_omni/backends/native.py
+++ b/unirl/rollout/engine/vllm_omni/backends/native.py
@@ -21,6 +21,19 @@ logger = logging.getLogger(__name__)
 _ENGINE_PROC_PREFIX = "VLLM::"
 
 
+@contextmanager
+def _preserve_cudnn_sdp_state():
+    """Restore the process-wide cuDNN SDPA setting after vLLM imports."""
+    import torch
+
+    enabled = torch.backends.cuda.cudnn_sdp_enabled() if torch.cuda.is_available() else None
+    try:
+        yield
+    finally:
+        if enabled is not None:
+            torch.backends.cuda.enable_cudnn_sdp(enabled)
+
+
 def _import_omni_runtime() -> Dict[str, Any]:
     """Lazy import of the vllm-omni runtime types. Imported once per process."""
     from transformers import AutoTokenizer
@@ -117,10 +130,6 @@ class VLLMOmniBackend:
     @classmethod
     def boot(cls, intent: Dict[str, Any]) -> "VLLMOmniBackend":
         """Spell the intent into ``Omni`` ctor kwargs and spawn."""
-        from unirl.rollout.engine.vllm_omni.patches import install as install_patches
-
-        install_patches()
-
         import multiprocessing as mp
 
         try:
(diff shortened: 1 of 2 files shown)


CjhHa1 (approved):
Reviewed the final diff against its merge base. The cuDNN SDPA state is preserved across both patch installation and vLLM runtime imports, restored on success and failure, and spawned workers retain independent process-local state. Focused local True/False and exception-path tests passed; PR checks are green.

CjhHa1:
@yuxiaooye Thanks again for finding the bug, I've revised the code for a better solution. You can check if you prefer.

PR #436, opened 2026-09-11
fix(diffusion): remove duplicate sampling fields

Yuanjia-Lee:
## Summary

Diffusion sampling carried two duplicate sources of truth:

- `samples_per_prompt` plus the legacy `num_samples_per_prompt` alias. `dataclasses.replace()` copied both fields, so any eval fan-out differing from a training fan-out greater than one could be silently restored to the training value.
- BAGEL inherited `guidance_scale` while separately storing and consuming `cfg_text_scale`, forcing recipes and eval code to keep both fields aligned.

Remove both duplicate fields. Runtime sampling now uses only `samples_per_prompt` for fan-out and `guidance_scale` for text CFG across every diffusion family. The vLLM-Omni adapter requires `BagelDiffusionParams` and translates `guidance_scale` to BAGEL's external `cfg_text_scale` worker argument at that boundary.

Redundant SenseNova and BAGEL recipe settings, model-family eval branches, duplicate adapter defaults, and the one-use CFG resolver are removed.

## Related Issue

N/A

## Test Plan

- Focused CPU checks cover `dataclasses.replace(..., samples_per_prompt=1)` for base and SenseNova params.
- Focused BAGEL checks cover canonical `guidance_scale`, eval overrides, CFG interval gating, wire translation, and rejection of generic diffusion params by the BAGEL adapter.
- Confirmed both retired sampling fields fail at construction/overlay boundaries.
- Hydra-instantiated every changed BAGEL sampling recipe successfully.
- `pre-commit run --all-files` — pass.
- Full pre-push hook suite — pass.

## Compatibility / Risk

This intentionally removes the sampling-level `num_samples_per_prompt` and `cfg_text_scale` configuration/API fields. Configurations using them now fail instead of maintaining duplicate runtime state. BAGEL's vendor/worker API still receives `cfg_text_scale`; the adapter derives it from canonical `guidance_scale`. The PE/Unified top-level `eval_cfg_text_scale` trainer knob remains unchanged and writes canonical `guidance_scale`.

Five shipped DiffusionTrainer recipes now honor their declared `eval_samples_per_prompt: 4` instead of silently using the larger training fan-out (BAGEL 8→4 and LTX2 16→4). This intentionally changes eval cost and sample aggregation to match configuration.

## Reviewer Notes

The final diff was reviewed against the PR base. Repository-wide scans found no additional sampling aliases. Remaining `cfg_text_scale` occurrences are the rejected retired eval key, the unchanged PE/Unified trainer-knob name, or BAGEL vendor/wire-boundary names—not duplicate runtime sampling state. Qwen `distilled_guidance_scale` remains because it controls a transformer guidance embedding independently of true CFG.

## Checklist

### DIFF
--- a/examples/diffusion/bagel/bagel_editreward.yaml
+++ b/examples/diffusion/bagel/bagel_editreward.yaml
@@ -176,15 +176,16 @@ data_source:
     algorithm:
       prompts_per_rollout: ${batch_size}
 
-# BAGEL edit sampling (BagelDiffusionParams). guidance_scale amplifies the INSTRUCTION,
-# cfg_img amplifies the SOURCE (cfg_img active only when guidance_scale>1; cfg_img>1 = 3rd forward).
+# BAGEL edit sampling (BagelDiffusionParams). guidance_scale amplifies the
+# INSTRUCTION; cfg_img_scale adds a SOURCE-CFG forward, whose result is combined
+# only when guidance_scale>1.
 sampling:
   _target_: unirl.models.bagel.diffusion.BagelDiffusionParams
   num_inference_steps: 14
-  guidance_scale: 1                    # 1 = No-CFG single forward (validated config). >1 enables CFG
+  guidance_scale: 1.0                  # 1 = No-CFG single forward (validated config). >1 enables CFG
                                        # (amplifies the instruction) = +1 forward/step -> more memory.
-  cfg_img_scale: 1.0                   # source<->diversity knob; only active when guidance_scale>1
-  cfg_interval: [0.4, 1.0]             # CFG schedule; only active when guidance_scale>1
+  cfg_img_scale: 1.0                   # >1 adds a source-CFG forward; keep 1 unless guidance_scale>1
+  cfg_interval: [0.4, 1.0]             # timestep interval for both CFG scales
   cfg_renorm_min: 0.0
   cfg_renorm_type: text_channel        # CFG renorm; only active when guidance_scale>1
   eta: 1.0                             # SDE noise scale (per-step stochasticity = GRPO exploration)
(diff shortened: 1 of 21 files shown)


CjhHa1:
@Yuanjia-Lee Thanks for identifying the eval fan-out bug. I pushed follow-up commits to this PR and changed the fix from synchronizing the alias inside `build_eval_sampling` to removing the duplicate runtime state at its source.

What changed:
- Removed `DiffusionSamplingParams.num_samples_per_prompt`; `samples_per_prompt` is now the sole fan-out field.
- Removed the alias lockstep branch from `build_eval_sampling` and the redundant SenseNova recipe overrides.
- Found the same duplicate-source problem in BAGEL: removed sampling-level `cfg_text_scale` and made `guidance_scale` canonical for trainside, eval, PE/Unified, and vLLM-Omni. The adapter translates it to BAGEL's required wire key `extra_args.cfg_text_scale` only at the external boundary.
- Updated all affected BAGEL recipes/docs, added a strict `BagelDiffusionParams` request contract in the adapter, and removed duplicate adapter defaults/fallbacks.

Why: keeping two stored names for one value makes every `dataclasses.replace()` call responsible for updating both. That is what caused `eval_samples_per_prompt` to be silently restored to the training fan-out, and it leaves the same failure mode available in future call sites. Keeping one canonical runtime field removes the bug class instead of patching one resolver. Stale configs now fail rather than creating divergent state.

Behavior note: all declared `eval_samples_per_prompt` values are now honored, not only `1`. Five shipped recipes therefore change from silently using the larger training fan-out to their configured eval value `4` (BAGEL 8→4 and LTX2 16→4).

Validation: focused base/SenseNova/BAGEL behavior checks, BAGEL wire-boundary checks, Hydra construction of changed recipes, full `pre-commit run --all-files`, pre-push hooks, and two independent deep reviews. Current remote checks are all green.

CjhHa1 (approved):
LGTM

PR #437, opened 2026-09-11
fix(ar): prepare replay anchors using the planned microbatches

Yuanjia-Lee:
## Summary

When `old_logp_source=replay`, DPPO/CPPO/DRPO already freeze π_old into `segment.log_probs`, but `recomputes_anchor()` stayed false, so TrainStack prepared the whole worker shard in one forward. They now declare it like GSPO, and replay uses the same micros as the training step.

## Related Issue

N/A

## Test Plan

- `ruff check` / `ruff format --check` on `unirl/algorithms/{dppo,cppo,drpo}.py` — pass
- Local smoke checks for `rollout` (still whole-shard), `replay` (follows the planned micros), and a missing declared field (raises). No test files included in this PR.
- Live train: Not run; reason: this is the `recomputes_anchor` declaration; TrainStack's sliced path is the existing GSPO / FlowGRPO one.

## Compatibility / Risk

No config, checkpoint, data format, or public API changes expected. Risk is limited to `old_logp_source=replay` on DPPO/CPPO/DRPO; default recipes stay on `rollout`.

## Reviewer Notes

No overlapping open PR touches `dppo.py` / `cppo.py` / `drpo.py`.

## Checklist

### DIFF
--- a/unirl/algorithms/README.md
+++ b/unirl/algorithms/README.md
@@ -10,7 +10,7 @@
   <img src="../../assets/algorithm-contract-new.png" alt="UniRL algorithm contract: a StageAlgorithm combines new_logp from replay, the frozen pi_old anchor, and advantages into a loss (four interchangeable families: GRPO, FlowDPPO, DRPO, and DiffusionNFT as the ratio-free exception), and declares knobs — requires_ema_rollout, supports_multi_update, anchor_fields/recomputes_anchor — that reconfigure the sampler and train stack around it" width="100%">
 </div>
 
-*A `StageAlgorithm` is two things: a **loss combine** (`stage.replay → new_logp`, mixed with the frozen **π_old** anchor and advantages — four interchangeable families) and a few **declared knobs** (`requires_ema_rollout`, `supports_multi_update`, `anchor_fields`/`recomputes_anchor()`) that reconfigure the sampler and the train loop around it.*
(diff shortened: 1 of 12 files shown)


CjhHa1 (approved):
LGTM

CjhHa1:
@Yuanjia-Lee I pushed a follow-up cleanup on top of this PR after reviewing the anchor contract.

What changed:

- `recomputes_anchor` is now an algorithm property rather than a callable hook. `StageAlgorithm` defaults it to `False`; PPO/FlowDPPO declare constant `True`; DPPO/CPPO/DRPO/GSPO/FlowGRPO expose a read-only property derived live from `old_logp_source`.
- Both train stacks read the property directly and validate the contract at construction: it must resolve to `bool`, and `True` must be paired with non-empty `anchor_fields` so per-micro results cannot be silently discarded.
- Removed redundant `getattr` fallbacks already guaranteed by the `StageAlgorithm` interface and updated the extension/anchor documentation.

Why:

`recomputes_anchor` is declarative algorithm metadata, alongside `anchor_fields` and `supports_multi_update`; a property makes that contract explicit without a trivial method call. Deriving it live from `old_logp_source` avoids caching duplicate state that could become inconsistent. The validation makes malformed implementations fail early instead of silently preparing anchors on temporary micro slices.

The original behavior fix is unchanged: DPPO/CPPO/DRPO with `old_logp_source=replay` prepare π_old on the exact planned training micros, while `rollout` behavior remains unchanged. Focused packed-varlen smoke checks, ruff/format/compile checks, and the full pre-push suite pass.

Commits: `8d968eb8`, `62b5f9f3`, `8249d2b9`.

PR #440, opened 2026-09-12
fix(checkpoint): preserve cold AdamW state during checkpoint export

Yuanjia-Lee:
## Summary

When `optimizer_step` skips and `maybe_save_checkpoint` still fires, DCP `get_optimizer_state_dict` dummy-steps the empty AdamW and leaves `state['step']=1` on the live process and in the file. `sharded_state` now writes `step=0` in the exported state and restores the live optimizer to empty. Loading that step-0 state preserves the first real AdamW update. Already-stepped optimizers are unchanged.

## Related Issue

N/A

## Test Plan

- `ruff check` / `ruff format --check` / `python lint/check_docstring_lines.py` on `unirl/train/backend/sharded_state.py` — pass
- CPU three-way (never-save / save-continue / save-resume) first AdamW update matches; raw DCP dummy-step does not. 2×4090 FSDP2 gather + `dcp.save`/`dcp.load` same. No test files included in this PR.

## Compatibility / Risk

No config or public API changes. Already-stepped checkpoints are unchanged. A never-stepped AdamW file now stores `step=0` rather than the dummy `1`; old dummy-`1` files are not rewritten on load.

## Reviewer Notes

No overlapping open PR touches `sharded_state.py`. Draft #432 still calls `sharded_optimizer_state_dict`.

## Checklist

### DIFF
--- a/unirl/train/backend/sharded_state.py
+++ b/unirl/train/backend/sharded_state.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import logging
-from typing import Dict, Iterator, Optional
+from typing import Dict, Iterator, List, Optional
 
 import torch
 from torch import nn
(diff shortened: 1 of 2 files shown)


CjhHa1 (approved):
LGTM

CjhHa1:
Maintainer follow-up summary of the changes I made on top of the original patch:

- Narrowed the workaround to a genuinely cold built-in AdamW: empty optimizer state and no pending gradients.
- Kept one shared export boundary for both gathered and sharded checkpoints. It resets the DCP-materialized `step` to zero and clears the dummy state from the live optimizer.
- Removed the load-side workaround. Stock DCP loading of the exported step-zero state already preserves the first real AdamW update, while correctly retaining checkpoint learning rates and rank-0 broadcast semantics.
- Removed the generic state-scanning helpers, private `_step_count` handling, LR restoration state machine, redundant `dict()` copy, and ineffective `TypeError` fallback.
- Shortened the README note to document only the observable behavior.

Verified locally with save/continue/resume AdamW comparisons, warm and cold two-rank broadcast checks, and the repository pre-commit suite. The regression tests were kept local rather than added to this PR.

PR #442, opened 2026-09-12
perf(qwen): project only the final token during trainside generation

Yuanjia-Lee:
## Summary

The Qwen trainside autoregressive loops only sample `logits[:, -1, :]`, but prefill projected every prompt position through `lm_head`. Pass the models' native `logits_to_keep=1` contract so Qwen3, Qwen2.5-VL, and Qwen3.5 materialize only `[B, 1, vocab]` logits.

Qwen3.5 applies the same selection on its cacheless VeOmni path, where the full growing sequence is forwarded each step. Replay remains unchanged because it needs every response-token log-probability.

## Related Issue

N/A

## Test Plan

- `ruff check` / `ruff format --check` / `python lint/check_docstring_lines.py` on the three changed AR modules — pass
- Qwen3-0.6B, Transformers 5.6: last-token logits and KV match the full forward within tolerance; prefill peak 2.50 -> 1.39 GiB (B=4, L=512, bf16)
- Transformers 5.6 tiny-model checks:
  - Qwen2.5-VL: final-token logits match the full projection within tolerance; KV length unchanged
  - Qwen3.5 cacheless path: final-token logits match the full projection within tolerance
- Live PE train: not run; sampling and cache-update paths are unchanged, and this matches the native `generate()` behavior.

## Compatibility / Risk

No config, checkpoint, state-dict, or public API changes. The repository pins Transformers 5.6, and the HF plus VeOmni Qwen forwards all support `logits_to_keep`. Prompts are left-padded, so the final row is each sample's last real token.

## Reviewer Notes

Open #430 edits `qwen3/ar.py` on the replay exact-actor path, not these generation-loop arguments.

## Checklist

### DIFF
--- a/unirl/models/qwen3/ar.py
+++ b/unirl/models/qwen3/ar.py
@@ -316,6 +316,7 @@ class Qwen3ARStage(ARStage[Qwen3ARConditions]):
                 past_key_values=model_kwargs.get("past_key_values"),
                 attention_mask=model_kwargs.get("attention_mask"),
                 use_cache=True,
+                logits_to_keep=1,
             )
             with torch.no_grad():
                 out = transformer(**model_inputs, return_dict=True)


CjhHa1 (approved):
LGTM

PR #384, opened 2026-08-20
perf(bagel): pack FlowGRPO replay steps

zzhuoxin1508:
## Summary

- Add `pipeline.batch_replay_steps` as an opt-in switch for packing Bagel FlowGRPO replay steps.
- Pack the per-step runtime-resolved SDE targets into isolated Navit varlen sequences computed in a single forward pass (block-diagonal attention via `cu_seqlens`), replacing the per-step log-prob recompute loop.
- Repeat the generation KV context for T2I/IT2I and vectorize the Flow SDE transition.
- Preserve per-sequence projection/MLP reduction boundaries so gradients stay equivalent to the serial path under BF16.

Scope: Bagel FlowGRPO T2I/IT2I only; no CFG support (requires `cfg_text_scale = cfg_img_scale = 1`); rollout/replay anchor policies unchanged; no unified-model, shared-algorithm, or benchmark changes.

## Related Issue

N/A

## Test Plan

- Correctness (serial vs packed parity): full-LoRA relative gradient error ~0.4%.
- Performance (300-rollout vLLM-Omni IT2I comparison): ~10.8% faster training (48.05s vs 53.24s) and ~5.7% faster end-to-end (83.25s vs 87.97s) per rollout.
- End-to-end: completed all 300 rollouts; final reward 1.523; final-50 mean 1.393 vs main serial 1.157.

## Compatibility / Risk

- New `pipeline.batch_replay_steps` flag defaults to `false`; existing recipes and checkpoints are unaffected.
- Only valid with no-CFG sampling; the packed forward scales activation/KV memory with the replay window (`repeats`), so peak memory grows for larger windows (e.g. T2I `num_inference_steps`).
- BF16 packed vs per-step is not bitwise-identical, but the `ratio == 1` invariant at update-1 still holds (pi_old/pi_new share the packed path).

## Reviewer Notes

- AI-assisted; checked for overlapping open replay/perf PRs before proposing.

## Checklist

- [x] I reviewed the changed code and removed unrelated/generated artifacts.
- [x] I updated tests, docs, and configs where needed, or explained why not.

PR #444, opened 2026-09-12
fix(diffusion): keep replay transition means at SDE-native fp32

Yuanjia-Lee:
## Summary

FlowDPPO's KL mask and its log-prob score different distributions for the same
transition. The SDE kernel computes the transition mean μ_θ in fp32 and scores
log-prob against that mean, but `stage.replay()` narrowed the returned
`prev_sample_means` to `trajectory_dtype` (typically bf16), and the mask
`(μ_θ − μ_old)²/(2σ_t²)` compares those narrowed means against
`kl_mask_threshold: 1e-5` (`sd3_flowdppo.yaml`). One bf16 rounding of a
|μ|≈1.6 value has std 2.3e-3, the same size as the μ-delta the threshold is
sized for (RMS ≈ 2e-3 at the recipe's σ window): small legitimate updates get
masked out, sub-ULP deltas read as KL = 0, and the β>0 reference penalty on the
same means reads 0.

The fix removes that cast from every diffusion replay implementation: SD3,
Qwen-Image/Edit-Plus, WAN 2.1/2.2, HunyuanVideo 1.0/1.5, HunyuanImage 3,
Z-Image, FLUX.2-klein, LTX2, Boogu-Image, Bagel, and SenseNova U1, plus the
shared `BatchedStepReplayMixin`. It also preserves the rollout-owned old-mean
anchors in Bagel and SenseNova U1. MiniMax-H3 already returned un-narrowed
means. `log_probs` keeps its `logprob_dtype` cast.

Measured through the real replay→loss→optimizer code at the SD3 recipe's
hyperparameters: replay's log-probs recomputed from the returned means match
exactly after the fix (max|Δ| = 0.0) and deviate by ~2e-5 with the cast. A
small weight perturbation flipped the mask on 5/24 sample-steps (native KL
~1e-6, narrowed up to 1.25e-5) and reversed 61/256 gradient components (cosine
0.76); with no flips the two arms' gradients are identical (cosine 1.0), so the
reversal is the mask, not noise.

## Related Issue

N/A

## Test Plan

- `SKIP=no-commit-to-branch pre-commit run --files <all 17 changed files> --show-diff-on-failure`
  → all hooks pass locally; the pre-push hook also passes on the follow-up commit.
- Focused kernel check with bf16 `sample`/`noise_pred` verifies both the returned
  transition and `prev_sample_mean` remain `torch.float32`.
- Source sweep verifies no diffusion replay or rollout-owned old-mean anchor
  narrows `prev_sample_means`/`sde_means` to `trajectory_dtype`.
- Original SD3 harness (uncommitted, no `tests/` tree by policy #99/#267): full
  `SD3DiffusionStage.diffuse/replay` (serial and batched, bit-identical) and
  the `FlowDPPO` loss + AdamW (lr 3e-4, wd 0) at eta 0.7 / threshold 1e-5 /
  10-step shift-3 schedule / B=8, with a small deterministic transformer
  standing in for the SD3.5 checkpoint; CPU and CUDA bf16-autocast on one 4090.
  Numbers are in the Summary.

## Compatibility / Risk

- Means now stay at the SDE kernel's native fp32 across all replay stages.
  Consumers already align old/reference means to `new_means.dtype` or compute
  the KL in fp32; no dtype assertions exist. Numerics of mean-based KL paths
  change by design; `log_probs` are untouched.
- A retained mean anchor takes twice the memory when `trajectory_precision` is
  bf16. For the SD3 recipe's `[8, 3, 16, 64, 64]` micro-batch, the frozen CPU
  anchor is 6 MiB instead of 3 MiB. Bagel and SenseNova U1 also retain their
  rollout-owned old-mean anchors in fp32 rather than bf16.

## Reviewer Notes

- The implementation now matches the repo-wide contract documented on
  `ReplayResult.prev_sample_means` and in `unirl/sde/README.md`; serial and
  batched replay no longer differ by dtype.
- Duplicate check: no open PR or issue covers mean precision. #394 only adds
  per-step batching and carried the same cast in its new code; #432, #384, and
  #428 do not touch this boundary.

## Checklist

  (Contract recorded in `unirl/sde/README.md` Gotchas and on
  `ReplayResult.prev_sample_means`.)

### DIFF
--- a/unirl/models/qwen_image/diffusion.py
+++ b/unirl/models/qwen_image/diffusion.py
@@ -483,7 +483,7 @@ class QwenImageDiffusionStage(BatchedStepReplayMixin, DiffusionStage[QwenImageCo
                     prev_sample_means.append(prev_mean)
 
         log_probs_t = torch.stack(log_probs, dim=1).to(dtype=self.logprob_dtype)
-        means_t = torch.stack(prev_sample_means, dim=1).to(dtype=self.trajectory_dtype) if prev_sample_means else None
+        means_t = torch.stack(prev_sample_means, dim=1) if prev_sample_means else None
         return ReplayResult(log_probs=log_probs_t, prev_sample_means=means_t)
 
     def _batched_step_kwargs(self, segment: LatentSegment, params: DiffusionSamplingParams) -> Dict[str, Any]:
--- a/unirl/models/sd3/diffusion.py
+++ b/unirl/models/sd3/diffusion.py
@@ -397,7 +397,7 @@ class SD3DiffusionStage(BatchedStepReplayMixin, DiffusionStage[SD3Conditions]):
                     prev_sample_means.append(prev_mean)
 
         log_probs_t = torch.stack(log_probs, dim=1).to(dtype=self.logprob_dtype)
-        means_t = torch.stack(prev_sample_means, dim=1).to(dtype=self.trajectory_dtype) if prev_sample_means else None
+        means_t = torch.stack(prev_sample_means, dim=1) if prev_sample_means else None
         return ReplayResult(log_probs=log_probs_t, prev_sample_means=means_t)
 
     @staticmethod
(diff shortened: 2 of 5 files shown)


CjhHa1 (approved):
LGTM

PR #459, opened 2026-09-14
fix(experimental): supply the refl videoalign recipe's missing dataset

YSunLIN:
## Summary

`wan21_t2v_videoalign_refl` required `DATA_PATH` to point at an out-of-tree prompt file with
no default, so the one verified video recipe in `experimental/refl` (PASS smoke plus an
835-rollout contributor run, per the package README's verification table) was the one that
could not be launched from a clean checkout. This fills that gap: it commits the prompt set
the experiment runs on and makes it the recipe default.

- `datasets/refl_videoalign/` — 677 internally constructed T2V prompts, split
  `train.txt` 653 / `test.txt` 24, plus a README covering the prompt set, format and splits.
- `experimental/refl/examples/wan21_t2v_videoalign_refl.yaml` — `data_path` /
  `eval_data_path` now default to those files. `DATA_PATH` / `EVAL_DATA_PATH` still override.
  Seeded training shuffle prevents `drop_last=True` from permanently excluding the same five
  prompts from the 653-row training split.
- `.gitignore` — allowlist entry for `datasets/refl_videoalign/*.txt`. Committing data under
  `datasets/` is deliberately gated on an explicit line there.
- `datasets/README.md`, `experimental/refl/README.md` — contents row and launch snippet.

`test.txt` is sized at 24 = 3 full rollouts at the recipe's `batch_size: 8`, kept small
because an 81-frame 480x832 generation per prompt makes held-out scoring expensive. Note
that `experimental/refl` does not run periodic eval today (its README's "Deliberately not
ported" section), so `test.txt` is a held-out set waiting on an eval loop rather than one
the recipe currently scores.

## Related Issue

N/A

## Test Plan

- `SKIP=no-commit-to-branch pre-commit run --all-files --show-diff-on-failure` — all 19 hooks
  Passed (`no-commit-to-branch` skipped), including `check-recipe-targets`
  (2715 `_target_` paths resolve), `check-experimental-boundaries`, `check-docstring-lines`
  and `check-core-dependencies`.
- Focused pre-commit and pre-push checks after review fixes — all hooks passed.
- Config default resolution, no env vars set:
  ```
  data_path      = datasets/refl_videoalign/train.txt
  eval_data_path = datasets/refl_videoalign/test.txt
  ```
  And with `DATA_PATH=/custom/prompts.txt` exported, both keys resolve to
  `/custom/prompts.txt` — the override path is intact.
- Seeded-shuffle coverage check: all 653 training prompts were sampled within two DataLoader
  epochs despite `drop_last=True`.
- `git check-ignore -v datasets/refl_videoalign/*.txt` → matched by
  `.gitignore:97:!datasets/refl_videoalign/*.txt`, i.e. both files are tracked rather than
  silently ignored.
- Committed split integrity: 653 train + 24 test = 677 unique prompts, with no train/test
  overlap, blank lines, CRLF, or leading/trailing whitespace.
- Format contract confirmed against the reader: `unirl/data/datasets.py:230-236` handles
  `.txt` as one stripped prompt per line, skipping blanks.
- Not run: a training smoke test. No GPU in this environment; this PR changes only data and
  data-source configuration, and the recipe's grad path is unchanged.

## Compatibility / Risk

`data_path` moves from a required `${oc.env:DATA_PATH}` to a defaulted one. Existing launches
that export `DATA_PATH` still use that file; launches that did not set it previously failed
at config resolution and will now train on the committed set.

Training prompt order changes from fixed to seeded shuffle (`seed: 42`) so the same five
examples are not discarded at every DataLoader epoch boundary. The order remains deterministic.

Adds 68 KB of text to the repo. No code, checkpoint, or API changes.

## Reviewer Notes

- Scope is `experimental/refl` only. `examples/diffusion/wan21_t2v_videoalign_dancegrpo.yaml`
  is also a VideoAlign recipe but was left alone deliberately, as it is not the ReFL path.
- The 24-prompt `test.txt` is a judgment call — worth a look if you would rather size the
  held-out split to the eval loop that eventually lands.
- The prompt set was internally constructed by the Tencent Hunyuan team; the committed files
  define the dataset split.
- Initial preparation used Claude Code; the verification commands above were run and their
  results checked.
- Duplicate-work check found no other PR adding a ReFL VideoAlign prompt set.

## Checklist


CjhHa1 (approved):
LGTM

PR #462, opened 2026-09-14
fix(reward): grade math-verify in a bounded child process

celve:
## Summary

`MathVerifyRewardScorer._compute_model_rewards` grades inline on the calling thread with both of math-verify's own guards disabled:

```python
verify(
    parse("\\boxed{" + gt + "}", parsing_timeout=None),
    parse(text or "", parsing_timeout=None),
    timeout_seconds=None,
)
```

So a pathological expression blocks the reward call with **no deadline and no way to interrupt it**. Those `None`s are not an oversight: math-verify's timeouts are `signal.alarm`-based and require the main thread. A synchronous Ray actor may run reward code on its main thread, but threaded actors (`worker_max_concurrency > 1`) do not, so inline grading behaves differently across worker configurations. A thread cannot be killed and a runaway grade may hold the GIL, so the scorer needs a process boundary.

Downstream this surfaces as a reward call that never returns. Where reward is dispatched across ranks and gathered with an untimed `ray.get`, one non-returning shard holds the whole gather indefinitely, with no exception and every rank apparently idle.

**The change:** grade the whole batch in one child process over a `Pipe`, with per-operation timeouts (`UNIRL_MATHVERIFY_TIMEOUT_S`, default 10s) and a bounded parent deadline (`3 * timeout * jobs + 60s`). Inside the child the main thread is the child's own, so math-verify's signal-based timeouts work consistently. The parent budget covers two parse windows and one ordinary verification window per job while remaining a hard aggregate cap. If the child is OOM-killed, exits early, or misses that cap, the scorer raises so reward fail-fast rejects the step instead of accepting a successful batch of zeros.

**One child per reward call, not per grade.** The forkserver preloads both `math_verify` and the module that defines the process target, so later children inherit those imports instead of paying the previously measured ~1.2 s target-module import on every call. The first call still starts and preloads the forkserver; batching pays process and IPC overhead once per reward call rather than once per grade.

Three details are load-bearing, and are recorded in `unirl/reward/README.md` under *Gotchas* (the docstring hook holds docstrings to one line):

- **`forkserver`, not `fork`.** `fork` inherits Ray's signal handlers, `atexit` and CUDA state, plus copies of locks held by other threads. It also runs `logging`'s registered at-fork handler in the parent (`_prepareFork` → `_lock.acquire()`, unbounded), which can block before the fork.
- **`wait()` on the child's sentinel as well as the pipe.** `Connection.poll()` waits for *data*; a child killed at once sends none, so the call would wait out the full deadline before reporting failure.
- **`proc.start()` inside the `try`.** `forkserver` forks the child *before* the parent writes the job payload to it, so a child dying in that window breaks the pipe mid-write and `start()` raises `BrokenPipeError`; keeping it inside the `try` guarantees bounded cleanup while preserving that failure.

**The reason to grade out-of-process with `kill()` rather than a `Pool`** is that `Pool.terminate()` is not bounded, which we confirmed with a stack from a reproduced 32-GPU hang. `_terminate_pool` sends each worker `SIGTERM` and then joins it with **no timeout** (CPython 3.12 `pool.py:732`):

```python
        if pool and hasattr(pool[0], 'terminate'):
            for p in pool:
                if p.is_alive():
                    p.join()        # unbounded
```

A pool worker created with `fork` can run inherited Ray signal and exit handlers against inconsistent post-fork state and fail to exit. The captured stack does not identify the exact child-side stall; it confirms the parent remained blocked in the pool's unbounded join. Captured stack, 1 of 384 rank dumps at the stall:

```
    join (multiprocessing/process.py:149)
    _terminate_pool (multiprocessing/pool.py:732)
    terminate (multiprocessing/pool.py:657)
    _reset_pool (mathverify.py:60)  <- the code this PR removes
    _verify_with_deadline (mathverify.py:78)
    score_and_attach (reward/service.py:114)
```

`proc.kill()` sends `SIGKILL`, which cannot be caught, blocked or ignored and needs no bytecode boundary, followed by a bounded `join(timeout=1.0)`. That is why this change uses it rather than a pool.

**Infrastructure failures are not rewards.** A wrong or unparsable answer remains a valid 0.0, but a missing dependency, dead child, malformed IPC result, or expired outer deadline raises. This prevents the previously observed case where a worker missing `math_verify` silently reported `eval_acc` 0.0000 instead of failing the reward step.

## Related Issue

N/A — tracked internally.

## Test Plan

**1. Focused, no GPUs.** Local-only guarded checks (not included in the PR diff) covered real grading, process recovery, IPC failures, cleanup, and timeout validation:

These checks must run from a real file with an `if __name__ == "__main__":` guard. `forkserver` re-imports `__main__`; an unguarded script, or a heredoc where `__main__` is `<stdin>`, fails child startup. Production uses Ray's `default_worker.py` and is unaffected.

```
PASS  right=True, wrong=False
PASS  SIGKILL propagated RuntimeError (exit code -9)
PASS  a healthy call still succeeds after the killed child
PASS  a missing child dependency propagated instead of returning a verdict
PASS  malformed IPC and invalid timeout values were rejected
```

**2. End to end**, 8×H20 (TaiJi 1×8, gz), `Qwen3-4B-Base`, 60-prompt AIME set at 16 samples/prompt, `num_devices=8 batch_size=8 eval_interval=10`:

```
python -m unirl.train_ar --config-name=ar/qwen3_grpo_4b_base_dapo_sglang \
  num_devices=8 batch_size=8 num_rollouts=1 eval_interval=10 \
  ++data_source.args.algorithm.prompts_per_rollout=8 logging.report_to_wandb=false

EVAL rollout 0  eval_acc(avg@16 over 60 prompts, 8 batches of <=8)=0.0219
```

8/8 eval batches, zero grader warnings, **48.4 s per batch against a 48 s pre-change baseline**. `eval_acc=0.0219` sits inside the 0.0156–0.0240 spread of repeat runs at temperature 1.0.

**3. Reproduced the original hang and confirmed the mechanism**, 4x8 H20 (32 GPUs), unpatched pre-change code: hung in eval batch 7, stuck rank captured with the stack above. A control run with the grader failing closed on every sample completed 8/8 batches without hanging, which isolates the cause to a child that will not exit rather than one that errors.

**4. Lint.** `lint/check_docstring_lines.py`, `check_core_dependencies.py`, `check_experimental_boundaries.py` all clean on this branch.

## Compatibility / Risk

- One new config surface: `UNIRL_MATHVERIFY_TIMEOUT_S` (env var, default 10s). No YAML, checkpoint, or data-format change. `MathVerifySpec` and the `LocalRewardBackend` contract are untouched.
- **New behaviour worth naming:** per-answer parse/verify timeouts return `False`, while process-level failures (early exit, OOM/SIGKILL, malformed IPC, or the outer deadline) raise and fail the reward step.
- The first call starts and preloads the forkserver. Later children inherit `math_verify` and the target module; their remaining overhead is process creation and IPC rather than repeated module import.

## Reviewer Notes

- One file of code plus a README *Gotchas* entry. Start with `_grade_in_child` — the three load-bearing details above are all in those ~20 lines, and each one is a bug if it is written the obvious way.
- **AI assistance:** written with Claude Code. Every claim in the Test Plan is a command that was run and whose output is quoted; the two defects listed were found by those tests rather than by reading.
- **Duplicate-work check:** no open PR touches `unirl/reward/local/mathverify.py` (`gh pr list --search "mathverify"` empty). Upstream `main` does not carry the fork-side hardening commits this replaces, so this is the design ported onto `main` rather than a cherry-pick.
- The implementation remains two focused functions. Earlier scaffolding (module globals, a `threading.Lock`, and a respawn path) existed to reuse one child across grades; batching the call made it unnecessary.

## Checklist


CjhHa1 (approved):
LGTM

PR #467, opened 2026-09-15
feat(fsdp): support copy-engine all-gather

CjhHa1:
## Summary

Opt-in `training.fsdp.copy_engine_all_gather` so native FSDP2 all-gather can run on NCCL copy engines instead of SM kernels.

When the flag is on, `FSDPBackend` creates WORLD with zero-CTA NCCL options (`cpu:gloo,cuda:nccl`, bound `device_id`) and `fsdp_wrap` calls `FSDPModule.set_symm_mem_for_comm("NCCL")` on every fully-sharded module. DeviceMesh shard groups inherit the policy from that WORLD. The CTA policy is communicator-wide, not all-gather-only.

Requires PyTorch >= 2.13, NCCL >= 2.28, and a node-local shard group (`full` on one node, or `hybrid` with `hsdp_shard_size: devices_per_node`). `no_shard` is rejected. VeOmni is not wired (it wraps through `parallelize_model_fsdp2`); left as a TODO on the reject path.

## Related Issue

N/A

## Test Plan

- [x] `pre-commit` on the changed files
- [x] Taiji 8×H20 synthetic FSDP2 A/B: numerical match vs default all-gather
- [x] Taiji MiniMax-H3 1×8 trainside smoke with the flag on
- [x] Taiji 8×8 HSDP (`world_size=64`, `hsdp_shard_size=8`) production-path smoke
- No unit tests or recipes in this PR: the path needs a live NCCL group and was checked on cluster instead. On H20 these jobs were throughput-neutral; the flag stays default-off.

## Compatibility / Risk

- Default `false`; existing training is unchanged.
- Copy-engine WORLD init must happen before any other `init_process_group`. `NCCL_CTA_POLICY` must stay unset or `2`.
- Hybrid mode logs an expected Gloo split "Falling back to default options" warning per rank.
- VeOmni and `no_shard` raise if the flag is set.

## Reviewer Notes

- The interesting bits are WORLD init in `unirl/train/backend/fsdp/backend.py` and `set_symm_mem_for_comm` in `fsdp_wrap`. Do not reintroduce DeviceMesh `backend_override` / `ncclCommSplit` with different child options (segfault on Torch 2.13 / NCCL 2.29.7).
- `ensure_dist_initialized` is unchanged; the flag-off path still uses it.

## Checklist


CjhHa1:
Cluster validation update (8x H20, PyTorch 2.13.0+cu130, NCCL 2.29.7):

- The synthetic FSDP2 A/B smoke passes numerically in both modes: identical final loss, replica checksum spread `0.0`, and essentially identical peak memory.
- Two fresh 50-step runs gave baseline median step times of 130.27/130.09 ms and copy-engine times of 132.74/133.22 ms (about 2.15% slower here). This microbenchmark has no useful compute/communication overlap, so it validates correctness but does not demonstrate a throughput gain.
- A prior 64-GPU HSDP smoke also passed correctness in both modes (`world_size=64`, `shard_size=8`, replica checksum spread `0.0`).
- The copy-engine runs consistently emit `ncclCommWindowDeregister failed: invalid argument` / `corrupted comm object detected` during symmetric-memory teardown on this NCCL build. The process still exits successfully, but this needs resolution or documentation before marking the draft ready.
- MiniMax-H3 with the locally staged checkpoint successfully reached `FSDPBackend` initialization with `copy_engine_all_gather=true`; the run then failed while constructing the unrelated PickScore reward because its offline CLIP processor/model assets were incomplete. No H3 training-step throughput number yet.

Keeping this PR in draft while the teardown warning and real-workload performance validation are unresolved.

CjhHa1:
Follow-up MiniMax-H3 validation (matched 1x8 H20 runs):

- Located the shared PickScore directories, but they contain configs/tokenizer metadata only; neither the mounted copies nor the pod-local HF cache has `model.safetensors`/`pytorch_model.bin`.
- To isolate the FSDP path, reran with a dependency-free test reward and a node-local MiniMax-H3 checkpoint. Both modes completed one full trainside rollout/replay cycle with 2 denoise steps.
- Copy-engine: 164.425 s from rollout-engine readiness to `rollout 1/1`; 229 s total process time.
- Baseline: 161.844 s for the same interval; 224 s total.
- On this single-run configuration, copy-engine was about 1.6% slower in the measured rollout interval (about 2.2% total), so there is no demonstrated speedup. MiniMax-H3 uses `root_wrap=false`, which also prevents enabling the existing FSDP forward-prefetch overlap path; that likely limits the opportunity for freeing SMs to help.

The functional H3 path is validated, but performance evidence currently does not justify enabling this by default.

CjhHa1:
Paired rank-0 PyTorch traces explain the result:

| metric (100 FSDP all-gathers) | baseline | copy-engine flag | delta |
|---|---:|---:|---:|
| traced rollout | 95.861 s | 97.550 s | +1.76% |
| NCCL device work | 461.182 ms | 411.543 ms | -10.76% |
| FSDP `all_gather` host annotation | 1.486 s | 8.733 s | +7.247 s |

Key observations:

- Baseline uses `ncclDevKernel_AllGather_RING_LL` with a 24-CTA grid.
- The enabled path uses `ncclSymkDevKernel_AllGather_STMC` with a 2-CTA grid. It substantially reduces SM/CTA usage, but it is a symmetric-memory kernel fallback, not a zero-CTA Copy Engine transfer. NCCL INFO logs contain no `AllGather [Copy Engine]` selection.
- Symmetric-memory first-use setup is expensive in this one-rollout test: `cuMemExportToShareableHandle` totals 1.907 s and `cuMemImportFromShareableHandle` 1.158 s, plus rendezvous/mapping overhead. This dominates the ~49.6 ms saved by the NCCL device kernels.
- Even baseline NCCL all-gather device time is only about 0.48% of the 95.9 s traced rollout. H3 compute dominates, so the maximum end-to-end gain is small.
- H3 has `root_wrap=false`, so the existing `forward_prefetch` overlap path cannot be enabled. Reducing communication CTA usage therefore has little compute-overlap value here.

Conclusion: the flag currently provides low-CTA symmetric all-gather on this H20/NCCL 2.29.7 stack, but does not prove actual CE selection. The API/config should avoid guaranteeing “copy engine” without runtime observability, and performance should be evaluated after warmup over multiple rollouts.

CjhHa1:
Root cause and fix: the previous implementation only set `NCCL_CTA_POLICY=2`. On this PyTorch 2.13/NCCL 2.29.7 stack, the communicator still selected `ncclSymkDevKernel_AllGather_STMC` (2 CTAs).

Using explicit communicator options:

```python
opts = dist.ProcessGroupNCCL.Options()
opts.config.cta_policy = dist.ProcessGroupNCCL.NCCL_CTA_POLICY_ZERO
dist.init_process_group(..., pg_options=opts)
```

changed the H20 trace to `cudaMemcpyBatchAsync` plus a 117,440,512-byte P2P memcpy, with no all-gather CUDA kernel. This confirms true zero-CTA Copy Engine execution.

Commit `deeecc64` now creates and forwards explicit PG options from `FSDPBackend`; the smoke test uses the same path. Targeted tests: 5 passed, lint/format passed.

CjhHa1:
Post-fix true-CE performance A/B (1x8 H20, MiniMax-H3 trainside, 3 rollouts, identical inputs/config, first rollout treated as warmup):

| mode | steady rollout 2 | steady rollout 3 | steady mean | total process |
|---|---:|---:|---:|---:|
| baseline | 157.061 s | 147.381 s | 152.221 s | 529 s |
| explicit zero-CTA CE | 158.040 s | 147.743 s | 152.892 s | 531 s |

True CE is +0.44% in the steady mean and +0.38% total, which is within run-to-run noise. This confirms that the ~3 s first-registration cost is not material once amortized; the lack of speedup is instead because H3 all-gather is a very small fraction of the compute-heavy rollout and this `root_wrap=false` recipe has no forward-prefetch overlap to benefit from freed SMs.

CjhHa1:
Additional real-model benchmark: SD3.5-Medium on 1x8 H20 with `root_wrap=true`, `reshard_after_forward=true`, and `forward_prefetch=true`; 50 rollouts per arm, first completion excluded from interval statistics.

| mode | steady mean (49) | median | p10 / p90 | total process |
|---|---:|---:|---:|---:|
| baseline | 0.9002 s | 0.8930 s | 0.869 / 0.943 s | 100 s |
| explicit zero-CTA CE | 0.8882 s | 0.8770 s | 0.865 / 0.922 s | 102 s |

Steady-state improvement: **1.32% mean / 1.79% median**. CE startup/registration adds about 2 s to this short process, but steady-state benefits once amortized. This is the expected use case: root-wrapped FSDP with forward-prefetch overlap, unlike MiniMax-H3.

CjhHa1:
Final review/validation update:

- Removed all tests and validation recipes from the PR diff; local copies remain untracked and on `backup/fsdp-copyengine-with-tests`.
- Independent Opus/Grok reviews found and drove fixes for root-only symmetric-memory enablement, HSDP option scoping, env leakage, unsupported backends, timeout mismatch, and Torch 2.11 capability handling.
- Runtime validation found that passing different child options through `ProcessGroupNCCL::split` segfaults on Torch 2.13/NCCL 2.29.7. Final implementation leaves WORLD unchanged and uses `DeviceMesh` -> `new_group(pg_options=...)` for a dedicated `dp_shard` communicator.
- Exact final full-FSDP production path completed 6 SD3.5-Medium rollouts. Rank-0 trace contains 97 `cudaMemcpyBatchAsync` all-gathers / 683 D2D copies and **zero** `ncclDevKernel_AllGather` or `ncclSymkDevKernel_AllGather` kernels, confirming true CE execution across all nested FSDP blocks.
- The earlier SD3 +1.3% number is withdrawn: that run preceded the root-only fix and was noise. The corrected path is functionally validated; no performance claim is made yet.
- Exact 8x8 HSDP hardware validation is still pending because the 64-GPU allocation is currently occupied by another workload; static review found the `new_group` hybrid path correct.

CjhHa1:
Ready-for-review gate completed after rebase:

- Rebasing onto current `upstream/main` completed; head is `da91832c`.
- Exact rebased production implementation passed an 8-node x 8-GPU HSDP smoke (`world_size=64`, `hsdp_shard_size=8`) with finite loss and replica checksum spread `0.0`.
- Rank-0 trace shows 8 `cudaMemcpyBatchAsync` / P2P Copy Engine all-gathers and zero `ncclDevKernel_AllGather` or `ncclSymkDevKernel_AllGather` kernels.
- `dp_replicate` all-reduce and `dp_shard` reduce-scatter correctly remain ordinary NCCL kernels; zero-CTA is scoped only to all-gather on the shard communicator.
- Synthetic-process teardown still emits the known NCCL symmetric-window deregistration warning, but the run exits successfully; this does not occur during the validated long-lived trainer path.

The full and 8x8 HSDP production paths are now both hardware-validated.

nussejzz (approved):

Issue #473, opened 2026-09-16
[Tracking] Audit and retire vLLM/vLLM-Omni patches for 0.28/0.29


leviking98z-rgb:
## Context

UniRL currently pins:

- `vllm==0.27.0`
- `vllm-omni==0.27.0rc1`

and carries a quarantined set of monkey patches under
`unirl/rollout/engine/vllm_omni/patches/`.

This issue tracks which patches are still necessary when upgrading the runtime,
which ones can be retired, and which implementations should be tightened before
an upgrade.

Related historical upgrade issue: #129.

## Audit baselines

- UniRL `origin/main`: `7ec3683d581fe0e5e9543f6103024e94a20edc9a`
- Runtime-tested stable pair:
  - `vllm==0.28.0`
  - `vllm-omni==0.28.0`
  - `torch==2.13.0+cu130`
  - `transformers==5.6.0`
  - NVIDIA H20
- Source-audited development pair:
  - vLLM `b3b13c12923f0b68f36ee515ebbc3efc7679eb4f` (roughly 0.29-era main)
  - vLLM-Omni `21d86ec9239d5aba08ea5fcac7e20a8f0679b0c9`

`vllm 0.29.0 + vllm-omni 0.28.0` should not be treated as a supported pair.
The stable paired upgrade target is 0.28/0.28; current vLLM-Omni main is the
appropriate source baseline for vLLM 0.29.

## Summary

| Patch / behavior | vLLM 0.28 + Omni 0.28 | vLLM 0.29 + Omni main |
| --- | --- | --- |
| Explicit `register_capture_flush()` call inside the bundle | Remove | Remove |
| `vllm_omni.general_plugins` capture-flush entry point | Keep: UniRL feature | Keep: UniRL feature |
| `compat_hi3_lora` tuple unwrapping | Keep | Remove / version-gate |
| DiT and AR in-memory tensor LoRA loaders | Keep | Keep |
| HI3 DiT namespace alias and GQA-interleaved QKV-B conversion | Keep | Keep |
| HI3 AR fused-QKV tensor conversion | Keep | Keep |
| fp32 LoRA wrapping skip | Keep for HI3 LoRA | Keep for HI3 LoRA |
| HV1.5 packed-module mapping | Keep | Keep |
| HV1.5 token-refiner `nn.Linear` LoRA support | Keep | Keep |
| `lora_request` passthrough through `Omni.generate()` | Keep | Keep |
[...]

PR #413, opened 2026-09-02
feat(rollout): migrate vLLM-Omni to stable 0.28.0

celve:
## Summary

Migrates UniRL from `vllm-omni==0.27.0rc1` to stable `0.28.0` (`vllm==0.28.0`, `transformers==5.12.1`, `torch==2.13.0+cu130`). This is a breaking runtime migration, not only a version pin.

- Replaces all legacy `stage_configs_path` / `stage_args` assets with stable `deploy_config` YAMLs and registered `PipelineConfig` topologies.
- Adapts the flat `OmniRequestOutput.outputs` contract and routes AR sleep/wake through EngineCore while retaining diffusion worker task ACKs.
- Adapts the direct vLLM rollout engine added on latest `main` from its explicit vLLM 0.27 runtime contract to 0.28; native IPC, sleep/wake, collective RPC, and cache-reset APIs remain signature-compatible.
- Registers UniRL patches through the stable general-plugin entry point so spawned stages receive them.
- Fixes lane validation before MoE workspace indexing, preventing an invalid lane from aliasing another ubatch's workspace.
- Keeps only compatibility required by real model runs: HI3 expert mapping, tensor LoRA loading, worker parameter discovery, and Qwen3-Omni thinker LoRA support.

Dependency overrides remain coherent across mutually exclusive engines: the vLLM extra resolves `diffusers==0.40.0` and its `kernels==0.14.1`, while the sglang extra carries `kernels>=0.12,<0.13`; installation docs no longer enable prereleases globally.

## Risk closure

- **Removed configuration API:** all adapter-selected topologies resolve through stable `StageConfigFactory`; no production reference to `stage_configs_path`, `stage_yaml`, or `stage_args` remains.
- **HI3 AR mode split:** comprehension adapters retain `engine_output_type=text`; terminal think/recaption uses `latent`, preserving the legacy generation sampler while still returning text.
- **Output envelope change:** AR token/text extraction reads the stable flat completion fields.
- **Lifecycle split:** AR and diffusion stages use their owning control plane, and nested unsupported/error/ACK results fail closed.
- **Workspace corruption:** invalid lanes delegate to vLLM before any flattened workspace index is touched; multi-ubatch/multi-lane H20 smoke covers slot isolation.
- **LoRA regressions:** SD3 tensor push/checksum/remove/re-push and HI3/Qwen model-specific paths were exercised on GPU.

## Test Plan

Exact stable environment on one 8×H20 node:

```
[...]

PR #474, opened 2026-09-16
fix(vllm-omni): retire per-request AR seed patch

leviking98z-rgb:
## Summary

Remove the vLLM-Omni per-request AR seed monkey patch and its patch-table entry.

The patch replaced every unset `SamplingParams.seed` with `os.urandom`, based on
the assumption that reusing one `SamplingParams` object collapses a GRPO group to
identical tokens. vLLM 0.27 and 0.28 instead keep `seed=None` on the ordinary
random-sampling path, where the GPU RNG advances across batch rows. The override
is therefore unnecessary for UniRL's current single-replica recipes and makes
rollouts impossible to reproduce from the configured engine seed.

## Related Issue

Refs #473.

## Test Plan

- `SKIP=no-commit-to-branch pre-commit run --all-files --show-diff-on-failure` — passed.
- `python -m compileall -q unirl/rollout/engine/vllm_omni/patches` — passed.
- One-off sampler probe with `vllm==0.28.0`, `vllm-omni==0.28.0`,
  `torch==2.13.0+cu130` on one NVIDIA H20: 64 `seed=None` rows produced 55
  distinct tokens, and resetting the Torch/CUDA seed reproduced the complete
  batch. No model checkpoint or dataset is involved in this sampler-level check.
- Source-checked vLLM `v0.27.0`: `seed=None` likewise avoids a per-request
  generator and uses the advancing batched GPU RNG.

## Compatibility / Risk

Current UniRL vLLM-Omni recipes use one engine replica, so request diversity is
preserved by vLLM's advancing RNG. Independently initialized multi-replica
engines can still begin with correlated RNG streams; if that topology is added,
it should use deterministic seeds derived from request identity rather than host
entropy from `os.urandom`.

No config, checkpoint, data-format, or API migration is required.

## Reviewer Notes

Open PR #413 removes the separate capture-flush direct registration as part of
the vLLM-Omni 0.28 migration, but retains this seed patch. No open PR was found
that removes or replaces the seed override.

AI assistance was used. The full two-file diff and all reported validation
output were reviewed before publication.

## Checklist


CjhHa1 (approved):
LGTM

PR #477, opened 2026-09-16
fix(reward): make prompt semantics explicit for composed rollouts

seshakolli:
## Summary

`RewardService` previously collapsed all text conditioning into one implicit nearest-ancestor prompt. In composed AR→image rollouts, that made intent-based judges such as WISE score against the generated rewrite while root metadata still described the user's original prompt.

This refactors `RewardRequest` to carry both meanings explicitly:

- `original_prompt` — the root user prompt, aligned to final outputs by sample lineage
- `generation_prompt` — the nearest text that actually conditioned the generated frontier
- `conditioning` — non-text source media for edit/V2V rewards
- `generated` — the frontier output being scored

Prompt-aware backends declare `prompt_source: original|generation`. The default is `generation`, preserving existing scorer behavior; the WISE PE recipe selects `original`.

### Changes

- Replace the ambiguous `RewardRequest.primitives['text']` / `prompts` contract with explicit prompt fields
- Align original prompts by sample IDs, including non-uniform fan-out
- Preserve source images/videos in typed `conditioning`
- Add prompt-source capability to local and remote reward specs/backends
- Keep the standalone RewardService HTTP history protocol unchanged
- Remove unused request fields and migrate all built-in local/composite call sites
- Configure WISE to score images against the original user prompt

## Related Issue

N/A

## Test Plan

CPU-only verification:

- Local focused prompt/conditioning contract suite — 9 passed
- WISE PE and BAGEL EditReward configs compose with `--cfg job --resolve`
- `ruff check` and `ruff format --check` pass on the reward package
- All changed-file pre-commit hooks pass
- `git diff --check` passes

No GPU training run was performed.

## Compatibility / Risk

The internal `RewardRequest` API changes atomically across all repository call sites. Existing prompt-aware rewards retain generation-prompt scoring by default. WISE intentionally switches to original-prompt scoring. Edit/V2V source media remain available through `conditioning`.

The remote HTTP wire schema and protocol version are unchanged. The main residual risk is runtime model integration; configuration construction and local CPU contract tests pass, but no GPU scorer or training run was performed.

## Reviewer Notes

AI assistance was used during investigation, drafting, implementation, and verification. The final diff was reviewed against the real upstream base and the repository redundancy checklist.

PR #313 touches the same reward-service area; if it lands first, this branch may need a small rebase.


CjhHa1:
@seshakolli Thanks for identifying the composed-rollout reward-conditioning issue. I kept the intent of the original fix, but refactored the implementation on the PR branch so the prompt semantics are explicit rather than selected indirectly at `RewardService` level.

What changed:

- `RewardRequest` now carries `generated`, non-text `conditioning`, `original_prompt`, and `generation_prompt` separately.
- Prompt-aware backends select `prompt_source: original|generation`; the default remains `generation` to preserve existing scorer behavior, while the WISE recipe explicitly selects `original`.
- Request construction now reuses `Sample.turns()`, which aligns ancestors to the frontier by sample-id lineage. This removes the batch-ratio/repeat assumption and handles non-uniform fan-out.
- Edit/V2V source images or videos remain available through typed `conditioning` instead of sharing the old prompt/primitives bucket.
- All local, remote, video, and per-domain rebuild/slice paths were migrated. Unused request fields (`RewardType`, `prompt_ids`, `reward_types`, and `return_components`) were removed.
- Follow-up cleanup removed duplicate prompt state/checks and shortened the README to configuration-facing guidance.

Why this shape:

The original `conditioning_source` switch fixed WISE, but it still collapsed two different meanings into one field and reconstructed root prompts using a uniform fan-out assumption. Keeping both prompts in the request lets each reward declare what it evaluates, while the framework's existing lineage API owns row alignment.

Verification:

- local prompt/conditioning contract suite: 9/9 passed (kept local, not added to the PR)
- changed-file pre-commit checks passed
- full diff passes `git diff --check`
- two independent full-diff reviews found no blocking issue in the PE/WISE path, non-uniform lineage alignment, or backend propagation
- no GPU training/model-inference run was performed

One deliberate compatibility choice: the existing PickScore PE recipes still use the default generation prompt; only WISE switches to the original user prompt. The review also noted two non-blocking API-hardening follow-ups—fail fast when a selected prompt is absent, and define role semantics for system/tool text turns—but neither affects the current composed PE construction, where both prompts are populated and the root text is the user prompt.

CjhHa1 (approved):
LGTM

PR #479, opened 2026-09-17
fix(hi3): support batched trainside AR with dynamic KV cache

### LINKED ISSUE #452: fix(hi3): support batched trainside AR with dynamic KV cache
## System info

- UniRL: `5a534331` (`upstream/main`; reproduced while validating #446)
- Python 3.13
- PyTorch `2.10.0-rc6`, CUDA 12.9
- Transformers `5.6.0`, Diffusers `0.38.0`, PEFT `0.20.0`
- 8x NVIDIA H20, FSDP2 full shard

## Rollout / inference engine

Train-side only

## Domain

`unified_model` / HunyuanImage3 AR

## Recipe / config-name

Based on `unified_model/hi3_vllmomni`, with the AR prefill isolated under its production FSDP/LoRA settings.

## How you ran it

- [x] Modified recipe or my own script

## Reproduction

Build four unequal-length text prompts through `HunyuanImage3TextEmbedStage.embed_for_ar`, combine them with `HunyuanImage3FusedMultimodalCondition.concat`, then initialize and run the first cached AR step:

```python
embedded_rows = [
    embedder.embed_for_ar(Texts(texts=[prompt]), bot_task="think", max_length=512)
    for prompt in prompts
]
fused = HunyuanImage3FusedMultimodalCondition.concat(
    [row["fused"] for row in embedded_rows]
)
real_pos = torch.cat(
    [row["tokenizer_output"].real_pos for row in embedded_rows], dim=0
)
conditions = HunyuanImage3ARConditions(
    fused=fused,
    tokenizer_output=SimpleNamespace(real_pos=real_pos),
)
state = HunyuanImage3ARStep().init_state(bundle, conditions, max_new_tokens=1)
HunyuanImage3ARStep().step(bundle, conditions, state)
```

Observed input:

```text
input_ids.shape = [4, 490]
real_pos = [13, 220, 490, 450]
```

Traceback:

```text
HunyuanImage3ARStep.step
  -> HunyuanImage3ForCausalMM.forward
  -> HunyuanImage3Attention.forward
  -> HunyuanStaticCache.update

File modeling_hunyuan_image_3.py, line 965, in update
    assert len(cache_position) == 1
AssertionError
```

UniRL constructs this cache with `dynamic=True` in `unirl/models/hunyuan_image3/ar.py::_build_kv_cache`. For a 2D cache position, the upstream cache updates each batch row, then explicitly rejects every dynamic batch larger than one.

`B=1, L=490` runs successfully with identical logits and KV across repeated first-prefill forwards.

## Expected behavior

Cached trainside HunyuanImage3 AR should support `B > 1`, including right-padded prompts with different `real_pos` values. Batched logits and KV should match running each sample independently at `B=1` for both prefill and subsequent decode steps.

## Suggested validation / possible fixes

Possible implementation directions:

1. Use `dynamic=False` when `batch_size > 1`, if the full static cache remains correctly masked; the diffusion path already constructs `HunyuanStaticCache(dynamic=False)`.
2. Extend the dynamic 2D path to return a slice ending at the maximum active cache position instead of asserting `len(cache_position) == 1`.

Please validate before choosing either approach:

- `B=2+` with unequal right-padded prompt lengths / `real_pos`;
- batched vs independent `B=1` logits and KV parity;
- at least two decode steps after prefill;
- finished-sequence masking and cache growth;
- both plain and FSDP trainside forwards.

## Impact

The surrounding UniRL AR implementation is batch-shaped, but this cache assertion currently limits the actual cached trainside path to one sample. External vLLM-Omni rollout is unaffected.

PushpakAg:
## Summary

Cached train-side HunyuanImage3 AR only worked at `B=1`. The crash reported in #452 is
upstream's `assert len(cache_position) == 1` in `HunyuanStaticCache.update`, where `len()` on a
2-D cache position is the *batch* dim. A dynamic cache trims the returned KV to one scalar end
position, which cannot represent per-row right-pad, so `_build_kv_cache` now requests
`dynamic=True` only at `B == 1` and lets `B > 1` use the full static cache. That makes the
assertion unreachable (it is guarded by `if self.dynamic`) instead of monkey-patching upstream
remote code — the direction the issue suggested first.

Removing the crash was not sufficient for the behavior the issue asks for (batched logits/KV
matching independent `B=1`). Three further defects were hiding behind it, each invisible at
`B=1` because `real_pos == L` there:

- **Decode mask.** Upstream deletes the attention mask during `gen_text` decode ("Remove
  attention mask to use full attention of 1 x seqlen"), which is only safe for a cache trimmed
  to the valid length. Over a static cache the decode query also attends the right-pad KV
  written at prefill and the never-written tail slots. `_cache_attention_mask` supplies a bool
  mask admitting row `i`'s keys `[0, real_pos_i + step_idx)`, and right-pads the *prefill*
  mask's key axis, since a static cache returns its full `max_cache_len` key axis while the
  prefill mask is only `L` wide.
- **Token placement.** Upstream re-reads the next input token with
  `input_ids.gather(1, position_ids)` where `position_ids` is `real_pos` (the first `<pad>`
  index, advanced once per step). The sampled token must therefore be scattered to
  `real_pos_i + step_idx` (`_place_token`); appending it at the padded tail fed every shorter
  row its own pad token as the next input.
- **`position_ids` padding.** `FusedMultimodalCondition.concat` padded the ragged `L` axis with
  `0`, but `position_ids` *are* the KV-cache write indices (upstream passes them straight
  through as `cache_position`, applied with `index_copy_`). Every pad position of every short
  row therefore wrote its KV onto cache slot 0 — last write wins, silently replacing that row's
  first real token. Measured on the repro shape: slot 0 off by ~4.0 while every other slot was
  bit-identical. `_pad_positions` pads with each row's own continuing indices instead.

`B=1` keeps the dynamic cache and stays bit-identical. Batched AR without `real_pos` now raises
in `init_state` rather than failing later on an opaque SDPA shape mismatch. The four upstream
constraints are written down in a new `unirl/models/hunyuan_image3/README.md` `## Gotchas`, with
the one-line docstrings pointing at it.

## Related Issue

Fixes #452

## Test Plan

Hardware available for this change was one 4 GB RTX 3050, so the ~80B checkpoint could not be
loaded. Validation instead builds a **tiny random-init `HunyuanImage3ForCausalMM` from the real
upstream remote code** (`tencent/HunyuanImage-3.0-Instruct`, `skip_load_module={"vae","vit"}`,
2 layers / 128 hidden / dense MLP, fp32 CPU, greedy `top_k=1`) and asserts that a batched
right-padded run matches running each row independently at `B=1`. Batch-vs-`B=1` parity is a
plumbing property, so random weights exercise it faithfully. Per `CLAUDE.md` §5 the harnesses
are run and quoted here, not committed.

Harness 1 — uniformly-batched conditions, prefill + decode parity, plus ablations that each
disable one part of the fix:

```
$ python harness_452.py
=== parity: batched right-padded AR vs independent B=1 ===
[PASS] unequal real_pos, 3 decode steps
        tokens equal=True  logp max|d|=4.768e-07  prefill-KV max|d|=1.192e-06
[PASS] extreme length skew, 5 decode steps
        tokens equal=True  logp max|d|=4.768e-07  prefill-KV max|d|=1.192e-06
[PASS] all rows padded (max real_pos < L)
        tokens equal=True  logp max|d|=0.000e+00  prefill-KV max|d|=0.000e+00
[PASS] B=1 unchanged (dynamic cache path)
        tokens equal=True  logp max|d|=0.000e+00  prefill-KV max|d|=0.000e+00

=== ablations: each part of the fix must be load-bearing ===
[FAIL] decode mask removed -> expect FAIL
        tokens equal=False  logp max|d|=9.038e-02
[FAIL] tail-append restored -> expect FAIL
        tokens equal=False  logp max|d|=1.318e-01
ablation: decode mask load-bearing=True  token placement load-bearing=True

RESULT: PASS
```

Harness 2 — the issue's exact construction (per-row `embed_for_ar`, then
`HunyuanImage3FusedMultimodalCondition.concat`), which is what exposed the `position_ids`
padding defect:

```
$ python harness_452_concat.py
repro shape: input_ids=(4, 64) real_pos=[13, 37, 64, 51]

=== parity: concat-built batch vs independent B=1 ===
[PASS] unequal real_pos, 3 decode steps
        tokens equal=True  logp max|d|=4.768e-07  prefill-KV max|d|=1.192e-06  (cache slot 0 max|d|=0.000e+00)
[PASS] extreme length skew, 5 decode steps
        tokens equal=True  logp max|d|=4.768e-07  prefill-KV max|d|=1.192e-06  (cache slot 0 max|d|=0.000e+00)

=== ablation: constant position_ids padding must break parity ===
[FAIL] position_ids pad reverted to 0 -> expect FAIL
        tokens equal=False  logp max|d|=1.292e-01  prefill-KV max|d|=4.069e+00  (cache slot 0 max|d|=4.069e+00)
ablation: position_ids padding load-bearing=True

RESULT: PASS
```

Residual `logp`/KV deltas of ~1e-6 are fp32 batched-GEMM noise: the batched path reduces over a
wider masked key axis than the `B=1` trim, so bitwise identity is not achievable. Token
sequences are exactly equal. Reverting the diff turns both harnesses back into the reported
`AssertionError` at `modeling_hunyuan_image_3.py:965`.

Against the issue's requested checklist: `B=2+` with unequal right-padded `real_pos` ✅;
batched-vs-independent `B=1` logits and KV parity ✅; two-plus decode steps after prefill ✅
(3 and 5); cache growth ✅. **Not run:** FSDP train-side forwards, the real checkpoint, and
finished-sequence masking beyond the per-row cache mask — all need the 8×H20 pod.

Hooks:

```
$ pre-commit run --files unirl/models/hunyuan_image3/{ar.py,conditions.py,README.md}
ruff check...............................................................Passed
ruff format..............................................................Passed
recipe _target_ paths resolve............................................Passed
docstrings are one line..................................................Passed
core dependency directions hold..........................................Passed
experimental-tier boundaries hold............................(see note)...Failed
```

`check-experimental-boundaries` fails only because the local venv is Python 3.10 and the script
imports `tomllib` (3.11+); re-run under `python3.11` it reports `check-experimental-boundaries:
ok`. Every other hook passes.

## Compatibility / Risk

- No config, recipe, checkpoint, data-format, or resource changes.
- `B=1` is unchanged by construction: it keeps `dynamic=True`, gets no injected mask, and its
  token placement is arithmetically identical to the previous tail append. Verified bit-identical.
- `B>1` previously raised `AssertionError`, so there is no prior batched behavior to regress.
- `_pad_positions` changes `concat`-produced `position_ids` padding, which the `gen_image`
  diffusion path also consumes. Unique per-row cache write indices are required there for the
  same `index_copy_` reason, but only the AR path was validated in this PR; a reviewer with pod
  access may want a diffusion smoke test.
- `init_state` now raises `ValueError` when `B>1` and `conditions.tokenizer_output.real_pos` is
  absent. Both in-tree producers (`embed_for_ar` and the vLLM adapter) populate it; an
  out-of-tree caller that does not would newly fail fast.
- The fix depends on upstream remote-code behavior (`modeling_hunyuan_image_3.py`), so the
  README gotchas should be re-checked when the checkpoint revision moves.

## Reviewer Notes

- **Start here:** `_cache_attention_mask` and `_place_token` in `ar.py`, then `_pad_positions`
  in `conditions.py`. The `README.md` gotchas explain what stock upstream does and why each
  piece is load-bearing; the harness ablations above fail deliberately to prove that.
- **Scope:** the issue reports one assertion; this PR fixes four coupled defects, because the
  assertion alone is only the crash. Fixing it in isolation would have produced silently wrong
  batched logits — worse than the crash. The issue's suggested direction 1 was conditional on
  *"if the full static cache remains correctly masked"*; it is not, which is why the decode mask
  is here rather than a bare `dynamic=False`.
- **Duplicate-work check:** scanned the 40 open PRs and the hi3/AR/KV-cache keywords; none touch
  this path (#474 is vLLM-Omni AR seeding, #446 is qwen3-omni prefill logits). No overlap.
- **AI assistance:** written with Claude Code. The submitter has reviewed the full diff; the
  harnesses, their ablations, and the pre-fix reproduction were run locally with the results
  quoted verbatim above.
- Upstream `HunyuanStaticCache.update` could support batched dynamic trimming by slicing at the
  max active position, which was the issue's direction 2. It is not used here because it still
  needs the same per-row decode mask, and it would mean patching remote code that the checkpoint
  owns. Worth an upstream report regardless.

## Checklist


CjhHa1:
Thanks for the thorough write-up. I checked the four fixes against upstream `modeling_hunyuan_image_3.py` and the diagnosis holds: the `len(cache_position)` assert in `HunyuanStaticCache.update` hits the batch dim, `gen_text` decode drops the mask, `prepare_inputs_for_generation` gathers the next input at `position_ids` (= `real_pos`, then `+1` per step), and `position_ids` go straight into `index_copy_` as `cache_position`. The approach looks right. A few things before merging:

**1. Rebase onto #508 (merged).** #508 removed `_build_kv_cache` and its silent `return None` fallbacks. A `None` cache made prefill correct and every decode step context-free, because upstream never creates a cache itself. Cache construction now lives inline in `init_state`, so:
- pass `dynamic=batch_size == 1` to the `HunyuanStaticCache(...)` call in `init_state` instead of adding a helper parameter;
- `past_key_values` can no longer be `None` in the AR loop, so `getattr(past_key_values, "dynamic", True)` in `_cache_attention_mask` can be plain `past_key_values.dynamic`. The default currently only absorbs that removed `None` case.

**2. `pad_id` in `_place_token` is dead.** `state.input_ids` is only consumed by upstream's `input_ids.gather(1, position_ids)`, and every position it reads (`real_pos + s`) was scattered at the previous step. The appended column's fill value is never read. I simulated this with a sentinel fill over mixed `real_pos` and 5 steps, and the gather never returned it. So `int(getattr(transformer.config, "pad_id", 0) or 0)` and the `pad_id` parameter can go; `torch.zeros_like(input_ids[:, :1])` is enough.

**3. Diffusion path needs a check.** `_pad_positions` changes `HunyuanImage3FusedMultimodalCondition.concat`, which `gen_image` also consumes. Diffusion also uses `position_ids` as `cache_position`, so continuing indices should be a fix there too. But it's unvalidated, and `rope_cache` is still zero-padded on the same axis. Could you run a mixed-length diffusion prefill, or at least confirm the diffusion capture path never hits `concat` with ragged `L`? Related but out of scope: `adapters/hi3.py` (`_pad_to(c.get("position_ids"), ..., value=0)`) has the same zero-pad on the vLLM-Omni capture path. Worth a follow-up issue.

**4. Description accuracy.** "`B=1` is unchanged / bit-identical" holds only when `real_pos == L`. For `B=1` with right-pad, the old tail append fed the pad token back as the next input, and `_place_token` now fixes that. That's a behavior change (a fix), so please say so rather than calling `B=1` unchanged.

**5. Scope note.** Finished-sequence masking from #452's checklist isn't addressed: finished rows keep sampling and writing KV. That's harmless for other rows since the mask is per-row, and the packed `TextSegment` stops at the stop token. It's fine to leave, but please list it as not done so #452 isn't closed as fully resolved.

Issue #483, opened 2026-09-18
[RFC] Trainside optimization roadmap: VAE decode on UP, SM-free all-gather, micro-batch scheduling, long-tail batching, embed cache, quantitative offload


nussejzz:
**Status:** Draft
**Scope:** trainside rollout engine (`unirl/rollout/engine/trainside`), FSDP backend (`unirl/train/backend/fsdp`), diffusion trainer step loop (`unirl/trainer/diffusion.py`), reward stack

## Summary

This RFC collects six trainside throughput/memory items into one roadmap so that they can be sequenced, measured with one method, and kept default-off until each one shows a number. Several already have an open PR or RFC; this document says where they fit together, what the measured ceilings are, and which ones are worth doing first.

1. VAE decode parallelism on top of Ulysses SP (UP)
2. SM-free (copy-engine) all-gather for FSDP overlap, gated by measured exposed communication
3. Configurable micro-batch scheduling for trainside rollout, with rollout / remote-reward overlap
4. Long-tail mitigation for video-reference workloads: micro-batch granularity + DiT continuous batching
5. `input_embed` cache reuse and encoder offload after conditioning
6. Quantitative CPU layerwise offload + FSDP overlap planner (MFU/MBU model, resident/offloaded split; vLLM-Omni DLO as the reference design)

All numbers below are measured on 8×H20 96GB (Taiji) unless stated otherwise.

## Motivation

The trainside path is now the highest-throughput recipe we have for large video DiTs (Leo2 720p, 16 GPUs: DP trainside 100 clips/h/machine vs 73.6 for the best colocated CP8 bf16 arm), but a rollout is still 281 s per clip (266 s denoise + 15 s VAE decode), the training step is 72% of the end-to-end round, and the whole loop is serial: generate the DP shard → score it → replay/backward. Each item below attacks one of those seams. The common thread is that the naive version of every item has already been tried somewhere and was *not* free, so each proposal comes with the measurement that decides whether it is enabled.

### 1. VAE decode parallelism on top of UP

Under Ulysses SP every rank finishes denoising with identical full-sequence latents and decodes the whole video independently; #469 measured the H3 video VAE at 73.7% of the rollout step (51.4 s/sample at 768×768×124, `sp_size=8`) and shards the decoder's temporal `_decode_clip` calls across the SP group.

Two measured constraints shape the general version:

[...]

PR #485, opened 2026-09-18
fix(rollout): validate aligned prompt-group chunk boundaries

0z5a:
## Summary

Replace whole-group shape bucketing with an optional `validate_group_boundaries` check (default false). GRPO siblings can run in separate forwards: results are regrouped before advantages are computed. The check accepts whole groups or aligned equal subgroups and never changes `forward_batch_size`, row order, or memory limits.

For example, group size 16 with FBS 4 keeps four 4-row forwards. Unsupported heterogeneous-shape scheduling is deferred until per-row geometry and ragged trajectories exist.

## Related Issue

Refs #483. This is boundary validation only, not completion of shape bucketing or continuous scheduling.

## Test Plan

RTX 5090, torch 2.14.0+cu130. External harness: `PYTHONPATH=. python ../validate_unirl.py`, using the real engine, Sample/Part, trajectory and GRPO advantage code with a deterministic tiny BF16 linear module; no checkpoint or dataset downloads.

| Check | Result |
| --- | --- |
| CPU/GPU validation on/off parity | 14 cases passed; identical calls, row/noise IDs, latents, log-probs, rewards and advantages |
| Invalid boundaries | 3 helper cases rejected; engine rejects before forward and restores training mode |
| Built-in trainside recipes | 61/61 boundary checks passed |
| Group/FBS 16/4, 16/8, 16/1, 24/1 | All accepted with the original chunk sizes |

`ruff check unirl/rollout/engine/trainside/`, `ruff format --check unirl/rollout/engine/trainside/`, `python lint/check_docstring_lines.py`, and `python lint/check_core_dependencies.py` pass.

Full pretrained training was not run; this change only validates existing chunk boundaries. No throughput or reward-improvement claim.

## Compatibility / Risk

Existing recipes retain their exact schedule by default. The earlier proposed shape-bucket options are removed from this unmerged PR; the only new option is `validate_group_boundaries`.

## Reviewer Notes

Checked open overlapping work, including #445 and #465; this adds no reward-overlap or replay-backward machinery. Heterogeneous execution scheduling remains separate.

## Checklist

- [x] I updated documentation and validated the change with a repo-external harness.


CjhHa1 (changes requested):
Thanks for working on this. Before discussing the implementation details, I am not yet convinced that keeping every prompt group in one model forward is an algorithmic requirement.

I checked the 61 built-in trainside recipes at this PR's merge base:

- 48 split one prompt group across multiple `pipeline.generate` calls.
- 13 avoid splitting through aligned/unset `forward_batch_size` or group size 1.
- The existing split configurations are deliberate aligned sub-chunks (`group_size % forward_batch_size == 0`); I found no built-in recipe with an accidental non-divisor boundary.

The split does not currently break GRPO grouping:

- chunks are concatenated in original `sample_ids` order before reward scoring;
- group advantages are computed on the complete reconstructed frontier;
- driver-authored `x_T` is keyed by sample/group identity, so `init_same_noise` remains consistent across calls;
- rollout trajectories and log-probs remain row-aligned.

Changing chunk geometry can change global SDE RNG assignment and BF16 numerics, but that is a reproducibility/batch-geometry effect, not by itself evidence that GRPO siblings must share one forward.

Conversely, many recipes intentionally use a smaller forward batch for memory:

- WAN21: group 16, FBS 4
- Z-Image: group 16, FBS 8
- WAN22: group 16, FBS 1
- VideoAlign: group 24, FBS 1

`shape_bucket` rejects these unless one whole group fits the declared capacity, so it cannot replace most current trainside configurations without increasing peak memory, potentially to OOM.

Could you first establish the algorithmic motivation with a concrete invariant or experiment?

1. What becomes incorrect when siblings are generated in separate forwards?
2. Is there a measured reward/optimization regression, rather than only small batch-shape numerical differences?
3. Which real recipe can enable this mode while fitting a whole group in memory?
4. What throughput or tail-latency improvement does that recipe obtain?

Also, heterogeneous shape bucketing is not reachable today: `Part.sampling_params` and `init_noise_latent_shape` are shared, `_execution_keys()` repeats one key for every row, and dense trajectories cannot merge different geometries. No existing recipe enables this mode.

If there is no demonstrated same-forward algorithm invariant, I would prefer a much smaller change: validate or plan group-aware boundaries while preserving memory-safe aligned sub-group execution. The full execution-key/shape-bucket machinery should wait until per-row geometry and ragged trajectory support exist.

0z5a:
Agreed. Replaced the shape-bucket planner with a small optional boundary check, preserving aligned subgroups and the existing forward-size cap. No claim that GRPO siblings must share a forward; heterogeneous scheduling is deferred.

| Validation | Result |
| --- | --- |
| CPU/GPU output and advantage parity | 14 cases passed |
| Built-in recipe boundary checks | 61/61 passed |
| 16/4, 16/8, 16/1, 24/1 | Accepted, unchanged chunk sizes |
| Lint and core guards | Passed |

These use the real engine with a tiny deterministic BF16 module; full pretrained training and throughput were not measured.

Issue #488, opened 2026-09-18
[RFC] Complete decay schedulers and explore TempFlow-GRPO / Flash-GRPO


ruiling-smartbear:
## 1. `decay` / `exp_decay` are unreachable

`unirl/sde/index_schedule.py:10` declares:

```python
Strategy = Literal["all", "progressive", "random", "decay", "exp_decay"]
```

but `WindowScheduler.WINDOW_STRATEGY_TO_METHOD_NAME` only maps `all` / `progressive` / `random`, so setting either one raises `Bad strategy configuration` at `index_schedule.py:116`.

The config surface is already fully in place: `WindowConfig` carries `max_iters_per_window`, `min_iters_per_window`, `exp_decay_threshold=13`, `exp_decay_k=0.1`, and `__post_init__` (line 28) has a dedicated `strategy == "decay"` branch that fills the two `iters_per_window` bounds.

Three example configs also set the decay-only fields while running `progressive`, so those lines are currently dead:

| config | strategy | dead fields |
| --- | --- | --- |
| `examples/diffusion/sd3/sd3_mixgrpo.yaml` | L140 | L146-147 |
| `examples/diffusion/wan21/wan21_t2v_mixgrpo.yaml` | L146 | L152-153 |
| `examples/diffusion/wan22/wan22_t2v_14b_mixgrpo.yaml` | L145 | L151-152 |

**Question:** is this pending work, or was `decay` dropped on purpose? If it's pending I'd like to implement both resolvers — it looks like ~50 lines plus a determinism check on `get_sde_indices`, and it would make the MixGRPO-Flash variant reachable from the existing configs.

## 2. Follow-up: TempFlow-GRPO and Flash-GRPO

I'd also like to contribute to these two methods, separately from #489:

- **[TempFlow-GRPO](https://arxiv.org/abs/2508.04324):** explore support for trajectory branching, noise-aware optimization weights, and seed grouping. This needs rollout and training changes beyond timestep selection.
- **[Flash-GRPO](https://arxiv.org/abs/2605.15980):** #244 already implements the algorithm and SGLang rollout. I'd be happy to help validate that work, starting with reference parity and a short Wan2.1-1.3B training run.

The aim would be small, focused changes with correctness checks and reward-versus-time comparisons under matched compute budgets.

Would TempFlow-GRPO support and validation help on #244 be useful? Is anyone already working on TempFlow-GRPO?

ruiling-smartbear:
[...]

PR #489, opened 2026-09-18
feat(sde): implement WindowScheduler `decay` / `exp_decay` strategies

ruiling-smartbear:
## Summary

`Strategy` declares `decay` and `exp_decay`, and `WindowConfig` fully describes them (`max_iters_per_window`, `min_iters_per_window`, `exp_decay_threshold`, `exp_decay_k`, plus the `decay` branch in `__post_init__`) — but `WindowScheduler` has no resolver for either, so selecting one raises `Bad strategy configuration`. Four shipped mixgrpo recipes (`sd3_mixgrpo`, `wan21_t2v_mixgrpo`, `wan22_t2v_14b_mixgrpo`, `qwen_image_mixgrpo`) already set the decay-only fields while running `progressive`, so those lines are currently inert.

This adds both resolvers. Semantics follow MixGRPO's `GRPOTrainingStates` (`fastvideo/utils/grpo_states.py`), where `t` is the window start index:

- `decay` — `iters(t) = max(min, int(max*(1-t/T) + min*t/T))`
- `exp_decay` — `iters(t) = ceil(iters_per_window * exp(-k * relu(t - threshold)))`

The stateful reference is mapped onto the existing stateless `get_sde_indices(step)` contract by walking the same window sequence `_resolve_progressive` uses (full windows only; `roll_back` wraps modulo the sweep length, otherwise clamp to the last window), with a per-window dwell instead of a constant `iters_per_window`.

## Related Issue

Refs #488

## Test Plan

Parity harness against the MixGRPO reference over 216 parameter combinations (`num_timesteps` ∈ {10,14,25} × `window_size` ∈ {2,3,4} × `iters_per_window` ∈ {4,10,25} × `overlap_size` ∈ {0,1} × `roll_back` ∈ {false,true}), replaying `GRPOTrainingStates.update_iteration()` step by step against `get_sde_indices(step)`:

- first-sweep window sequence matches the reference exactly — **216/216**
- post-sweep convention (periodic under `roll_back`, else clamped to the last full window) — **216/216**
- determinism, window size, index bounds, monotone non-increasing dwell — all pass

Resulting dwell with `num_timesteps=10, window_size=4, overlap_size=1`:

| strategy | dwell per window |
| --- | --- |
| progressive | 25 / 25 / 25 |
| decay (max=10, min=1) | 10 / 7 / 4 |
| exp_decay (threshold=3, k=0.1) | 25 / 25 / 19 |

Lint: `ruff check unirl/sde/index_schedule.py` and `ruff format --check unirl/sde/index_schedule.py` — clean.

Regression check on the untouched strategies: the pre-change and post-change `WindowScheduler` were driven side by side over 93,312 `(config, step)` pairs (`strategy` ∈ {`all`, `progressive`, `random`}) plus 360 `AllSDEScheduler` pairs — **0 mismatches**, and the `strategy="all"` short-circuit is unaffected.

Training smoke test: not run. The change only affects which indices `get_sde_indices` returns; the parity check above covers that surface directly.

The harness was run but not committed, per `CLAUDE.md`.

## Compatibility / Risk

No API, config-schema, or checkpoint changes. Existing `all` / `progressive` / `random` paths are untouched. Previously `decay` / `exp_decay` raised at construction, so no behaviour can regress — the new `decay` bounds check (`1 <= min <= max`) only fires on a strategy that was unusable before.

## Reviewer Notes

One judgement call worth a look: UniRL's `_resolve_progressive` already differs from the reference past the end of a sweep (it keeps to full windows and wraps, where the reference clips `cur_timestep` to `max_timesteps`). I followed UniRL's convention rather than the reference's so the three strategies behave consistently; happy to switch if you'd rather match MixGRPO exactly.

AI-assisted. Checked open issues and PRs for overlapping work before opening.

## Checklist


Jayce-Ping (approved):
Looks okay. Though I think most of time, users will use FlowGRPO-Fast's random SDE-step selection rather than MixGRPO, it's still worth to fix.

PR #493, opened 2026-09-20
fix(sglang): tighten rollout argument guidance

CjhHa1:
## Summary

- inline the two single-use dropped-argument helpers into the backend filter that owns the behavior
- align the safe server-port bound with pinned SGLang 0.5.12.post1's actual `grpc_port = port + 10000` derivation
- narrow the Qwen3 guidance to a 4B dense baseline and document larger/MoE warmup, attention-backend, CUDA-graph, and temporary-port exceptions without changing recipe values

## Related Issue

N/A — follow-up to #392.

## Test Plan

- `SKIP=no-commit-to-branch uvx pre-commit run --all-files --show-diff-on-failure` — passed
- CPU assertion harness for UniRL-only key filtering, unknown-key warn/strict behavior, required-key precedence, the 55535 port boundary, repeated port reservation, and YAML parsing of all 9 touched recipes — passed
- GPU rollout: Not run; production behavior is unchanged apart from expanding the valid server-port range to match the pinned SGLang derivation. Recipe edits are comments only.

The focused assertions remain one-off because the repository intentionally keeps no committed top-level test tree.

## Compatibility / Risk

Default warn/strict behavior and filtered output are unchanged. The server-port candidate range expands from 1024–35535 to 1024–55535; under pinned SGLang 0.5.12.post1, an automatically derived gRPC port therefore remains within 65535. No checkpoint, data-format, recipe-value, GPU-count, or memory-requirement changes.

## Reviewer Notes

PR #471 migrates SGLang to 0.5.19 and removes the derived-port cap while renaming CUDA-graph fields. If it merges first, this branch should be rebased: the 0.5.12-specific config hunk and wording become obsolete, while the helper inlining and baseline/exception documentation still apply.

AI-assisted implementation; the submitter reviewed the full diff. Duplicate-work review found #471's version-migration overlap described above, but no open PR covering the #392 helper/documentation follow-up.

## Checklist


CjhHa1:
Superseded by the rebased #471: the helper inlining and documentation cleanup are now included there, while the 0.5.12-specific gRPC port correction is obsolete under the 0.5.19 migration.

PR #496, opened 2026-09-20
fix(diffusion): broadcast transition sigma to replay mean rank

### LINKED ISSUE #447: fix(diffusion): broadcast transition sigma to replay mean rank
## Problem

`_transition_sigma()` always returns a five-dimensional tensor with shape
`[1, S, 1, 1, 1]`, regardless of the rank of the replayed transition means:

```python
return sigma_t.reshape(1, -1, 1, 1, 1)
```

That shape only broadcasts correctly for image means shaped
`[B, S, C, H, W]`. The same helper is shared by FlowDPPO, FlowGRPO's reference
KL, and DiffusionOPD, while supported stages also return:

- video means: `[B, S, C, T, H, W]` (6-D)
- SenseNova packed-pixel means: `[B, S, seq, C]` (4-D)

## Impact

For video means, KL computation fails with a broadcast error. For example:

```text
means:   [2, 3, 16, 10, 64, 64]
sigma_t: [1, 3, 1, 1, 1]
RuntimeError: The size of tensor a (16) must match the size of tensor b (3)
              at non-singleton dimension 2
```

For four-dimensional SenseNova means, the common `micro_batch_size: 1` case
silently broadcasts to a step cross-product instead of failing:

```text
means:   [1, 3, 128, 3]
sigma_t: [1, 3, 1, 1, 1]
result:  [1, 3, 3, 128, 3]  # expected [1, 3, 128, 3]
```

With `B > 1` and `B != S`, the same path raises a broadcast error instead.

This is reachable in the shipped
`examples/diffusion/sensenova_u1_5/sensenova_u1_5_ocr_trainside_lora_kl_4x8.yaml`:

- `FlowGRPO`
- `beta: 0.01`, which enables `_reference_kl_loss`
- `micro_batch_size: 1`
- three SDE steps

The silent extra `S` dimension changes both the reference-KL value and its
memory use.

## Expected behavior

The transition standard deviation should broadcast only over the batch and
latent dimensions while preserving the step axis, for example:

```python
sigma_t.reshape(1, S, *([1] * (means.ndim - 2)))
```

The exact API can either pass the mean rank into `_transition_sigma()` or shape
the returned per-step values at each mean consumer.

## Suggested validation

Add focused checks for:

- image means `[B, S, C, H, W]`
- video means `[B, S, C, T, H, W]`
- packed means `[B, S, seq, C]`
- `B=1` and `B>1`
- FlowDPPO, FlowGRPO reference KL, and DiffusionOPD callers

## Context

Surfaced while reviewing #444; the bug predates that PR and is independent of
the replay-mean dtype fix.

ruiling-smartbear:
## Summary

`_transition_sigma()` returned a hardcoded five-dimensional `[1, S', 1, 1, 1]`, so it only broadcast correctly against image means shaped `[B, S', C, H, W]`. That helper is shared by FlowDPPO's KL-ADV mask loss, FlowGRPO's reference KL and DiffusionOPD, and the supported stages also return video means `[B, S', C, T, H, W]` and SenseNova packed-pixel means `[B, S', seq, C]`.

This returns the per-step `[S']` straight from `strategy.transition_std()` and lets `_gaussian_kl_div` — the single choke point that every one of those consumers goes through — broadcast it as `[1, S', *([1] * (p.ndim - 2))]`.

Measured for `S' = 3`:

| means | before | after |
| --- | --- | --- |
| image `[2, 3, 4, 8, 8]` | `[2, 3, 4, 8, 8]` | `[2, 3, 4, 8, 8]` (unchanged) |
| video `[2, 3, 4, 5, 8, 8]` | `RuntimeError` | `[2, 3, 4, 5, 8, 8]` |
| packed `[1, 3, 128, 3]` | `[1, 3, 3, 128, 3]` — **silently wrong** | `[1, 3, 128, 3]` |
| packed `[2, 3, 128, 3]` | `RuntimeError` | `[2, 3, 128, 3]` |

The packed `B = 1` case is the dangerous one: it does not raise, it turns the per-step sigma into a step cross-product, so both the KL value and its memory use change with no error. It is reachable in the shipped `examples/diffusion/sensenova_u1_5/sensenova_u1_5_ocr_trainside_lora_kl_4x8.yaml` (FlowGRPO, `beta: 0.01`, `micro_batch_size: 1`).

## Related Issue

Fixes #447

## Test Plan

Harness (run, not committed, per `CLAUDE.md`) that AST-extracts the pre-fix and post-fix functions verbatim from `unirl/algorithms/{base,flowdppo}.py` and drives them against an independent per-step reference computation (`torch.stack` over `s`, no broadcasting), because `unirl.algorithms.base` imports `ray` at module level:

| check | result |
| --- | --- |
| `_gaussian_kl_div` shape and value, all four mean ranks, `B = 1` and `B > 1` | 4/4 correct, matches the per-step reference |
| `_reference_kl_loss` (FlowGRPO / FlowDPPO reference KL) | 4/4 match the reference |
| `_flowdppo_kl_adv_loss` loss shape and `kl_new_old_mean` | 4/4 match the reference |
| `add_coefficient=False` | shape `[S']`, all ones |
| image `[B, S', C, H, W]` regression vs `upstream/main` | `torch.equal(before, after)` → `True` |

Lint:

```bash
SKIP=no-commit-to-branch pre-commit run --files unirl/algorithms/base.py unirl/algorithms/flowdppo.py
```

All hooks pass, including the local `docstrings are one line` guard.

Training / rollout smoke test: not run. The change only affects how a per-step scalar is broadcast onto the replayed means; the harness above covers that surface directly for every consumer, and the 5-D path is bit-identical to the previous behaviour.

## Compatibility / Risk

No API, config, or checkpoint change. `_transition_sigma` and `_gaussian_kl_div` are module-private, and all four call sites reach them only through `_gaussian_kl_div`. Image means are bit-identical to before, so existing 5-D recipes cannot regress; video and packed means move from a crash or a silent wrong shape to the intended arithmetic.

## Reviewer Notes

I took the second of the two options in the issue — shape the per-step values at the mean consumer — because `_gaussian_kl_div` is the only place the mean rank is known, which keeps all four call sites untouched.

AI-assisted. Checked open issues and PRs for overlapping work before opening; no other open PR references #447.

## Checklist


CjhHa1 (approved):
LGTM

